PR #19, opened 2020-09-11
Add Linux instructions to HOWTO.md

s-zeng:
Some updates to the HOWTO.md to provide (currently very hacky) instructions to deploy on Linux. There's still much to go in terms of usability, but in principle all the steps are automatable.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -8,14 +8,18 @@ MacOS.
 
 
 ## Prerequisites
+
 * docker
-* homebrew
 * minikube
 * python3
 * a ZeroTier network and api access token
 * jq
 
-## Installation
+## Installing prerequisites 
+
+This section varies depending on OS.
+
+### Mac w/ homebrew
 
 Make sure homebrew is installed:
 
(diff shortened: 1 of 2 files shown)


HOWTO.md
```diff
@@ -69,17 +87,44 @@ ZT_NET=1c33c1ced02a5eee
...
 sudo sh -c 'echo "/System/Volumes/Data/Users/ -alldirs -mapall=501:1000 $(minikube ip)" >> /etc/exports'
 sudo nfsd restart
 ```
 
+For Linux:
+
+``` shell
+sudo sh -c 'echo "$HOME/.tq $(minikube ip)(rw,sync,no_subtree_check,no_root_squash)" >> /etc/exports'
+sudo exportfs -arv
+mkdir ~/.tq; mkdir ~/.tq/node; mkdir ~/.tq/client/; touch ~/.tq/client/wallet_lock
```
itkach, line 102: `mkdir -p .tq/{node,client}` should work on both Linux and Mac... is touching `wallet_lock` really necessary?

HOWTO.md
```diff
@@ -69,17 +87,44 @@ ZT_NET=1c33c1ced02a5eee
...
+chmod -R 777 ~/.tq/node
```
itkach, line 103: these look like excessive permissions

HOWTO.md
```diff
@@ -69,17 +87,44 @@ ZT_NET=1c33c1ced02a5eee
...
 Give your private chain a name:
 
 ``` shell
 CHAIN_NAME=my_chain
 ```
 
+If you're on Linux, you may also need to enforce unbuffered IO for your python:
+
+``` shell
+PYTHONUNBUFFERED=x
```
itkach, line 125: why, what problem user may encounter if this is not set?

s-zeng: Using regular buffered Python IO can intermittently lead to cases where pipes from mkchain to kubectl can hang. Setting unbuffered Python IO resolved this on my machine.

itkach: perhaps it is better then to just instruct users to set this always instead of "you may also need", or just change all python invocations to include equivalent `-u` command line option (we didn't run into these issues in MacOS but it seems like running unbufferred is the right thing to do there as well)

HOWTO.md
s-zeng: Apologies for these hacky workarounds. I've implemented a better solution, pushing it up now.

itkach (approved):

phpcodecrafting (approved):

PR #29, opened 2020-10-15
NFS removal (fix #27) - --generate-constants - EKS support - open up rpc

### LINKED ISSUE #27: remove nfs from default deployment
nfs setup is cumbersome

for default minikube setup rely on auto provisioned storage

nicolasochem:
Removes the requirement of having NFS-mounted /var/tezos dir.

When creating a new chain, there is one extra command to generate
constants: bootstrap account keys, chain id and timestamp.

The command is

  mkchain --generate-constants <CHAIN_NAME>

It generates 2 files: <CHAIN_NAME>_chain.yaml and
<CHAIN_NAME>_chain_invite.yaml

Then everything works as expected.

2 new init containers have been introduced:

* import-keys: injects the keys in the node (private keys for bootstrap
node, public keys for invited nodes)

* generate-tezos-config creates config.json and parameters.json from
the given constants (this code was previously in mkchain)

This is an intermediary step towards making the chain creation process
declarative: give your constants as input, get a working k8s spec as
output, for both bootstrap nodes and invited nodes.

For now:

The --generate-constants uses local docker containers (tezos and flextesa)
to generate constants.

The --create step is self-contained and just doing templating.

The --invite step queries the bootstrap cluster for the nodeport and
requires the ip of the bootstrap node to be passed manually, but it
no longer downloads its (config|parameter).json files from the cluster,
instead generating them using the same mechanism that the bootstrap node.

Other changes:

* open up the RPC port of the nodes

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -41,16 +41,6 @@ brew install minikube
 pacman -Syu && pacman -S python3 jq minikube kubectl kubectx linux
 ```
 
-Ensure that the kernel used is compiled with the `nfs` module. The default arch 
-linux kernel should suffice.
-
-Enable and start the `rpcbind` and `nfs-server` systemd services:
-
-```shell
-sudo systemctl enable rpcbind && sudo systemctl start rpcbind
-sudo systemctl enable nfs-server && sudo systemctl enable nfs-server
-```
-
 ## Setting up mkchain
 
 Ensure minikube is running:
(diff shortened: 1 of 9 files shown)


s-zeng (changes requested):
Looks good and works on my machine
Just needs a `setup.py` change to make sure that the shell scripts make it into the pip package

s-zeng (approved):

Issue #17, opened 2020-09-04
Implement ability to spin up a local test network with configured number of bakers and nodes


harryttd:

harryttd:
WIP

- Currently using a statefulSet which provisions storage for each tezos node that will be spun up. Using kubectl cli to increase/decrease replicas while testing.
- Nodes are communicating with each other properly through the `tezos-p2p` service.
- I gradually increased the number of replicas to 32, and things started to crash at that point. Need to look more into resource provisioning for all of the nodes.

Occasionally I am experiencing issues with logs looking similar to:
```
Sep  4 20:47:59 - validator.peer: Worker started for NetXHFw7TkU5h:idsB7s3QQeVB
Sep  4 20:47:59 - block_validator_process_external: The process terminated abnormally with value 1
Sep  4 20:47:59 - block_validator_process_external: Block validator started on pid 639
Sep  4 20:47:59 - validator.block: Validation of block BM1nDiexDsYU7iF4YVXMsCU5jN3bSX2exvj3VUGfPqHHmdef3AR failed
Sep  4 20:47:59 - validator.block: Request pushed on 2020-09-04T20:47:59-00:00, treated in 3.459us, completed in 69.801ms ,
Sep  4 20:47:59 - validator.block: (Invalid_argument "Json_encoding.construct: consequence of bad union")
Sep  4 20:47:59 - node.validator.bootstrap_pipeline: unexpected error while fetching headers
Sep  4 20:47:59 - validator.peer: Worker crashed [node_peer_validator]:
Sep  4 20:47:59 - validator.peer: (Invalid_argument "Json_encoding.construct: consequence of bad union")
``` 

This may be related to similar issues I'm sometimes seeing with the `activate-job` #18.

nicolasochem:
@harryttd with the new approach on #29 , I believe we can have all nodes (progenitor and regular nodes) come online at the same time, since they now have all the data that they need in a configmap.

As a consequence, it would no longer be necessary to run kubectl commands from within the k8s environment.

The feature can be made part of mkchain - just specify a desired number of nodes as a parameter.

brandisimus:
Closing as duplicate of the #35

PR #31, opened 2020-10-23
Multinode test network with configurable number of nodes. #17

nicolasochem:
Based on aryeh/test-chain branch

Pod names have been changed as follows for clarity:

* the baking node is now called "tezos-progenitor"
* the non-baking nodes are called simply "tezos-nodes"

A new parameter `--number-of-nodes N` lets you choose
how many nodes should be in your test topology.

You can scale up or down the topology by reissuing the same
mkchain command with a different number of nodes, then pipe
to apply to an existing cluster.

Here is an example output of `get pods` with 3 nodes:

```
NAME                                READY   STATUS      RESTARTS   AGE
activate-job-4ks45                  0/1     Completed   0          50s
tezos-node-0                        1/1     Running     0          49s
tezos-node-1                        1/1     Running     0          49s
tezos-progenitor-77c6c87586-5gg7h   2/2     Running     0          49s
```

The --number-of-nodes parameter is also supported for "invite chains".
It is possible to create a full mesh of nodes connecting to a
progenitor in another cluster.

The yaml template files have been reorganized:

* progenitor.yaml contains progenitor deployment config
* node.yaml contains node statefulset config

Other changes:

* give zerotier its own persistent volume to save its state
(it used to go to the node storage)
* remove some commented out code
* remove stale pv.yaml file


HOWTO.md
```diff
@@ -117,13 +117,36 @@ kubectl -n tqtezos get pods
...
+a number of nodes of your choice by passing the following argument to
+mkchain:
+
+```
+mkchain --create --baker --number-of-nodes 3 <...> mychain
+```
+
+The nodes will establish peer-to-peer connection in a full mesh topology.
+
+If you previously spun un the chain with just one node, you may scale
```
s-zeng, line 138: "spun un" should be "spun up"

s-zeng (commented):
Changes look good and works on my machine :+1:

HOWTO.md
```diff
@@ -117,13 +117,36 @@ kubectl -n tqtezos get pods
...
 
-## Add Nodes
+## Add nodes within the cluster
+
+You can launch a self-contained testnet within your cluster with
+a number of nodes of your choice by passing the following argument to
+mkchain:
+
+```
+mkchain --create --baker --number-of-nodes 3 <...> mychain
```
harryttd, line 133: I personally found it a little confusing which exact mkchain command this is referring to from earlier in the howto. Any way you can either rewrite the full cmd, `mkchain --create --baker --number-of-nodes 3 --zerotier-network $ZT_NET \
--zerotier-token $ZT_TOKEN $CHAIN_NAME` or at least make the last word `mychain` into `$CHAIN_NAME` to make it consistent?

HOWTO.md
```diff
@@ -117,13 +117,36 @@ kubectl -n tqtezos get pods
...
+
+If you previously spun un the chain with just one node, you may scale
+up your setup to an arbitrary number of nodes by applying the resulting
+yaml on the cluster:
+
+```
+# create
+mkchain --create --baker mychain | kubectl apply -f -
+# scale up
+mkchain --create --baker --number-of-nodes 3 mychain | kubectl apply -f -
```
harryttd, line 146: Can this be from `mychain` -> `$CHAIN_NAME`

tqchain/deployment/bootstrap-node.yaml
```diff
@@ -0,0 +1,168 @@
...
+  ports:
+  - port: 8732
+    nodePort: 31732
+  selector:
+    app: tezos-bootstrap-node
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: tezos-bootstrap
```
harryttd, line 84: Any way we can rename this to reflect more that this is the service for the p2p network? I don't think it is so obvious from the name. One might only realize it is p2p if they know that is what port 9732 is used for. So maybe like for the rpc, `tezos-bootstrap-node-p2p`

tqchain/mkchain.py
```diff
@@ -192,6 +192,7 @@ def get_args():
 
     subparsers = parser.add_subparsers(help="clusters")
 
+    parser.add_argument("--number-of-nodes", help="number of peers in the cluster", default=1, type=int)
```
harryttd, line 195: Could there be a reason to allow for the default peers amount to be 0? Maybe someone just wants a bootstrap node.

tqchain/mkchain.py
```diff
@@ -362,6 +351,25 @@ def main():
                             get_baker(args.docker_image, args.baker_command)
                         )
 
+                if safeget(k, "metadata", "name") == "tezos-node":
+                    # set the docker image for the node
+                    k["spec"]["template"]["spec"]["containers"][0][
+                        "image"
+                    ] = args.docker_image
+
+                    # add key import for bootstrap node
```
harryttd, line 360: I think technically this comment should be referring to the peer node. Not the bootstrap node as the `if` block we are in is handling for the statefulSet nodes.

tqchain/mkchain.py
```diff
@@ -219,6 +220,10 @@ def main():
     zerotier_network = args.zerotier_network
     zerotier_token = args.zerotier_token
 
+    if args.number_of_nodes < 1:
+        print(f"Invalid argument --number-of-nodes {arg.number_of_nodes}, must be 1 or more")
```
harryttd, line 224: `arg` should be `args`

tqchain/mkchain.py
```diff
@@ -362,6 +351,25 @@ def main():
...
+                        "initContainers"
+                    ].insert(0, get_import_key_job(args.docker_image))
+
+                    # add the identity job
+                    k["spec"]["template"]["spec"][
+                        "initContainers"
+                    ].append(get_identity_job(args.docker_image))
+
+                    # set replicas
+                    k["spec"]["replicas"] = args.number_of_nodes - ( 1 if args.create else 0 )
```
harryttd, line 371: Scaling down the statefulSet only works when you have more than 1 peer node. I believe the reason is because if say you have just one peer node and set `--number-of-nodes` to 1, the newly generated yaml lacks the statefulSet, so k8s keeps whatever statefulSet it already has running. You can still generate the statefulSet yaml but set the `replicas` field to 0. That will remove all instances of the statefulSet.

nicolasochem:
@harryttd all resolved

harryttd (approved):

s-zeng (approved):

PR #36, opened 2020-10-29
Aryeh/rpc-auth

harryttd:
- Install [devspace](https://devspace.sh/cli/docs/getting-started/installation)
- `mkchain generate-constants....`
- `devspace use namespace tqtezos`
- `cd ./docker/rpc-auth`
- `devspace dev --var=CHAIN_NAME="$CHAIN_NAME"`
- (If you want to pass more mkchain args, you can do `--var=CHAIN_ARGS="--docker-image=tezos/tezos:v8-release ..."`)

The above will build (if not already built) the python and zerotier docker images, and apply the yaml for mkchain and rpc auth server. It also enables the minikube nginx addon so that the rpc-auth ingress can work.

Thoughts/Comments on first iteration of this auth mechanism:
- If this will turn into a separate repo, we need to especially keep in mind that the namespaces are consistent across them so that all the services/pods can talk to each other (easily).
- Clients can request as many new url's as they want.
- Need to understand more what potential attack vectors there are and how to prevent. (I believe a replay attack with regards to signing the server-sent-nonce is avoided, by running the getting and deleting of the nonce (that we are verifying for) in a redis transaction.)
- Probably need to configure redis in more detail. For example, how often it persists data.


rpc-auth/server/backend.yaml
```diff
@@ -0,0 +1,161 @@
...
+            timeoutSeconds: 5
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - "redis-cli -h $(hostname) ping"
+            initialDelaySeconds: 5
+            periodSeconds: 3
+          # resources:
```
harryttd, line 64: k8s noob question: at what point in time do we configure resources for all of our different infrastructure objects? How do we determine what the resources should be?

rpc-auth/server/index.py
```diff
@@ -0,0 +1,135 @@
...
+    return urljoin(request.url_root, f"tezos-node-rpc/{access_token}")
+
+
+def is_valid_access_token(access_token):
+    if redis.get(create_redis_access_token_key(access_token)) != None:
+        return True
+    return False
+
+
+## Need a proper server for production
```
harryttd, line 131: Need to configure a proper server for production

rpc-auth/server/index.py
```diff
@@ -0,0 +1,135 @@
...
+        print("Error verifying signature:", e)
+        return False
+
+
+def create_redis_access_token_key(access_token):
+    return f"access_token:{access_token}"
+
+
+def generate_secret_url(public_key):
+    access_token = str(uuid4())
```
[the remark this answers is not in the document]
harryttd: I will remove the create_nonce() func as I don't think it is necessary. create_nonce() right now returns a uuid in hex format as that is required for the tezos-client while here it is plain uuid (cd470ace-264f-418b-95a2-4202bd8956dc). Not sure what you mean by "spelled differently". Would like this function to generate an access token in hex format as well?

rpc-auth/server/index.py
```diff
@@ -0,0 +1,135 @@
...
+from flask import request
+from pytezos.crypto import Key
+from redis import StrictRedis
+from urllib.parse import urljoin
+from uuid import uuid4
+import os
+import requests
+
+TEZOS_CHAIN_ID = os.getenv("TEST_CHAIN_ID")
+TEZOS_RPC = f"{os.getenv('TEZOS_RPC')}:{os.getenv('TEZOS_RPC_PORT')}"
```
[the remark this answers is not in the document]
itkach: Ah, I see generate_secret_url actually uses request.url_root... I'm not sure how Flask application would not it's public url (must explicit configuration?), but this assumes that this service would be doing the proxying of actual requests, which it shouldn't

rpc-auth/server/index.py
```diff
@@ -0,0 +1,135 @@
...
+
+    request_method = getattr(requests, request.method.lower())
+    return request_method(urljoin(f"http://{TEZOS_RPC}", rpc_endpoint)).text
+
+
+## HELPER FUNCTIONS
+
+
+def verify_chain_id(chain_id):
+    global TEZOS_CHAIN_ID
```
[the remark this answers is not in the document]
itkach: > Without setting global TEZOS_CHAIN_ID i get a local var referenced before assignment error.

only because you try to assign it; caching function (e.g. using https://docs.python.org/3/library/functools.html#functools.cache) would provide a cleaner implementation (less global mutable state)

rpc-auth/server/index.py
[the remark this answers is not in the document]
itkach: the result is exactly the same as far as I can tell - hex string without dashes
```
❯ python
Python 3.8.6 (default, Oct  8 2020, 14:07:53)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import uuid
>>> uuid.uuid4()
UUID('b7e0552e-6de4-4d5b-965e-48b9970f52db')
>>> uuid.uuid4().hex
'4f5359ef73f74663897ee9e903be51bc'
>>> str(uuid.uuid4().hex)
'a4d43e241bd640d08098f46fa8dcdab8'
>>> uuid.uuid4().hex
'7259e167589946659a5f0288f32c2d53'
>>>
```

rpc-auth/server/backend.yaml
nicolasochem: It's important to set resources once we are in production and have metrics collection in place. The main benefit is to have sensible autoscaling behaviour. In a minikube context it matters less since there is only one node anyway.

rpc-auth/server/index.py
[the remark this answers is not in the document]
harryttd: I don't think at this point the Flask app needs to know its public url. Whatever host the client requests from in order to get a secret url in the first place is the host of the Flask app. They couldn't talk to our app without knowing the host's address. Then we can just use that host as the host of the secret url returned to the client.

rpc-auth/server/index.py
```diff
@@ -0,0 +1,135 @@
...
+    if not is_valid_signature:
+        return "Unauthorized", 401
+
+    secret_url = generate_secret_url(public_key)
+    return secret_url
+
+
+@app.route("/tezos-node-rpc/<access_token>/<path:rpc_endpoint>",
+           methods=["GET", "POST", "PATCH", "DELETE", "PUT"])
+def rpc_passthrough(access_token, rpc_endpoint):
```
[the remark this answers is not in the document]
harryttd: now using nginx auth-url annotations to handle the auth subrequest

rpc-auth/server/Dockerfile
```diff
@@ -13,18 +11,29 @@ RUN apt-get update
...
 EXPOSE 8080
 
-# Running Python Application
+FROM base AS dev
+ENV FLASK_ENV=development
 CMD ["python3", "index.py"]
+
+FROM base AS prod
+ENV FLASK_ENV=production
+CMD ["uwsgi", \
```
harryttd, line 32: based on: https://github.com/tqtezos/chefs-club-collective
- currently running as root. Set a diff user?
- more preferable config? (processes, threads, etc.)

docker/rpc-auth/server/index.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def verify_chain_id(chain_id):
+    if chain_id != get_tezos_chain_id():
+        return False
+    return True
+
+
+@cache
+def get_tezos_chain_id():
```
harryttd, line 81: it is extremely unlikely that a chain is restarted (i guess from scratch), and so the chain id would change, right?

docker/rpc-auth/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+USER appuser
+
+CMD ["uwsgi", \
+  "--http-socket", "0.0.0.0:8080", \
+  "--callable", "app", \
+  "--threads", "100", \
+  "--processes", "1", \
+  "--wsgi-file", "index.py", \
+  "--touch-reload", "index.py", \
```
harryttd, line 65: touch-reload is different than pyauto-reload as it will reload even if there are syntax errors

harryttd:
Taquito and Pytezos both work.
```js
const { TezosToolkit } = require("@taquito/taquito")
const tezos = new TezosToolkit(
  "http://192.168.64.52/tezos-node-rpc/300c4ae403f343d49595076ce7bd97da"
)
tezos.rpc.getChainId().then(console.log)
// NetXvmLTRGRTusY
```
```python
from pytezos import rpc
rpcNode = rpc.node.RpcNode("http://192.168.64.52/tezos-node-rpc/300c4ae403f343d49595076ce7bd97da")
print(rpcNode.get("chains/main/chain_id"))
# NetXvmLTRGRTusY
```

docker/rpc-auth/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+RUN pip install -r requirements.txt \
+  --no-index --find-links ./wheels \
+  && rm -rf ./wheels ./requirements.txt
+
+RUN groupadd -g 999 appuser && \
+  useradd -r -u 999 -g appuser appuser
+
+COPY --chown=appuser:appuser ./server/index.py .
+
+ARG FLASK_ENV=development
```
itkach, line 51: prod is probably a better default

docker/rpc-auth/server/index.py
itkach: I don't think there's a meaningful scenario where chain id would change

tqchain/deployment/rpc-auth.yaml
```diff
@@ -0,0 +1,174 @@
...
+              name: redis-data
+      volumes:
+        - name: redis-data
+          persistentVolumeClaim:
+            claimName: redis-pv-claim
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: rpc-auth-service
```
nicolasochem, line 81: Let's just call it `rpc-auth`?

docker/rpc-auth/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+USER appuser
+
+CMD uwsgi \
+  --http-socket 0.0.0.0:8080 \
+  --callable app \
+  --threads 100 \
+  --processes 1 \
+  --wsgi-file index.py \
+  --worker-reload-mercy 0 \
+  $(if [ "${FLASK_ENV}" = "development" ] ; then echo "--touch-reload index.py" ; else : ; fi)
```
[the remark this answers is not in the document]
itkach: @nicolasochem  looks like devspace is set up to sync files rather than rebuild images, so this makes sense

docker/rpc-auth/server/index.py
```diff
@@ -0,0 +1,167 @@
+import os
```
itkach, line 1: putting rpc-auth under `docker` directory makes no sense to me; rpc-auth is a distinct component, just like mkchain. If we are not moving it out to a separate repo then rpc-auth and mkchain need to be sibling subdirs in the project, each with its own setup.py. Right now mkchain is "dominant" with its setup.py sitting in project root and rpc-auth is shoved under the rug in a weird place.

docker/rpc-auth/Dockerfile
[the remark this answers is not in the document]
harryttd: this is true

docker/rpc-auth/server/index.py
nicolasochem: It's a container centric view of the environment. mkchain does not belong here because it's not part of the cluster. It's a helper script to generate helm chart values.

See also: https://github.com/midl-dev/tezos-on-gke/tree/master/docker

One of the things I am planning is to migrate `generateTezosConfig.py` and `import_keys.sh` into init containers in the docker folder.

Perhaps mkchain could move to the `scripts` folder that has been added in this PR?

itkach: mkchain is a software component,  so is rpc-auth. They have their own packaging, dependencies, documentation, possible usage and deployment scenarios.  I would really like to see component based structure here, perhaps even force it via putting them in separate repos (monorepo can work two, but not like this). Containers flow from that, not the other way around.

devspace.yaml
```diff
@@ -5,17 +5,44 @@ deployments:
...
     context: ./docker/zerotier
+  rpc-auth:
+    image: rpc-auth
+    dockerfile: ./docker/rpc-auth/Dockerfile
+    context: ./docker/rpc-auth
+    build:
+      docker:
+        options:
+          buildArgs:
+            FLASK_ENV: ${FLASK_ENV}
```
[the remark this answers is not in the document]
harryttd: @nicolasochem I just tried with no prompt:
```shell
❯ echo $FLASK_ENV

❯ devspace build --skip-push -b
[done] √ Done building image tezos-rpc-auth:qchX3Uf (rpc-auth)
[done] √ Done building image tezos-zerotier:KkAWwY3 (zerotier)
[done] √ Successfully built 2 images
```

How can I reproduce?

nicolasochem:
Maybe remove .devspace folder completely?

devspace.yaml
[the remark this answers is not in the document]
harryttd: @nicolasochem yes, removing .devspace prompted me. thanks

s-zeng (approved):

PR #37, opened 2020-10-30
format python code with black

itkach:


itkach:
organized imports with https://pycqa.github.io/isort/ (single line imports option, `isort -sl`)

nicolasochem (approved):

PR #41, opened 2020-11-05
Setup github actions

s-zeng:
This PR sets up some automated CI that does the following:

- Rejects all PRs / fails builds for commits that aren't formatted w/ black + don't have sorted and cleaned imports
- Runs tests (I put in a placeholder test - what matters for the sake of this PR is that the test infrastructure is there) and rejects PRs with failed tests
- Automatically pushes to TestPYPI upon creation of a release. We can change this to real pypi when we have a token to use - for now, I can supply a TestPYPI token. This job will need a PYPI_TOKEN secret, which I do not have permission to put in this repo. But we don't need the token yet if we're not tagging any releases.

I also include `./lint.sh`, which should automatically lint all python code in the repo to ensure that the lint job won't fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,108 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  release:
+    types: [created]
+
+jobs:
+  check-lint-and-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v1
+        with:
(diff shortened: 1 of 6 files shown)


nicolasochem:
lgtm, I had to pip install utilities to run lint.sh locally. can we have a special list of requirements for dev in setuptools? or just a dev-requirements.txt?

s-zeng:
I added test and lint requirements to setup.py. Now you can run
```
pip install -e .[dev]
```
when installing mkchain from the repo and it'll grab the lint tools for you

nicolasochem (approved):

PR #42, opened 2020-11-06
zerotier improvements + introduce Devspace environment

nicolasochem:
Zerotier improvements
=====================

The current zerotier implementation creates just one IP per cluster.

We are proposing a model where every tezos node gets its own ip.

Advantages:

1. it removes any routing and forwarding consideration - every node is directly connected to a virtual layer 2 switch
1. it makes no assumption on whether a tun interface can be shared between pods (which is not true on i.e. eks)

Implementation: when a zerotier network is passed, we assume zerotier is in use. We append 2 containers to the nodes: one init container which gets an ip from zerotier, and one container that runs the zerotier daemon and runs alongside the tezos node. The tezos node is configured to "think" that its ip is the zerotier ip.

We give predictable names to zerotier interfaces <CHAIN_NAME>_bootstrap. When a cluster is invited, it queries all interfaces, and uses the ones named as such as bootstrap nodes.

This removes the requirement to query the ip of the zerotier interface and pass it as an argument to the invite command.

introduce devspace environment
==============================

Zerotier container used to be in docker hub. It was based on https://github.com/leunamnauj/kubernetes-zerotier-bridge

We moved this source code into the current repo and modified it. Building the container is a prerequisite of running a private chain with zerotier enabled.

We introduce devspace to facilitate builing containers in a minikube environment. Devspace is a tool to facilitate development in a kubernetes cluster. It monitors files, and when they change, rebuilds containers and restarts pods.

README changes
==============

README gives instructions for the single-cluster setup

HOWTO is renamed to MULTICLUSTER.md and is the guide for multi-cluster including zerotier

Note
====

The invite process is made more complicated by this change, because now it requires the complete mkchain + devspace env.

This satisfies the immediate need of multiple developers collaborating on chain developent.

Future work will simplify this process. This may consist of releasing custom containers to a public registry as part of a release cycle.


docker/zerotier/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Set zerotier name"
+if [[ $(hostname) == *"bootstrap-node"* ]]; then
+    zerotier_name="${CHAIN_NAME}_bootstrap"
+    zerotier_description="Bootstrap node for chain ${CHAIN_NAME}"
+else
+    zerotier_name="${CHAIN_NAME}_node" 
+    zerotier_description="P2p node of chain ${CHAIN_NAME}"
+fi
+
+[ ! -z $NETWORK_ID ] && { sleep 5; zerotier-cli -D/var/tezos/zerotier join $NETWORK_ID || exit 1; }
```
s-zeng, line 19: What happens if $NETWORK_ID isn't set? It seems to me that the script will simply continue from here w/o having joined anything - is that acceptable?

Also, perhaps an explanation for why we're sleeping for 5 should be documented here

docker/zerotier/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+  curl -s -XPOST \
+    -H "Authorization: Bearer $ZTAUTHTOKEN" \
+    -d '{"hidden":"false","config":{"authorized":true}}' \
+    "https://my.zerotier.com/api/network/$NETWORK_ID/member/$HOST_ID"
+
+
+  echo "Waiting for a ZeroTier IP on $ZTDEV interface... Accept the new host on my.zerotier.com"
+done
+
+echo "Set zerotier name"
```
s-zeng, line 43: Why do we set the zerotier name again here?

docker/zerotier/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+  -H "Authorization: Bearer $ZTAUTHTOKEN" \
+  -d "{\"name\":\"${zerotier_name}\",\"description\":\"$zerotier_description\"}" \
+  "https://my.zerotier.com/api/network/$NETWORK_ID/member/$HOST_ID"
+
+zerotier-cli -D/var/tezos/zerotier -j listnetworks > /var/tezos/zerotier_data.json
+
+# fetch all the network members
+curl -s -XGET \
+  -H "Authorization: Bearer $ZTAUTHTOKEN" \
+  "https://my.zerotier.com/api/network/$NETWORK_ID/member" > /var/tezos/zerotier_network_members.json
```
s-zeng, line 61: What happens when more members join? Does this json file ever get updated, and if not, does it matter?

MULTICLUSTER.md
```diff
@@ -149,45 +153,24 @@ mkchain create --number-of-nodes 3 $CHAIN_NAME | kubectl apply -f -
...
+Send this file to the recipients you want to invite.
 
 ### on the computer of the joining node
 
-The member can apply the shared yaml to their Minikube cluster:
+The member needs to follow the installation instructions, put the invite file in the mkchain directory, then run:
 
 ``` shell
-kubectl apply -f join-$CHAIN_NAME.yaml
+devspace dev -n tqtezos --var=CHAIN_NAME=$CHAIN_NAME --var=ACTION=invite
```
harryttd, line 167: I'm not so sure we should make invitees dependent on devspace. I think it is a tool meant for us as devs to quickly iterate on k8s development. Now that I'm saying this, I'm not sure we should make the chain initiators dependent on it either. Shouldn't we just be providing a tool (mkchain) that gives the desired end product, i.e. a private chain? I don't believe we should be mixing any development context into it. What is everyone's thoughts?

harryttd: On a separate note, when the docker image is building on my mac, I keep seeing what looks like errors in the `make` step (the image eventually built though):
```
g++ -O3 -fstack-protector -fPIE -Wall -Wno-deprecated -std=c++11 -pthread  -DNDEBUG -DZT_USE_MINIUPNPC -DMINIUPNP_STATICLIB -DMINIUPNPC_SET_SOCKET_TIMEOUT -DMINIUPNPC_GET_SRC_ADDR -D_BSD_SOURCE -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -DOS_STRING=\"Linux\" -DMINIUPNPC_VERSION_STRING=\"2.0\" -DUPNP_VERSION_STRING=\"UPnP/1.1\" -DENABLE_STRNATPMPERR -DZT_BUILD_PLATFORM=1 -DZT_BUILD_ARCHITECTURE=2 -DZT_SOFTWARE_UPDATE_DEFAULT="\"disable\"" -D_MT_ALLOCATOR_H -D_POOL_ALLOCATOR_H -D_EXTPTR_ALLOCATOR_H -D_DEBUG_ALLOCATOR_H -DZT_USE_X64_ASM_SALSA2012 -DZT_USE_FAST_X64_ED25519   -c -o node/Trace.o node/Trace.cpp
In file included from node/Node.hpp:32,
                 from node/Peer.hpp:23,
                 from node/Topology.hpp:30,
                 from node/Switch.hpp:28,
                 from node/Trace.cpp:21:
node/Network.hpp: In constructor 'ZeroTier::Network::_IncomingConfigChunk::_IncomingConfigChunk()':
node/Network.hpp:422:70: warning: 'void* memset(void*, int, size_t)' clearing an object of type 'struct ZeroTier::Network::_IncomingConfigChunk' with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
  422 |   _IncomingConfigChunk() { memset(this,0,sizeof(_IncomingConfigChunk)); }
      |                                                                      ^
node/Network.hpp:420:9: note: 'struct ZeroTier::Network::_IncomingConfigChunk' declared here
  420 |  struct _IncomingConfigChunk
      |         ^~~~~~~~~~~~~~~~~~~~
```
It seems to repeat many times.

Also, once everything is up and running, devspace cannot connect to the pod `[fatal]  start sync: Unable to start sync: Error selecting pod: Couldn't find a pod with selector app=tezos-bootstrap-node in namespace tqtezos`

docker/zerotier/entrypoint.sh
nicolasochem: mkchain only appends the zerotier container and initcontainer if NETWORK_ID is set.

I am unsure why we are sleeping, it is pre-existing code from https://github.com/leunamnauj/kubernetes-zerotier-bridge

We were already using this code, it was just hidden in a docker hub artifact.

docker/zerotier/entrypoint.sh
nicolasochem: The json file gets updated only when the pod restarts. The only purpose of this list is to help the invited nodes find a bootstrap node. Once a bootstrap node is found, the p2p network will ensure constant connectivity to every node, so it's no longer needed.

docker/zerotier/entrypoint.sh
nicolasochem: We use zerotier name field as a means for other nodes to find their bootstrap node without the need to pass the zerotier internal ip of the bootstrap node to the invite chain.

MULTICLUSTER.md
itkach: I agree, devspace or other dev dependencies should not be imposed on invitee

harryttd:
I didn't investigate fully but I'm having trouble scaling nodes. 
- I run devspace as the first step. 
- I then run mkchain with --number-of-nodes. 
```
❯ mkchain create --number-of-nodes 2 $CHAIN_NAME | k apply -f -

namespace/tqtezos unchanged
secret/tezos-secret unchanged
configmap/tqtezos-utils unchanged
configmap/tezos-config unchanged
statefulset.apps/tezos-node configured
job.batch/activate-job unchanged
service/tezos-bootstrap-node-rpc unchanged
service/tezos-bootstrap-node-p2p unchanged
deployment.apps/tezos-bootstrap-node configured
persistentvolumeclaim/tezos-bootstrap-node-pv-claim unchanged
configmap/zerotier-config unchanged
```
Notice that the deployment is also configured instead of unchanged. It shouldn't be configured as the yaml is the exact same for it. The only change is for the StatefulSet replica amount. The bootstrap node is terminated and redeployed. And then I get `Failed to pull image "tezos-zerotier"` errors for both bootstrap and peer nodes. My guess is this has something to do with devspace, and running mkchain separately from it is causing issues.

MULTICLUSTER.md
nicolasochem: The error `[fatal] start sync: Unable to start sync: Error selecting pod: Couldn't find a pod with selector app=tezos-bootstrap-node in namespace tqtezos`: is due to the tezos container not being built by devspace. Since devspace can't find any container that it built, it's not able to sync.

What I can do for now is edit the guide to just have user do `devspace build` instead of `devspace dev`. Essentially, use devspace as the container build and push utility. The deploy instructions will remain `mkchain | kubectl`. `devspace dev` remains available but may throw errors.

We will have a release pipeline that builds and pushes these custom containers to a public registry, so tagged release versions of tezos-k8s will not have `devspace build` as a requirement (neither create nor invite). But we need to merge the code before we release, so the updated deployment instructions reflect this intermediary state.

Does this sound acceptable?

As for the zerotier compilation warnings, this is carried over from the upstream code, can we leave it aside for now?

tqchain/deployment/node.yaml
```diff
@@ -50,7 +50,7 @@ spec:
       - imagePullPolicy: Always
         name: tezos-config-generator
         image: python:alpine
-        command: ["python", "/opt/tqtezos/generateTezosConfig.py"]
+        command: ["python", "/opt/tqtezos/generateTezosConfig.py", "--generate-config-json", "--generate-parameters-json"]
```
harryttd, line 53: Please remind me, do the tezos peer nodes need parameters.json? I thought that is only something the bootstrap node needs as it is initiating the chain.

MULTICLUSTER.md
harryttd: >The error [fatal] start sync: Unable to start sync: Error selecting pod: Couldn't find a pod with selector app=tezos-bootstrap-node in namespace tqtezos: is due to the tezos container not being built by devspace. Since devspace can't find any container that it built, it's not able to sync.

It's weird, when i run `devspace dev -b` it can't find the pod. Without `-b` it does and it connects to the zerotier container.

I think the rest makes sense.

tqchain/deployment/node.yaml
nicolasochem: I tried and you are correct. The code has been fixed to not generate parameters.json for non-bootstrap nodes.

MULTICLUSTER.md
nicolasochem: OK, I changed the instructions to do `devspace build` only. `devspce dev` is left as an undocumented buggy feature (for now). Please review again.

nicolasochem:
@harryttd reported this but we were not able to replicate:

```
2020-11-10T19:16:55.360838791Z Starting tezos config file generation
2020-11-10T19:16:55.360851774Z Traceback (most recent call last):
2020-11-10T19:16:55.360941645Z   File "/opt/tqtezos/generateTezosConfig.py", line 245, in <module>
2020-11-10T19:16:55.361106380Z     main()
2020-11-10T19:16:55.361269399Z   File "/opt/tqtezos/generateTezosConfig.py", line 44, in main
2020-11-10T19:16:55.361414775Z     bootstrap_peers.extend(get_zerotier_bootstrap_peer_ips())
2020-11-10T19:16:55.361557625Z   File "/opt/tqtezos/generateTezosConfig.py", line 66, in get_zerotier_bootstrap_peer_ips
2020-11-10T19:16:55.361801894Z     return [
2020-11-10T19:16:55.361837276Z   File "/opt/tqtezos/generateTezosConfig.py", line 67, in <listcomp>
2020-11-10T19:16:55.362133371Z     n["config"]["ipAssignments"][0]
2020-11-10T19:16:55.362284370Z IndexError: list index out of range
```

Let's not do anything about it now and wait for it to happen again.

harryttd:
Just noting it may be due to zerotier maxing out amount of nodes. I deleted my old network and made a new one and everything worked.

harryttd (approved):

s-zeng (approved):

PR #44, opened 2020-11-14
Release pipeline

nicolasochem:
To release mkchain, create a Github release with PEP440 compliant release tag (for example `1.0b1`, `1.0`)

The following will happen:

* PyPI release of the package
* all containers in `docker` folder will be pushed to docker hub

Example trigger of the container build pipeline:

https://github.com/tqtezos/tezos-k8s/actions/runs/363520964

The `image` field for custom containers in the k8s yaml output will point to the corresponding version on docker hub. There is no need for end users to build containers: sending a k8s yaml as private chain invite will just work. This applies to pip installs from pypi, pip installs of a tagged github url, and manual cloning of a tagged version followed by `pip install -e .`.

We are not using docker hub build pipelines, instead using github actions. This is more sustainable as it reduces the amount of work needed to migrate to another registry.

This was done with a python utility named [versioneer](https://github.com/python-versioneer/python-versioneer). versioneer.py and tqchain/_version.py are files from this utility, and must be committed in our source tree for this to work.

Also added `mkchain --version`

When this PR is approved and merged, I will release 1.0b1


s-zeng (commented):
Looks good to me - have you ran the release pipelines before (perhaps on testpypi / test equiv for dockerhub) to verify that they work?

nicolasochem:
I did run the docker hub pipeline succesfully (link in PR description) but not pypi. We will get one "free" test with 1.0b1 release.

s-zeng (approved):

tqchain/mkchain.py
```diff
@@ -8,6 +8,11 @@
 import sys
 from datetime import datetime, timezone
 
+sys.path.insert(0, "tqchain")
+from _version import get_versions
```
harryttd, line 12: Should this be "dunder" version or just "under"? Everywhere else is referencing "dunder".

harryttd: I think i commented in the wrong spot. The references on line 148 and 174. I'm getting not defined errors

s-zeng, line 12: On this note, this import won't resolve unless it's relative. So should probably correct to `from ._version import get_versions`

nicolasochem: Versioneer added the `_version.py` file and it's necessary to import it by name to be able to get the current version in mkchain code, see:

https://github.com/python-versioneer/python-versioneer#version-identifiers

harryttd (approved):

tqchain/mkchain.py
nicolasochem: Done, thanks.

PR #47, opened 2020-11-17
Implement helm chart

s-zeng:
Mostly complete helm PR. With the introduction of optional zerotier containers and init containers present across the bootstrap nodes and the statefulset nodes, I felt that Dhall's inability to create non-string Go templates made it too unwieldy to implement the {{ if }} statements needed for those. Until dhall-lang/dhall-haskell#2040 is merged, there simply is no ergonomic way to create generic helm charts with Dhall. It is in a too immature state for us right now.

As such, this is a full helm implementation of mkchain, with the former mkchain command being relegated to only a values.yaml generating program. The helm chart handles all of the configuration.

Todo (perhaps not in this PR):
- Deploy the helm chart to a chart repo


tezos-helm/values.yaml
```diff
@@ -0,0 +1,11 @@
+additional_nodes: 0
+baker: true
+baker_command: tezos-baker-006-PsCARTHA
+bootstrap_mutez: "4000000000000"
+bootstrap_peers: ["tezos-bootstrap-node-p2p:9732"]
```
nicolasochem, line 5: so now bootstrap_peer has to be part of the values.yaml for the bootstrap cluster? previously it wasn't. Is it because there was some "intelligence" in mkchain that can't be transfered to helm?

tezos-helm/templates/configs.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: zerotier-config
+  namespace: tqtezos
+---
+apiVersion: v1
+data:
+{{- if .Values.is_invitation }}
+  KEYS_TYPE: cHVibGlj
+{{- else }}
+  KEYS_TYPE: c2VjcmV0
```
nicolasochem, line 32: it would probably be cleaner to add public/private attribute to keys in values.yaml

nicolasochem (commented):
Thanks for this,

I generated values.yaml and this is what I saw for keys:

```keys:
  BOOTSTRAP_ACCOUNTS: !!binary |
    WW1GclpYSWdZbTl2ZEhOMGNtRndYMkZqWTI5MWJuUmZNU0JpYjI5MGMzUnlZWEJmWVdOamIzVnVk
    Rjh5SUdkbGJtVnphWE09
  baker_secret_key: !!binary |
    WldSemF6TklRM05YZUVwbVRuUnhkVVZ0U0dWSFZXcE1kekZtVFRWcmRrRjBhREpGYzJaRlZHTjBO
    bVZYUTFGb1ozTkROamR2
  bootstrap_account_1_secret_key: !!binary |
    WldSemF6TTFhVzVrVmxCVlkydERSbWR2WmxsNlFsVXhkREZpY1ZoRlprWm1PV1owZDJOUk1XZEVU
    VW81YzNWV05GVkNaRlUx
  bootstrap_account_2_secret_key: !!binary |
    WldSemF6Uk5URlExT0haSGRXSkhiVFpaZUUxTmR6bERUR2cyT0ROelExUklObVpaUzBSNmVYZG9V
    a1ZLTkVWRFpGUjRXV1Za
  genesis_secret_key: !!binary |
    WldSemF6Uklhek5EWWpoSFVETkdTalE0UkU1UWNXSktaVGxMYW5STVIybzVjMFpGWTBkR1ZsTTBS
    VzgyUVhacFkwaFRUVEpM
```
What is BOOTSTRAP_ACCOUNTS doing here?

s-zeng:
> Thanks for this,
> 
> I generated values.yaml and this is what I saw for keys:
> 
> ```
>   BOOTSTRAP_ACCOUNTS: !!binary |
>     WW1GclpYSWdZbTl2ZEhOMGNtRndYMkZqWTI5MWJuUmZNU0JpYjI5MGMzUnlZWEJmWVdOamIzVnVk
>     Rjh5SUdkbGJtVnphWE09
>   baker_secret_key: !!binary |
>     WldSemF6TklRM05YZUVwbVRuUnhkVVZ0U0dWSFZXcE1kekZtVFRWcmRrRjBhREpGYzJaRlZHTjBO
>     bVZYUTFGb1ozTkROamR2
>   bootstrap_account_1_secret_key: !!binary |
>     WldSemF6TTFhVzVrVmxCVlkydERSbWR2WmxsNlFsVXhkREZpY1ZoRlprWm1PV1owZDJOUk1XZEVU
>     VW81YzNWV05GVkNaRlUx
>   bootstrap_account_2_secret_key: !!binary |
>     WldSemF6Uk5URlExT0haSGRXSkhiVFpaZUUxTmR6bERUR2cyT0ROelExUklObVpaUzBSNmVYZG9V
>     a1ZLTkVWRFpGUjRXV1Za
>   genesis_secret_key: !!binary |
>     WldSemF6Uklhek5EWWpoSFVETkdTalE0UkU1UWNXSktaVGxMYW5STVIybzVjMFpGWTBkR1ZsTTBS
>     VzgyUVhacFkwaFRUVEpM
> ```
> 
> What is BOOTSTRAP_ACCOUNTS doing here?

 The k8s manifest that mkchain/this helm chart makes passes on the bootstrap_accounts list by base64 encoding it, so here in this values.yaml I'm having mkchain pre-base64 encoding it and passing the encoded value. Now if bootstrap_accounts was a constant and we always expect the same 4 bootstrap accounts, we can hardcode this directly into the helm template. But I seem to recall we do expect the capability to change the bootstrap accounts at some point in time so this is the mechanism I'm using to allow that.

If you like, I can make it more readable by having bootstrap_accounts be a yaml list in the values.yaml (i.e.
```yaml
bootstrap_accounts:
  - baker
  - bootstrap_account_1
  - bootstrap_account_2
  - genesis
```
and then instead of joining and encoding in mkchain, we can include it in the helm template with:
```yaml
{{- join " " .Values.bootstrap_accounts | b64enc }}
```

Let me know which option you prefer

README.md
```diff
@@ -32,16 +34,14 @@ You can modify these parameters by:
 
 | YAML Parameter | mkchain argument | Description | Default |
 | ----- | ----------- | ------ | ----- |
-| number_of_nodes | --number-of-nodes |  Number of peers in the cluster | 1 |
+| additional_nodes | --additional-nodes |  Number of non-bootstrap nodes in the cluster | 0 |
```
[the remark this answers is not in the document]
s-zeng: We avoid arithmetic in the helm template. This is more a relic from my dhall PR where doing this made sense due to the type safety of the Natural type (integers >= 0). Let me know if you want me to revert this

tezos-helm/values.yaml
```diff
@@ -0,0 +1,11 @@
+additional_nodes: 0
```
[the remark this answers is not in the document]
s-zeng: I wasn't sure which place was better to put the defaults (mkchain or the default values.yaml) so I put it in both for this PR. Let me know what you think makes more sense

tezos-helm/values.yaml
s-zeng: Formerly mkchain had a bootstrap_peer argument that it passed onto the _chain.yaml files only if specified. Here I'm just making it explicit

tezos-helm/templates/configs.yaml
s-zeng: The is_invitation flag is also used for deciding whether or not the bootstrap stuff gets installed, so I figured might as well reuse it here too. I can move the KEYS_TYPE generation back to mkchain and have it in the keys dictionary if you like

MULTICLUSTER.md
```diff
@@ -85,12 +85,12 @@ Set unbuffered IO for python:
 PYTHONUNBUFFERED=x
 ```
 
-Run the following command to create the chain yaml and feed it to Kubernetes:
+Run the following command to create the helm configuration and feed it to helm:
 
 ``` shell
-mkchain generate-constants --zerotier-network $ZT_NET \
+mkchain -constants --zerotier-network $ZT_NET \
```
harryttd, line 91: i think `-constants` might be a typo

tezos-helm/values.yaml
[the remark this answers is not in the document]
nicolasochem: default values.yaml makes more sense. it will be a helm-first design. mkchain can just use it as source of truth for defaults.

MULTICLUSTER.md
s-zeng: Good catch

nicolasochem:
After zoom discussion, this would be a good structure for values.yaml :

```
#additional_nodes: 0
#baker: true
baker_command: tezos-baker-006-PsCARTHA
bootstrap_peers:
- tezos-bootstrap-node-p2p:9732
chain_name: helmdemo
is_invitation: false
zerotier_in_use: false
container_images:
  zerotier_docker_image: tezos-zerotier:dev
  tezos_docker_image: tezos/tezos:v7-release
protocol:
  bootstrap_mutez: '4000000000000'
  hash: PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb
genesis:
  genesis_chain_id: BL25uCY5x3ToPmz6wa2xG6nWkmvScRJu6freYKbhDBoesLg54Yz
  bootstrap_timestamp: '2020-11-18T07:00:45.937295+00:00'
accounts:
  - name: baker
    key: !!binary |
    WldSemF6TklRM05YZUVwbVRuUnhkVVZ0U0dWSFZXcE1kekZtVFRWcmRrRjBhREpGYzJaRlZHTjBO
    bVZYUTFGb1ozTkROamR2
    private: true
    bootstrap: true
    baker: true
  - name: bootstrap_account_1
    key: !!binary |
    WldSemF6TTFhVzVrVmxCVlkydERSbWR2WmxsNlFsVXhkREZpY1ZoRlprWm1PV1owZDJOUk1XZEVU
    bVZYUTFGb1ozTkROamR2
    private: true
    bootstrap: true
nodes:
  - {}
  - bake_for: baker
# TODO
# archive nodes, node with open rpc
# take into account fundraisers accounts that people need to activate
```

We can leave the multiple baker feature for another PR but let's make sure we are ready to add it.

tezos-helm/templates/configs.yaml
```diff
@@ -0,0 +1,87 @@
...
+    {{- if $account.bootstrap }}
+{{ $i }}: {{ $account.name | quote }}
+    {{- end }}
+  {{- end }}
+{{- end -}}
+
+apiVersion: v1
+data:
+  BOOTSTRAP_ACCOUNTS: !!binary |
+{{ include "bootstrapAccounts" . | fromYaml | values | join " " | b64enc | b64enc | indent 4 }}
```
s-zeng, line 64: So to replicate the original mkchain behaviour, we have to `b64enc` twice. Does anyone know why that is the case?

harryttd: Also, it is being encoded already in mkchain [here](https://github.com/tqtezos/tezos-k8s/blob/s-zeng%2Fhelm/tqchain/mkchain.py#L39)

README.md
```diff
@@ -32,28 +34,20 @@ You can modify these parameters by:
...
 | zerotier_network | --zerotier-network | Zerotier network id for external chain access | |
 | zerotier_token | --zerotier-token | Zerotier token for external chain access | |
 | bootstrap_peer | --bootstrap-peer | peer ip to join | |
-| genesis_key | --genesis-key | genesis public key for the chain to join | |
-| genesis_block | --genesis-block | hash of the genesis block | |
-| timestamp | --timestamp | timestamp for the chain to join | |
-| protocol_hash | --protocol-hash | Desired Tezos protocol hash | PsDELPH1Kxsxt8f9eWbxQeRxkjfbxoqM52jvs5Y5fBxWWh4ifpo |
-| baker_command | --baker-command | The baker command to use, including protocol | tezos-baker-007-PsDELPH1 |
-| rpc_auth | --rpc-auth | Include an RPC authentication server | |
+| rpc_autch | --rpc-auth | whether or not an RPC auth server will be spun up | False |
```
harryttd, line 41: should be rpc_auth

Also, why are so many of the other args deleted?

README.md
```diff
@@ -32,28 +34,20 @@ You can modify these parameters by:
...
 ## private chain
 
 ### create a self-contained chain
 $CHAIN_NAME: is your private chain's name
 
 ``` shell
-mkchain generate-constants $CHAIN_NAME
-mkchain create $CHAIN_NAME | kubectl apply -f -
+mkchain $CHAIN_NAME
+helm install $CHAIN_NAME tezos-helm/ -f ${CHAIN_NAME}_values.yaml
```
harryttd, line 50: should be `helm install $CHAIN_NAME tezos-helm/ -f ${CHAIN_NAME}_values.yaml --namespace tqtezos --create-namespace` like in multicluster.md. I think i forgot to add in my branch. The duplication will get cleaned up with code and doc reorganization.

tqchain/mkchain.py
```diff
@@ -1,33 +1,24 @@
 import argparse
```
nicolasochem, line 1: The help should be changed to say something like "Generate helm values for a tezos private chain helm chart"

tezos-helm/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+    bootstrap: true
+    baker: true
+  - name: sample2
+    key: !!binary |
+      SAMPLE==
+    private: true
+    bootstrap: true
+nodes:
+  - {}
+  - bake_for: baker
```
nicolasochem, line 38: What I meant here is "one node that bakes for the account named baker, and another node that does not bake", but when I launched the helm chart with default values, I saw only one node.

tezos-helm/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: tezos-chain
+description: A Helm chart for Kubernetes
```
nicolasochem, line 3: "a helm chart for tezos nodes on kubernetes"

README.md
```diff
@@ -17,10 +19,10 @@ Your chain is uniquely defined by a set of values such as bootstrap account keys
 Create these values:
 
 ``` shell
-mkchain generate-constants $CHAIN_NAME
+mkchain $CHAIN_NAME
 ```
 
-It will create two 2 yaml files, `<$CHAIN_NAME>_chain.yaml` and `<$CHAIN_NAME>_chain_invite.yaml`.
+It will create two 2 yaml files, `<$CHAIN_NAME>_values.yaml` and `<$CHAIN_NAME>_invite_values.yaml`.
```
nicolasochem, line 25: additional references to `generate_constants` must be removed.

MULTICLUSTER.md
```diff
@@ -120,26 +119,28 @@ chain running one node.
...
+a number of nodes of your choice by passing `--additional-nodes` to
 mkchain generate-constants.
 
 The nodes will establish peer-to-peer connections in a full mesh topology.
 
 If you previously spun up the chain with just one node, you may scale
-up your setup to an arbitrary number of nodes by overriding the --number-of-nodes
-parameter in the `create` command:
+up your setup to an arbitrary number of nodes by overriding the
+--additional-nodes parameter in the values yaml:
```
nicolasochem, line 129: does not work yet?

nicolasochem (changes requested):

README.md
s-zeng: Good catch
For args deleting, I deleted most things that have a default in the values.yaml - I think we want editing the values.yaml to be the main paradigm for customizing.

nicolasochem: I thought that in the helm world, people expect a list of parameters in the readme, with defaults, for example

https://github.com/helm/charts/tree/master/stable/prometheus-operator

s-zeng: Perhaps that's something we can have in the helm chart readme, not in mkchain's readme

README.md
s-zeng: Addressed

tqchain/mkchain.py
s-zeng: Done

tezos-helm/Chart.yaml
s-zeng: Done

README.md
s-zeng: Done

tezos-helm/values.yaml
s-zeng: To clarify, what was the final consensus on what we wanted here again? Did we want `number_of_nodes` and `number_of_bakers` values or did we want the list of nodes?

README.md
harryttd: not sure change was pushed

tezos-helm/values.yaml
nicolasochem: I think we want a list of nodes in values.yaml, and `--number-of-nodes` `--number-of-bakers` as a mkchain parameter, used to compose this list of nodes.

But this should be part of a follow up PR. for now we need `--number-of-nodes` in mkchain parameter, defaulting to 1, and if it is more than one, add extra empty dict nodes to the list.

nicolasochem:
Please remove mentions of `additional_nodes` from values.yaml and multicluster.md and we are good, I think.

tezos-helm/values.yaml
harryttd: this default will override any ZT ip's with the current logic of generateTezosConfig.py. I will remove this from the values.yaml and have generateTezosConfig.py add in the service if ZT is not use. Also, I will make the creation of the service in helm conditional if ZT is not in use.

nicolasochem (approved):

harryttd (approved):

PR #49, opened 2020-11-18
Set image pull policies to IfNotPresent

harryttd:
I hit the docker pull limit (somehow). Setting explicitly to IfNotPresent as I was experiencing attempted image pulls for images that I have locally. (might be a minikube issue)


nicolasochem (approved):
I tried and it works.

PR #50, opened 2020-11-18
Use Delphi protocol by default in mkchain

### LINKED ISSUE #48: Move mkchain to Delphi

harryttd:
closes #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ You can modify these parameters by:
 | genesis_key | --genesis-key | genesis public key for the chain to join | |
 | genesis_block | --genesis-block | hash of the genesis block | |
 | timestamp | --timestamp | timestamp for the chain to join | |
-| protocol_hash | --protocol-hash | Desired Tezos protocol hash | PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb |
-| baker_command | --baker-command | The baker command to use, including protocol | tezos-baker-006-PsCARTHA |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -42,8 +42,8 @@ You can modify these parameters by:
 | genesis_key | --genesis-key | genesis public key for the chain to join | |
 | genesis_block | --genesis-block | hash of the genesis block | |
 | timestamp | --timestamp | timestamp for the chain to join | |
-| protocol_hash | --protocol-hash | Desired Tezos protocol hash | PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb |
-| baker_command | --baker-command | The baker command to use, including protocol | tezos-baker-006-PsCARTHA |
+| protocol_hash | --protocol-hash | Desired Tezos protocol hash | PsDELPH1Kxsxt8f9eWbxQeRxkjfbxoqM52jvs5Y5fBxWWh4ifpo |
+| baker_command | --baker-command | The baker command to use, including protocol | tezos-baker-007-PsDELPH1 |
```
harryttd, line 46: I think we can leave for another PR, but if we have a default baker command, these two lines need to be dynamically using a protocol: [here](https://github.com/tqtezos/tezos-k8s/blob/55dbac7f3395660a9e516ee3a0cd4b1123478856/tqchain/mkchain.py#L44) and [here](https://github.com/tqtezos/tezos-k8s/blob/55dbac7f33/tqchain/utils/import_keys.sh#L7)

harryttd:
Works on `tezos/tezos:v8-release`.
Worth to mention that I also tested on master `tezos/tezos:master_08d3405e_20201113152010`. That is the version that contains the fix for the tezos-client --endpoint flag. Was getting a bunch of errors in activate-job. "Fitness too low", and "block in the future".

nicolasochem (approved):
It works.

PR #55, opened 2020-11-30
Dynamically set namespace value

harryttd:
Helm would not install easily if the namespace was defined in the manifest yaml file.
In general it seems the Helm way is to pass the namespace as an argument when installing.
 https://stackoverflow.com/a/55378654


s-zeng (approved):
Looks good. Thanks for the patch!

PR #59, opened 2020-12-14
Restructure codebase

### LINKED ISSUE #53: Re-organize the repo around components
This repo started life as a single component repo dedicated to mkchain. 
Now it holds several different, distinct components, each with its own dependencies, documentation, dev setup etc. It would help to re-organize the repo to reflect this, with some content also living at the root of the project to bring it all together - top-level README and perhaps a Makefile and/or devspace.yaml and or other artifacts.

```
❯ tree
.
├── LICENSE
├── Makefile
├── README.md
├── devspace.yaml
├── mkchain
│   ├── MANIFEST.in
│   ├── README.md
│   ├── doc
│   ├── mkchain-devspace.yaml
│   ├── scripts
│   ├── setup.cfg
│   ├── setup.py
│   ├── test
│   ├── tqchain
│   │   ├── __pycache__
│   │   │   ├── _version.cpython-39.pyc
│   │   │   └── mkchain.cpython-39.pyc
│   │   ├── _version.py
│   │   ├── deployment
│   │   │   ├── bootstrap-node.yaml
│   │   │   ├── common.yaml
│   │   │   ├── node.yaml
│   │   │   ├── rpc-auth.yaml
│   │   │   └── zerotier.yaml
│   │   ├── mkchain.py
│   │   └── utils
│   │       ├── generateTezosConfig.py
│   │       └── import_keys.sh
│   └── versioneer.py
├── rpc-auth
│   ├── Dockerfile
│   ├── requirements.txt
│   ├── rpc-auth-client.sh
│   ├── scripts
│   ├── server
│   │   └── index.py
│   └── test
└── zerotier
    ├── Dockerfile
    ├── entrypoint.sh
    └── supervisor-zerotier.conf
```

### LINKED ISSUE #54: Re-organize documentation
Currently documentation is split across several files, such as README.md, MULTICLUSTER.md, DEVELOPMENT.md and sequence for successful setup is not sequential reading. Lets consolidate and tidy this up to be a coherent set of docs, also taking into account component-based structure that will be implemented in #53

### LINKED ISSUE #60: github ci.yaml needs to be updated due to codebase restructure
Code restructure: #59

### LINKED ISSUE #62: Package and publish Helm chart
For now I am going to publish to github pages.

### LINKED ISSUE #63: Update mkchain default tezos image to v8

### LINKED ISSUE #65: Certain values specified in Helm values.yaml should be more consistent
e.g., the `container_images` prop is itself too long as well as the image names. The prop should really be `images`. Images themselves can just be `zerotier`, `tezos`, `rpc-auth`. Instead of `zerotier_docker_image`, `rpc_auth_image`, `tezos_docker_image`.

Making them consistent also helps with the CI getting the values so they can be updated as versions are bumped.

EDIT: Helm does not allow `-` in values. (It actually might be a Go templating issue). Will use `_`

harryttd:
This PR restructures the codebase.

It makes mkchain, helm charts (tezos, and rpc-auth), and all docker images separate components.

Also adds/edits a lot of documentation. The root level readme describes for users what they need to install and the steps to get a chain started. `/mkchain` has its own readme which describes what it does and how to install it. `/rpc-auth` also has a readme describing what it does and how to get it running in your cluster. I did not add a readme to `/zerotier`. The root level readme describes what the purpose of ZT is.

A lot of devspace updates to help with the dev cycle. Example being auto reload when your values yaml file changes.

seems i need to specify close for each one......
closes #65
closes #63 
closes #62 
closes #60 
closes #54 
closes #53


charts/tezos/utils/generateTezosConfig.py
```diff
@@ -36,12 +36,14 @@ def main():
 
     if main_args.generate_config_json:
         net_addr = None
-        bootstrap_peers = CHAIN_PARAMS["bootstrap_peers"]
+        bootstrap_peers = getattr(CHAIN_PARAMS, "bootstrap_peers", [])
```
s-zeng, line 39: I believe that this will always return an empty list, since CHAIN_PARAMS is from json.loads and json.loads only returns dicts, lists, and primitives, which don't have named attributes. Maybe `CHAIN_PARAMS.get("bootstrap_peers", [])` is what you want

harryttd: Thank you I will confirm this!

nicolasochem:
When using with `rpc_auth: False`, I had to delete the dependency of the tezos chart on the rpc-auth chart otherwise I get such message (both with devspace and without):

```
[fatal]  error deploying: Error deploying chain: Unable to deploy helm chart: error during 'helm upgrade chain ./charts/tezos --namespace default --values /tmp/921905247 --install --kube-context minikube': Release "chain" does not exist. Installing it now.                                                              
Error: found in Chart.yaml, but missing in charts/ directory: authenticated-rpc-tezos-chain                                                                                                                                                                                                                                   
 => exit status 1                                                                                                                                                                                                                                                                                                             
Run `devspace purge -d chain` and `devspace deploy` to recreate the chart                                         
```

Same with rpc-auth true in values.yaml and with `-v rpc-auth` passed.

harryttd:
@nicolasochem I believe you need to run `helm dependency update charts/tezos`. This packages rpc-auth into the tezos chart. Then you can run `devspace dev --var=CHAIN_NAME="$CHAIN_NAME" -p rpc-auth` to have devspace also spin up rpc-auth. 

(i wouldn't run `mkchain --rpc-auth` if you are using devspace. Right now it will try applying the same manifest twice bec `rpc_auth: true` in values yaml. It's a kink I have an idea of how to solve. Still working on cleaning up docs and adding new info about restructured codebase to explain everything)

harryttd:
After speaking with @nicolasochem , we decided to open the PR although some documentation is still being worked on. That should be finished obviously before the release

mkchain/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+To generate these values run:
+
+```shell
+mkchain $CHAIN_NAME
+```
+
+This command will create two yaml files in your current working directory. `<$CHAIN_NAME>_values.yaml` and `<$CHAIN_NAME>_invite_values.yaml`.
+
+### mkchain parameters
```
nicolasochem, line 57: These parameters should be autogenerated from helm values.yaml during release, so we are not sure that the mkchain parameters and helm config do not drift other time. We had that in the custom templating model, and we lost it in this branch. This should however be fixed in a follow-up PR.

nicolasochem (approved):
I have some reservations about moving all the init containers to top-level directories. I understand that every one of them should be in a ready state to be separated into its own repo, but in some cases it feels that the containers are doing something trivial. However, this should be addressed prior to release and should not prevent us from merging this PR.

We have identified a bug where nodes sometimes don't connect to the p2p service, but I do not think anything in this PR caused this behavior to change. I have another work in progress where I completely remove the p2p service which I think will make things wlookork better.

Otherwise, I have used this branch a lot (having merged some PRs into it) and it works well for me. We should merge and address the remaining issues that we are aware of prior to next release.

harryttd:
> I have some reservations about moving all the init containers to top-level directories. I understand that every one of them should be in a ready state to be separated into its own repo, but in some cases it feels that the containers are doing something trivial. However, this should be addressed prior to release and should not prevent us from merging this PR.

I agree we should discuss

> We have identified a bug where nodes sometimes don't connect to the p2p service, but I do not think anything in this PR caused this behavior to change. I have another work in progress where I completely remove the p2p service which I think will make things wlookork better.

Roland and I both ran into this. I believe a recent minikube version broke something as I ran the exact same code on a separate machine with an older version of minikube and it worked.

PR #67, opened 2021-01-06
do not put binary data in values.yaml

nicolasochem:
Values.yaml should contain native tezos keys starting with `edpk` and `edsk`. This solves several issues with respect to binary encoding in yaml.


mkchain/tqchain/mkchain.py
```diff
@@ -37,9 +37,7 @@ def gen_key(image):
     ).split(b"\n")
 
     def extract_key(index: int) -> bytes:
-        return base64.b64encode(
-            keys[index].split(b":")[index].strip().decode("utf-8").encode("ascii")
-        )
+        return keys[index].split(b":")[index].strip().decode("utf-8")
```
itkach, line 40: a nitpick, but these bytes really are ascii only, so decode("ascii") is more approriate

mkchain/tqchain/mkchain.py
```diff
@@ -37,9 +37,7 @@ def gen_key(image):
     ).split(b"\n")
 
     def extract_key(index: int) -> bytes:
```
itkach, line 39: this is probably better  spelled out as a standalone (not nested function) that simply takes a bytes arg and doesn't depend on keys variable in closure - this way it becomes unit testable

charts/tezos/values.yaml
```diff
@@ -18,14 +18,14 @@ genesis:
   bootstrap_timestamp: "YOUR_BOOTSTRAP_TIMESTAMP_HERE"
 accounts:
   - name: sample1
-    key: !!binary |
+    key: |
```
harryttd, line 21: should the default values also be in form of `edsk`?

harryttd (approved):

PR #68, opened 2021-01-12
edo fix + switch kubernetes ingress to beta + switch tezos to v8

nicolasochem:
* Edo fix:

The activate job used to do 2 things: activate protocol and bake once.
At some point, it was necessary to bake once manually to kick the chain
off, but this requirement has now been lifted. After activation, new
blocks will be producing automatically.

That works for delphi but it went unnoticed. However for edo the protocol
is now more picky in the time distance between blocks so bake_once fails.

I am removing it. It works on delphi and edo still, but carthage may
be broken.

In the process, I am creating a specific container for activation, because
a bash script is easier to understand and debug than a command in a
kubectl manifest.

Per our convention, this new container goes at the top level, although
this is maybe worth discussing because it is so simple it will never
become its own project.

Also updated tezos to v8. Delphi and edo work.

* Switch kubernetes efgress to beta:

This is necessary to make ingress-nginx work on EKS.

The syntax is different (serviceName/servicePort). I tested on minikube
as well.

Also replace readWriteMany (unsupported on cloud) with readWriteOnce


charts/tezos/values.yaml
```diff
@@ -9,7 +9,8 @@ zerotier_config:
   zerotier_token: YOUR_ZEROTIER_TOKEN_HERE
 images:
   zerotier: tezos-k8s-zerotier:dev
-  tezos: tezos/tezos:v7-release
+  tezos: tezos/tezos:v8-release
+  tezos-chain-initiator: tezos-chain-initiator:dev
```
harryttd, line 13: The way I figured out for the CI to update our images with the latest release version is by regexing on the image name. It looks for `tezos-k8s` prefix. Otherwise there is no way to differentiate between our images vs other images we use, in this case the tezos image. I need to add documentation for this in DEVELOPMENT.md.

If you have another suggestion for how to go about this I'm all ears. Maybe we should split `images` into `images` and `tezos_k8s_images`? Then CI goes through the latter list.

charts/tezos/templates/configs.yaml
```diff
@@ -15,6 +15,14 @@ metadata:
   namespace: {{ .Release.Namespace }}
 ---
 apiVersion: v1
+data:
```
harryttd, line 18: Can we put protocol.hash in the config map above that already has chain params (including protocol.bootstrap_mutez) ? Is it difficult to reference a single value that is part of a json string for use as an env var to pass into the pod?

I would think it is preferable to not make more config maps if we don't really need to. Just my thought

mkchain/tqchain/mkchain.py
```diff
@@ -120,6 +120,7 @@ def main():
         "chain_name": args.chain_name,
         "images": {
             "tezos": args.docker_image,
+            "tezos_chain_initiator": "tezos-chain-initiator:dev",
```
harryttd, line 123: Bec this image tag is not variable via mkchain, we can just hard code it in values.yaml. That's why i got rid of rpc-auth and zerotier images here as well. When cloning the repo, the tags will have dev by default as we discussed. When pulling from the helm repo, they will be tagged with the release version.

Just need to keep in mind my above comment/question regarding how the CI finds tezos-k8s images and tags them.

charts/tezos/values.yaml
nicolasochem: Your other suggestion sounds better to me. Will you fix it on your branch? then I can rebase.

harryttd: yes i will do. We should, to make it consistent, make the image from `tezos-chain-initiator:dev` to `tezos-k8s-chain-initiator:dev`

nicolasochem: so the docker image will be named tezos-k8s-chain-initiator but how should the folder be named in the repo?

I am probably going to move the two scripts in charts/tezos/utils into custom containers in a subsequent PR so let's agree before I start :)

harryttd: You should name the folder `chain-initiator`.

The CI here https://github.com/tqtezos/tezos-k8s/blob/docs/.github/workflows/ci.yml#L179 will prepend `tezos-k8s` to the folder name, so `tezos-k8s-chain-initiator`. That is what the image will be named on docker hub, under the `tqtezos` repo. So you would pull/push `tqtezos/tezos-k8s-chain-initiator`. All of our image names will and should have a single format. 

In values.yaml set image names to `tezos-k8s-chain-initiator:dev`. Then the CI will tag with new release versions.

So basically flow is: 
- folder: image name _without_ `tezos-k8s` prepended
- values.yaml: images have `tezos-k8s` prepended with tag `dev`
- CI updates values.yaml images with release tag
- devspace needs have the same name as the image in values.yaml. Doesn't need to reference tag `dev` as devspace will use their own tag

mkchain/tqchain/mkchain.py
```diff
@@ -72,7 +72,7 @@ def get_genesis_vanity_chain_id(seed_len=16):
     "bootstrap_peer": {"help": "peer ip to join"},
     "docker_image": {
         "help": "Version of the Tezos docker image",
-        "default": "tezos/tezos:v7-release",
+        "default": "tezos/tezos:v8-release",
```
harryttd, line 75: what do you think? should we make this 8.1 bec of all those issues in v8?

nicolasochem: `v8-release` points to 8.1 at the moment

charts/tezos/values.yaml
nicolasochem: thanks! that's useful. helm does not like dashes so I had to pot underscore as the key for helm values, otherwise I did as you said. I'll let you switch to a different dict for third-party / custom images.

devspace.yaml
```diff
@@ -16,6 +16,10 @@ images:
     image: tezos-k8s-zerotier
     dockerfile: ./zerotier/Dockerfile
     context: ./zerotier
+  tezos-k8s-chain-initiator:
```
harryttd, line 19: Sorry it's super confusing. (don't know of a better way)
this line can just be chain-initiator. Not a huge deal but just pointing out. This prop is just a reference internally for devspace. So easier to call it by the simple image name without the prefix. We know it is tezos-k8s bec devspace is building and watching tezos-k8's images.

It's line 20 that needs the prefix. So that is set good.

charts/tezos/values.yaml
```diff
@@ -9,7 +9,8 @@ zerotier_config:
   zerotier_token: YOUR_ZEROTIER_TOKEN_HERE
 images:
   zerotier: tezos-k8s-zerotier:dev
-  tezos: tezos/tezos:v7-release
+  tezos: tezos/tezos:v8-release
+  tezos_k8s_chain_initiator: tezos-k8s-chain-initiator:dev
```
harryttd, line 13: Sorry again, i might not have been clear. It is just the value that needs the prefix. Just like zerotier above. The key can be the simple image name bec it is just being used internally by helm to reference the image. Helm doesn't care what the key is. Just that the value is the correct image for it to pull.

I need to leave docs on this for devs to know the process of creating new images.

harryttd (approved):

PR #70, opened 2021-01-14
replace script injection with custom containers

nicolasochem:
We were lucky that script injection was working up to that point.

This allows for better readability, debuggability, and if we need
an extra utility like jq or a pip package, it's possible too.


charts/tezos/templates/bootstrap.yaml
```diff
@@ -18,34 +18,31 @@ spec:
...
+          args:
             - "--generate-parameters-json"
           envFrom:
             - configMapRef:
                 name: tezos-config
-          image: python:alpine
-          imagePullPolicy: IfNotPresent
-          name: tezos-config-generator
+          env:
+            - name: MY_POD_IP
```
harryttd, line 39: I don't believe this env var is need here. Only when `--generate-config-json` is used.

https://github.com/tqtezos/tezos-k8s/blob/docs/charts/tezos/utils/generateTezosConfig.py#L131
`--generate-config-json => get_node_config => generate_node_config`

It wasn't in code before

charts/tezos/templates/bootstrap.yaml
```diff
@@ -183,38 +177,29 @@ spec:
                 - SYS_ADMIN
             privileged: true
           volumeMounts:
-            - mountPath: /opt/tqtezos
```
harryttd, line 186: was this volume even being used before by this container?

nicolasochem: Yes, before version 1.0, but it should have been removed for v1.0 release.

harryttd (approved):

PR #74, opened 2021-01-20
support for multiple bakers

nicolasochem:
Previously we used to create 4 accounts always. We simplify by
creating just one account per baker. As a consequence, all blocks
are baked on priority zero from the beginning.

We add a new parameter to mkchain `--number-of-bakers`, defaults to
1. If we pass a higher number n, it will create n bakers and n nodes
baking for each baker.

Bakers and nodes can be mapped using values.yaml: specify which node
bakes for which address. This requires jq which requires a custom
baking container. This also allowed me to remove helm hacks.

We make a few assumptions for now:
* every baking node is also a bootstrap node
* the first baker's address is also the chain activator

I tested in non-zerotier mode and zerotier mode with one create
namespace and one invite namespace in the same minikube cluster; it
works, but required me to remove the hard-coded nodePort for p2p.


charts/tezos/templates/bootstrap.yaml
```diff
@@ -226,7 +223,17 @@ spec:
           name: dev-net-tun
         - emptyDir: {}
           name: config-volume
-        - name: var-volume
-          persistentVolumeClaim:
-            claimName: tezos-bootstrap-node-pv-claim
+        - configMap:
```
harryttd, line 226: i think this configmap snuck in. wonder why k8s doesn't error

config-generator/entrypoint.py
```diff
@@ -39,15 +39,15 @@ def main():
         if CHAIN_PARAMS["zerotier_in_use"]:
             with open("/var/tezos/zerotier_data.json", "r") as f:
                 net_addr = json.load(f)[0]["assignedAddresses"][0].split("/")[0]
-            if bootstrap_peers == [] and "bootstrap" not in socket.gethostname():
+            if bootstrap_peers == []:
```
harryttd, line 42: Is the logic here now that we want all bootstrap peers to know about each other? We should probably filter out the node's own ip like we talked about.

key-importer/entrypoint.sh
```diff
@@ -2,7 +2,10 @@
...
 
-for acct in ${BOOTSTRAP_ACCOUNTS}; do
-    key=$(eval echo \$${acct}_${KEYS_TYPE}_key)
-    tezos-client -d /var/tezos/client --protocol PsDELPH1Kxsx import ${KEYS_TYPE} key $acct unencrypted:${key} -f
+echo $ACCOUNTS | jq -c --raw-output .[] | while read line; do
+    key=$(echo $line | jq -r '.key')
+    name=$(echo $line | jq -r '.name')
+    keytype=$(echo $line | jq -r '.type')
+    printf "\nImporting key ${name}\n"
+    tezos-client -d /var/tezos/client --protocol PsDELPH1Kxsx import ${keytype} key ${name} unencrypted:${key} -f
```
harryttd, line 10: we should prob make proto a variable. Seems we can get it from values.yaml protocol.hash

mkchain/tqchain/mkchain.py
```diff
@@ -133,36 +139,42 @@ def main():
...
+                    "type": key_type,
                 }
             )
 
+    creation_nodes = [
+        {"bake_for": f"baker{n}"} for n in range(args.number_of_bakers)
+    ] + [{} for n in range(args.number_of_nodes - args.number_of_bakers)]
+
+    # for invitation, do not include baking nodes, and include at least one node
+    invitation_nodes = [{}] * args.number_of_nodes
```
harryttd, line 161: Should invite only get one node by default? https://github.com/tqtezos/tezos-k8s/issues/69

mkchain/tqchain/mkchain.py
```diff
@@ -109,12 +114,13 @@ def main():
...
-        "bootstrap_account_2",
-        "genesis",
-    ]
+    if args.number_of_bakers < 1:
+        print(
+            f"Invalid argument --number-of-bakers {args.number_of_bakers}, must be 1 or more"
+        )
+        exit(1)
+
+    bootstrap_accounts = [f"baker{n}" for n in range(args.number_of_bakers)]
```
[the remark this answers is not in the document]
nicolasochem: Yes, it's changed in this branch: https://github.com/nicolasochem/tezos-k8s/pull/1

config-generator/entrypoint.py
nicolasochem: Done in https://github.com/nicolasochem/tezos-k8s/pull/1 that I will submit next as a PR.

charts/tezos/templates/bootstrap.yaml
```diff
@@ -61,16 +71,16 @@ spec:
...
 {{- if not .Values.is_invitation }}
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: tezos-bootstrap-node
   namespace: {{ .Release.Namespace }}
 spec:
+  replicas: {{ include "bakers" . | fromYaml | values | len }}
+  serviceName: tezos-bootstrap-node-rpc
```
harryttd, line 80: Roland asked me why the bakers were spinning up serially. I checked here and confirmed the `podManagementPolicy: Parallel` was not set like it is for reg nodes. Wondering if it slipped by or there is a reason bakers/bootstrapers shouldn't do serially.

harryttd (approved):

PR #75, opened 2021-01-21
DO NOT MERGE Pulumi code to deploy a tezos chain on aws

nicolasochem:
This is what I used to deploy https://spruce.tqhosted.com on AWS.

It requires more work. We should probably set up some CI autodeploy
to keep the account organized.

I am not even sure if this code belongs to tezos-k8s or we should
set up a different repo.

To use:

```
mkchain pulumi
cd pulumi
pulumi up
```

There are 2 manual steps related to the https endpoint that require
further automation.


nicolasochem:
replaced by #118

PR #76, opened 2021-01-21
wait for bootstrap node to be up before launching other nodes

nicolasochem:
We replace the bootstrap service with a headless service and have
all nodes connect directly to the pod ips.

This exposes a bug where nodes are stuck with zero connections if
they try to connect too early. This was previously masked by kubeproxy.

We fix that by adding an init container which waits for the first
bootstrap node to accept connections.

The bootstrap attribute is now declarative: we no longer consider
every baking node as bootstrap. The bootstrap: true flag must be passed
in values.yaml. mkchain adds it to the first two baking nodes.

We add headless services in order to give the pods a meaningful FQDN.

Rename "bootstrap" statefulset to "baker" to reflect this new model.

Other changes:

* In values.yaml, baker nodes and regular nodes are now listed
separately, which simplifies the code in other places.

### DIFF
--- a/charts/rpc-auth/templates/rpc-auth.yaml
+++ b/charts/rpc-auth/templates/rpc-auth.yaml
@@ -134,7 +134,7 @@ spec:
             - name: REDIS_PORT
               value: {{ .Values.redis.port | quote }}
             - name: TEZOS_RPC_SERVICE
-              value: tezos-bootstrap-node-rpc
+              value: tezos-node-rpc
             - name: TEZOS_RPC_SERVICE_PORT
               value: "8732"
 ---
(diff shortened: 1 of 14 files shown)


charts/rpc-auth/templates/rpc-auth.yaml
```diff
@@ -134,7 +134,7 @@ spec:
             - name: REDIS_PORT
               value: {{ .Values.redis.port | quote }}
             - name: TEZOS_RPC_SERVICE
-              value: tezos-bootstrap-node-rpc
+              value: tezos-node-rpc
```
harryttd, line 137: We should at some point make these sort of values, i.e. service names (even tho one would expect them to never change), into helm variables

nicolasochem: ok but not now?

harryttd: totally just brining it up. I'll make an issue

elric1 (approved):

elric1:
I have been testing these changes and they appear to strictly improve things for me.

harryttd (approved):

PR #78, opened 2021-01-22
Fix ci updating docker tags in helm charts

harryttd:


nicolasochem (approved):
looks good; no way to test until it's merged and released so let's do it :)

PR #79, opened 2021-01-25
Improve wait logic in wait-for-bootstrap

elric1:
- Increase the length of the timeouts,
	- Backoff on increasing failures
	- Add random jitter
	- add an initial delay for non bootstrap node


nicolasochem (approved):
It works fine on my machine, waits only for the first boot, and does not make the experience worse for a low number of nodes.

PR #80, opened 2021-01-25
Only import the required keys in init containers

elric1:
We can improve this further in time for large scale nodes, specifically, maybe we can just create one bootstrap account on a single node and have bakers create their own accounts as they spawn and then drink from a faucet. This would scale much better as we wouldn't have to create all of the accounts serially in advance but would rather create them in parallel at spawn time.

### DIFF
--- a/key-importer/entrypoint.sh
+++ b/key-importer/entrypoint.sh
@@ -1,12 +1,52 @@
 #!/bin/sh
+
 mkdir -p /var/tezos/client
 chmod -R 777 /var/tezos/client
 
-echo $ACCOUNTS | jq -c --raw-output .[] | while read line; do
-    key=$(echo $line | jq -r '.key')
-    name=$(echo $line | jq -r '.name')
-    keytype=$(echo $line | jq -r '.type')
-    protocol=$(echo $CHAIN_PARAMS | jq -r '.protocol_hash')
-    printf "\nImporting key ${name}\n"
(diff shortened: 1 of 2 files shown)


key-importer/entrypoint.sh
```diff
@@ -1,12 +1,32 @@
 #!/bin/sh
+
+set -x
```
nicolasochem, line 3: this prints the secret keys in the init container log, please remove.

elric1: Okay, I've update this. I also put logic in to not double import baker0's key.

nicolasochem (approved):

nicolasochem (changes requested):
actually just realized it breaks the multiple baker use case

elric1:
The issue was that the keys were not imported into the activate-job. I've fixed that.

harryttd:
Only baker0 is baking for me. But the parameters.json seems the same. the only diff is that each node has diff bootstrap_accounts which makes sense

elric1:
Hmmm, all of my bakers appear to be "Injecting block"s.

harryttd:
do they say "no slot found at level..." compared to baker-0?

nicolasochem (approved):
It works now after rebase, so I suppose it didn't have all the changes in master.

In the process we realized that the nodes don't need parameters.json (only activate does) so it should be removed.

PR #81, opened 2021-01-25
Use tezos-client rpc get /network/peers for readiness

elric1:
In our test, we use tezos-client and wc(1) so that we do not pick up
a dependency on something that is unlikely to be installed on any image
that we want to test.  It would be much more efficient, e.g., to use
curl and jq---but these may not exist in a perfectly functioning image.
We do know that tezos-client actually must exist and we assume that
wc(1) is overwhelmingly likely to exist.

### DIFF
--- a/baker/entrypoint.sh
+++ b/baker/entrypoint.sh
@@ -3,6 +3,6 @@
 set -x
 
 baker_command=$(echo $CHAIN_PARAMS | jq -r '.baker_command')
-POD_INDEX=$(echo $POD_NAME | sed -e s/tezos-bootstrap-node-//)
-baker_account=$(echo $NODES | jq -r "[.[] | select(.bake_for)] | . [${POD_INDEX}].bake_for")
+POD_INDEX=$(echo $POD_NAME | sed -e s/tezos-baking-node-//)
+baker_account=$(echo $NODES | jq -r ".baking[${POD_INDEX}].bake_for")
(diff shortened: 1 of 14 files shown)


nicolasochem:
This does not work on my minikube. Hostname for tezos-baking-node-1 is not resolvable from other nodes until tezos-baking-node-1 is ready, and tezos-baking-node-1 is not ready until it has 5 peers, so I am deadlocked.

I have a more general concern about the usage of peer count for "readiness" state. A node is ready before it has peers. The previous code was checking whether the rpc server was running.

elric1:
Thanks for pointing this out, I was experiencing this issue and didn't get to the bottom of it. I didn't realise that DNS would be affected by the readinessprobe, that's a little counter-intuitive to me. I was expecting that it would remove the address from the load balancing (that we aren't using) and turn the pod's line on k9s red (which would make it easy to see.)

I was considering changing my test into a livenessprobe, actually, because that would provide a belt-and-suspenders for our wait-for-bootstrap option. If a node gets into a state where it is stuck with zero peers, maybe we should kick it.  OTOH, maybe it's actually better to leave it in place because it exposes a bug in the node which should be fixed. In this case, this logic can be used in a different kind of monitor which doesn't have side effects.

elric1:
I'm going to submit a different PR which simply removes the readiness probes. They make my minikubes environment unstable because DNS names come and go. Stability is greatly improved by removing them.

PR #86, opened 2021-01-26
Remove readiness probes

elric1:
Readiness probes have the unfortunate side-effect of causing the
DNS resolution of our hostnames to fail whenever kubernetes thinks
that the node is not ready.  This exposes some underlying bugs in
tezos-node and causes many nodes to simply stop trying to find
peers.  Waiting to start helps the issue, but we find that minikubes
sometimes decides that nodes are unavailable randomly even after it
considered them to be ready.  This makes our name service complete
non-deterministic.

### DIFF
--- a/charts/tezos/templates/baker.yaml
+++ b/charts/tezos/templates/baker.yaml
@@ -97,16 +97,6 @@ spec:
               name: tezos-rpc
             - containerPort: 9732
               name: tezos-net
-          readinessProbe:
-            exec:
-              command:
-                - nc
-                - "-z"
-                - '127.0.0.1'
-                - '8732'
-            initialDelaySeconds: 2
-            periodSeconds: 2
-            timeoutSeconds: 1
           volumeMounts:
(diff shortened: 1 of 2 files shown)


nicolasochem (approved):
It works well on my machine.

I also tried removing `wait-for-bootstrap` completely but this fails. So it's still necessary to wait for dns to resolve before starting the other nodes.

Did I merge this one too quickly? https://github.com/tqtezos/tezos-k8s/pull/79

Or does everything in PR#79 still make sense in this context where there is no longer a readiness probe?

elric1:
Both are required. Even without readiness probes, k8s will not provide DNS names until the pod has run through all of its init containers. What this patch does is ensure that once the names are exposed, they stay exposed. If they fail later, then you end up with races because just because wait-for-bootstrap can succeed but the name will be removed before the tezos-node is spun up if the baker0's readiness probe fails at the same time.

PR #87, opened 2021-01-26
Ensure that all bakers have a non-zero balance on start

elric1:
We update the entry point for the baker container s.t. each node will
transfer a set balance from baker0 to its account.  This ensures that
every baker has some stake.

Because we need to wait for both the tezos-node to come up and for the
balance transfer to complete, we now no longer start the baker before
the tezos-node is started.  This solves a race where sometimes the baker
would exit because it was started before the node.


elric1:
I'll rethink this one.

PR #90, opened 2021-01-26
add endorser to every baker

nicolasochem:


elric1:
It looks like it is working for me.

elric1 (approved):
I ran it in my minikubes environment and it worked as expected as far as I can tell.

baker-endorser/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -x
+
+proto_command=$(echo $CHAIN_PARAMS | jq -r '.proto_command')
+if [ "${WHAT}" == "baker" ]; then
```
harryttd, line 6: I personally find this var name a little confusing as to what it is referring to.

harryttd (approved):

PR #92, opened 2021-01-26
Make all bakers bootstrap value so that they can bake

elric1:


nicolasochem:
`bootstrap` flag indicates that all nodes will have the node as bootstrap parameter.

`bake_for` indicates that a baker daemon will be activated for this node, and which account it will bake for.

It's intentional that not every baking node is a bootstrap node, why are you linking the two?

elric1:
I got my wires crossed. I'll close this one.

PR #93, opened 2021-01-27
Restore ability to scale up number of nodes

nicolasochem:
To scale up, run mkchain again with a higher --number-of-nodes then run
`helm upgrade --install`

When running mkchain with a chain_name corresponding to helm values
files that already exist, it will preserve the timestamps and baking
accounts.

Scale up of nodes is permitted and works, scale up of number of
bakers is not permitted as it would not work.

Turning a regular chain into a zerotier chain works (I checked).

### DIFF
--- a/mkchain/tqchain/mkchain.py
+++ b/mkchain/tqchain/mkchain.py
@@ -146,12 +146,6 @@ def main():
         "images": {
             "tezos": args.docker_image,
         },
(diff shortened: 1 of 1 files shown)


harryttd (approved):

harryttd:
I would remove from comment turning chain into ZT based on our discussion

PR #94, opened 2021-01-27
fix #58 wait for bootstrap node to be registered on zerotier

### LINKED ISSUE #58: Occasionally nodes fail to connect over Zerotier
When I had a connection failure, I checked `/var/tezos/zerotier_network_members.json` in the container. I noticed that the `name` field for a member was set as `""`. I also noticed in one instance that there was nothing in the `ipAssignments` array either.

Empty `name` causes `generateTezosConfig.py` to not get the ip of the bootstrap node. Hence the peer nodes will not have the bootstrap node's ip in their `bootstrap-peers` list and they won't connect.

This is hard to reproduce.
Discussing with Nicolas, it seems we should fail the `entrypoint.sh` script on api call errors, and also fail the `generateTezosConfig.py` if it didn't find any bootstrap node peers.

nicolasochem:
This allows a zerotier network to establish a full mesh at first go

### DIFF
--- a/charts/tezos/templates/baker.yaml
+++ b/charts/tezos/templates/baker.yaml
@@ -163,6 +163,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          - name: CHAIN_NAME
+            value: "{{ .Values.chain_name }}"
 {{- end }}
         - image: {{ .Values.tezos_k8s_images.key_importer }}
           name: key-importer
--- a/charts/tezos/templates/node.yaml
+++ b/charts/tezos/templates/node.yaml
@@ -83,6 +83,8 @@ spec:
             - mountPath: /dev/net/tun
               name: dev-net-tun
           env:
+          - name: CHAIN_NAME
+            value: "{{ .Values.chain_name }}"
           - name: POD_NAME
             valueFrom:
               fieldRef:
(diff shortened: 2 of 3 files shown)


nicolasochem:
@harryttd I tried on master branch and could not replicate the problem. I always establish full mesh at first go. It seems that the `wait_for_bootstrap`init container is helping. If bootstrap node is up, then the zerotier init container necessarily ran, and the other nodes will be able to find the bootstrap ip from zerotier metadata.

It does not look like this PR is needed anymore. Does anything still need fixing?

harryttd:
@nicolasochem I believe there is a race condition. I have tested many times locally and sometimes all the nodes connect and sometimes not.

What i think is going on: All the nodes get a ZT ip at the same time. What if the Bnode (bootstrap node) did not finish getting an ip, updating name / description, etc, before other nodes fetch the list of network members? That would mean some nodes will not have the Bnode's ip, or it will have it but not with the name set. (Is there a way to add a member and at the same time give it a name/description so it doesn't have to be a separate api call?) I see locally how a peer node does not have any ip's in its bootstrap peers config. The wait-for-bootstrap container will always complete successfully if the Bnode is up bec the script is checking the internal cluster domain name of the pod, not the ZT ip of the container.

I think your current PR would fix it. Perhaps as we discussed a couple of weeks back, we should instead just reorder the initcontainer sequence. It might work that we can run `key-import => identity => wait-for-bootstrap => ZT ip => config gen`. Bnode would exit immediately from wait-for, get an ip, then gen its config. Meanwhile, all other nodes have not gotten their ip yet. Once Bnode is running, only then do they get their ip (guaranteeing getting Bnode ip) as well as other network members.

harryttd:
We should also fail the get ZT ip script if an api call errors, as discussed.

PR #96, opened 2021-01-27
baker/endorser containers must wait for node to start

elric1:
This solves a race where the baker/endorser may exit if the node has not
yet started when it starts.  Of course, k8s will restart the container,
however, this gives us RESTARTS on the console which we are hoping will
exclusively represent issues with the underlying code rather than our
test framework.

### DIFF
--- a/baker-endorser/entrypoint.sh
+++ b/baker-endorser/entrypoint.sh
@@ -2,10 +2,34 @@
 
 set -x
 
+TEZ_VAR=/var/tezos
+TEZ_BIN=/usr/local/bin
+CLIENT_DIR="$TEZ_VAR/client"
+NODE_DIR="$TEZ_VAR/node"
+
 proto_command=$(echo $CHAIN_PARAMS | jq -r '.proto_command')
 if [ "${DAEMON}" == "baker" ]; then
-    extra_args="with local node /var/tezos/node"
+    extra_args="with local node $NODE_DIR"
 fi
 POD_INDEX=$(echo $POD_NAME | sed -e s/tezos-baking-node-//)
 baker_account=$(echo $NODES | jq -r ".baking[${POD_INDEX}].bake_for")
-/usr/local/bin/tezos-${DAEMON}-${proto_command} -d /var/tezos/client run ${extra_args} ${baker_account:?Error: baker account not set}
+
+if [ -z "$baker_account" ]; then
+    echo "Baker account (bake_for:) not set" 1>&2
+    exit 1
+fi
+
+CLIENT="$TEZ_BIN/tezos-client -d $CLIENT_DIR"
+CMD="$TEZ_BIN/tezos-$DAEMON-$proto_command -d $CLIENT_DIR"
+
+#
+# All bakers need to wait for their local node to be bootstrapped:
+
+while ! $CLIENT rpc get chains/main/blocks/head; do
+    sleep 5
+done
+
+#
+# And, obviously, we need to actually bake:
+
+$CMD run ${extra_args} ${baker_account}


elric1:
Here's what I get that prompted this commit:

```$ kubectl logs -n tqtezos tezos-baking-node-11 -c baker -p   
++ echo '{' '"bootstrap_mutez":' '"4000000000000",' '"bootstrap_peers":' '[],' '"nodes":' '{"baking":[{"bake_for":"baker0","bootstrap":true},{"bake_for":"baker1","bootstrap":true},{"bake_for":"baker2"},{"bake_for":"baker3"},{"bake_for":"baker4"},{"bake_for":"baker5"},{"bake_for":"baker6"},{"bake_for":"baker7"},{"bake_for":"baker8"},{"bake_for":"baker9"},{"bake_for":"baker10"},{"bake_for":"baker11"}],"regular":[]},' '"chain_name":' '"elric",' '"genesis_block":' '"BMNrMM2HtAojkejcAcBHrMFzmTUgfC6vhxxXaCaMacfDRmdc9pz",' '"activation_account":' '"baker0",' '"timestamp":' '"2021-01-27T21:03:24.392858+00:00",' '"zerotier_in_use":' false, '"baker_command":' '"tezos-baker-007-PsDELPH1",' '"protocol_hash":' '"PsDELPH1Kxsxt8f9eWbxQeRxkjfbxoqM52jvs5Y5fBxWWh4ifpo"' '}'
++ jq -r .baker_command
+ baker_command=tezos-baker-007-PsDELPH1
++ sed -e s/tezos-baking-node-//
++ echo tezos-baking-node-11
+ POD_INDEX=11
++ echo '{"baking":[{"bake_for":"baker0","bootstrap":true},{"bake_for":"baker1","bootstrap":true},{"bake_for":"baker2"},{"bake_for":"baker3"},{"bake_for":"baker4"},{"bake_for":"baker5"},{"bake_for":"baker6"},{"bake_for":"baker7"},{"bake_for":"baker8"},{"bake_for":"baker9"},{"bake_for":"baker10"},{"bake_for":"baker11"}],"regular":[]}'
++ jq -r '.baking[11].bake_for'
+ baker_account=baker11
+ /usr/local/bin/tezos-baker-007-PsDELPH1 -d /var/tezos/client run with local node /var/tezos/node baker11
Connection refused, retrying in 1.00 seconds...
Waiting for the node to be bootstrapped...
Connection refused, retrying in 1.50 seconds...
Connection refused, retrying in 2.25 seconds...
Connection refused, retrying in 3.38 seconds...
Connection refused, retrying in 5.06 seconds...
Error:
  Rpc request failed:
     - meth: GET
     - uri: http://localhost:8732/monitor/bootstrapped
     - error: Unable to connect to the node: "Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")"
```

baker-endorser/entrypoint.sh
```diff
@@ -2,10 +2,34 @@
...
+CLIENT="$TEZ_BIN/tezos-client -d $CLIENT_DIR"
+CMD="$TEZ_BIN/tezos-$DAEMON-$proto_command -d $CLIENT_DIR"
+
+if [ -z "$baker_account" ]; then
+    echo "Baker account (bake_for:) not set" 1>&2
+    exit 1
+fi
+
+#
+# All bakers need to wait for their local node to be bootstrapped:
```
nicolasochem, line 26: but they will, natively. I think you should just wait for the node to be reachable with rpc.

elric1: tezos-client is just doing the RPC. It's just packaged up in the client program and therefore more likely to exist in the container than curl or nc.

nicolasochem: But then can you query another rpc that always returns instantly such as /chains/main/blocks/head ? This one will reply as soon as rpc server as come online, which is what we want to check. The daemon will wait for the node to be bootstrapped so there is no need to do it in the init container, we just want to avoid suprious restarts due to rpc server not being online yet.

baker-endorser/entrypoint.sh
```diff
@@ -2,10 +2,34 @@
...
 if [ "${DAEMON}" == "baker" ]; then
-    extra_args="with local node /var/tezos/node"
+    extra_args="with local node $NODE_DIR"
 fi
 POD_INDEX=$(echo $POD_NAME | sed -e s/tezos-baking-node-//)
 baker_account=$(echo $NODES | jq -r ".baking[${POD_INDEX}].bake_for")
-/usr/local/bin/tezos-${DAEMON}-${proto_command} -d /var/tezos/client run ${extra_args} ${baker_account:?Error: baker account not set}
+
+CLIENT="$TEZ_BIN/tezos-client -d $CLIENT_DIR"
+CMD="$TEZ_BIN/tezos-$DAEMON-$proto_command -d $CLIENT_DIR"
```
harryttd, line 18: don’t mean to nitpick. What do you thing of moving these vars after the if check right below. I find the flow a little easier to read as those vars are not needed before the if

nicolasochem (approved):

PR #99, opened 2021-02-04
wait-for-bootstrap: use either node-0 or node-1 to help with deadlocks.

elric1:


wait-for-bootstrap/entrypoint.sh
```diff
@@ -43,7 +49,13 @@ randomsleep() {
 
 sleep 10
 randomsleep
-echo "waiting for bootstrap node to accept connections"
-until </dev/null nc -q 0 ${FBN} 8732; do
+echo "waiting for bootstrap nodes to accept connections"
+
+while :; do
+    for node in $BOOTSTRAP_NODES; do
+	if </dev/null nc -q 0 ${node} 8732; then
```
harryttd, line 56: can we log on success that we have connected and to which node?

nicolasochem (changes requested):
It makes sense, but do not assume tezos-baking-node-0 and tezos-baking-node-1 are bootstrap. We define bootstrap nodes in helm values. For example, mkchain gives me:

```
nodes:
  baking:
  - bake_for: baker0
    bootstrap: true
  - bake_for: baker1
    bootstrap: true
  - bake_for: baker2
```

But anyone could launch a topology such as:

```
nodes:
  baking:
  - bake_for: baker0
  - bake_for: baker1
    bootstrap: true
  - bake_for: baker2
    bootstrap: true
```

In which case `BOOTSTRAP_NODE="tezos-baking-node-1.tezos-baking-node tezos-baking-node-2.tezos-baking-node"`

So BOOTSTRAP_NODES list has to be built generically; I recommend passing the `nodes` helm value as json and using jq to generate it.

See how it is done in config-generator:

https://github.com/tqtezos/tezos-k8s/blob/master/config-generator/entrypoint.py#L47-L54

wait-for-bootstrap/entrypoint.sh
elric1: I made it a lot more chatty.

elric1:
I've updated this to grab the bootstrap nodes from `CHAIN_PARAMS` as suggested. I force pushed as it's essentially a new attempt. I also included a couple of commits for small issues that I noticed along the way.

elric1:
So, after the structural changes to values files just merged in, I've restructured this to use `$NODES` rather than `$CHAIN_NAME`. It's still filtering for the configured bootstrap hosts without making assumptions.

elric1:
I've made this address #161 as a drive-by.

elric1:
Make new PR.

PR #100, opened 2021-02-04
Decouple number-of-nodes from number-of-bakers

elric1:
We now consider --number-of-nodes to be the nodes which do not
include a baker/endorser pair.  It's more natural to consider
that:

    $ mkchain --number-of-bakers 40 --number-of-nodes 5 my_chain

will bring up 40 baker nodes and 5 plain nodes.

This is helpful with pulumi because it has been waiting for zero nodes
to come up and timing out.  With the current code:

    $ mkchain --number-of-bakers 40 my_chain

will bring up 40 bakers and a single node.

### DIFF
--- a/mkchain/README.md
+++ b/mkchain/README.md
@@ -57,8 +57,8 @@ You can explicitly specify some values by:
 | -------------------------------- | ------------------ | --------------------------------------------------------------------------- | ---------------------- |
 | bootstrap_peer                   | --bootstrap-peer   | Peer ip to connect to                                                       |                        |
 | images.tezos                     | --docker-image     | Version of the Tezos docker image                                           | tezos/tezos:v8-release |
-| number_of_nodes                  | --number-of-nodes  | Number number of plain nodes in the cluster                                 | 1                      |
-| number_of_bakers                 | --number-of-bakers | Number of nodes in the cluster that are bakers                              | 1                      |
+| number_of_nodes                  | --number-of-nodes  | Number of non-baking nodes in the cluster                                   | 1                      |
+| number_of_bakers                 | --number-of-bakers | Number of baking nodes in the cluster                                       | 1                      |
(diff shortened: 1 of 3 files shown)


harryttd:
Can you please change docs: https://github.com/tqtezos/tezos-k8s/blob/master/mkchain/README.md for mkchain args

nicolasochem (changes requested):
ok in principle. Documentation should be updated, and the restriction that `number_of_nodes` should be at least 1 must be lifted, since it should be possible to launch a topology with only baking nodes.

`git grep number_of_nodes` and `git grep number-of-nodes` will give pointers as of which parts should be updated.

elric1:
I have made the requested updates.

mkchain/README.md
```diff
@@ -57,7 +57,7 @@ You can explicitly specify some values by:
 | -------------------------------- | ------------------ | --------------------------------------------------------------------------- | ---------------------- |
 | bootstrap_peer                   | --bootstrap-peer   | Peer ip to connect to                                                       |                        |
 | images.tezos                     | --docker-image     | Version of the Tezos docker image                                           | tezos/tezos:v8-release |
-| number_of_nodes                  | --number-of-nodes  | Total number of nodes in the cluster                                        | 1                      |
+| number_of_nodes                  | --number-of-nodes  | Number number of plain nodes in the cluster                                 | 1                      |
```
harryttd, line 60: 'number' repeated twice. Also, can you make it something like `Number of non-baker nodes in the cluster`? Might not be clear what 'plain' means.

harryttd:
Realized another thing: The default number of nodes should now be 1 (and reflected in mkchain docs). It used to be only 1 node was spun up by default and that node was the bootstrap/baker. With these changes, two nodes would be spun up, one peer and one baker. 

I'm working on the top level README.md to reflect all these changes.

harryttd (approved):

nicolasochem (approved):

Issue #102, opened 2021-02-04
Automatically delete Zerotier members when deleting a chain


harryttd:
This suggestion is primarily for devs. Because devs are often spinning up new chains and then deleting them, ZT ip's keep increasing.

A couple of issues:
- At a certain point, ZT website warns of too many members. A [script](https://github.com/tqtezos/tezos-k8s/blob/master/zerotier/zerotier-purge.sh) was provided to delete network members. You can filter nodes to delete by diff values. However, this is a manual process that requires you to remember to do so.
- Currently you are required to delete members in order for a new chain to work. This is bec bootstrap nodes are created and are assigned new ip's by ZT. Each peer node on launch looks for bootstrap node ip's to get the p2p service working, but often finds stale ip's that it thinks are live running nodes. Then nodes do not all connect to each other.

A possible solution to these problems is to create a [Helm hook](https://helm.sh/docs/topics/charts_hooks/) that on deletion of release (`post-delete`) fires a command to delete the ZT member. `helm uninstall` does not remove pvc's so that would need to be done likely via kubectl.

Most likely as this should be implemented only for developers, there needs to be a way to templatize the hook. Possible way is to create a separate values.yaml file for different envs. dev yaml will have a value that is used to add the hook yaml. This can be implemented with this simple suggestion. There are tools for this: https://github.com/roboll/helmfile. Nice intro to helmfile: https://www.arthurkoziel.com/managing-helm-charts-with-helmfile/

PR #103, opened 2021-02-09
fix #58 change order of init containers to prevent zerotier races

### LINKED ISSUE #58: Occasionally nodes fail to connect over Zerotier

nicolasochem:
replaces https://github.com/tqtezos/tezos-k8s/pull/94

There are two relevant init containers, get-zerotier-ip and
wait-for-bootstrap. get-zerotier-ip used to run first.

In addition to getting an ip from zerotier, get-zerotier-ip also
downloads all interfaces for the config generator to generate a list
of bootstrap ips.

When get-zerotier-ip runs first, it may download this list before the
bootstrap node has registered its ip as a bootstrap ip, sometimes causing a
race condition where the non-bootstrap nodes do not have a bootstrap
ip configured.

With this PR, wait-for-bootstrap comes first. It waits for bootstrap
node to be up (by probing it internally, not through zerotier). Only
then will it create an ip and download the bootstrap peer ip from
zerotier. Since at this point the init container for the bootstrap
node has already ran, the bootstrap ip will for sure be registered in
zerotier.

I tried several times in zerotier and non zerotier modes, it always
works for me.

See PR 94 for a better explanation.

Also make the zerotier script fail on error, so any error with zerotier
api will fail the pod and trigger a retry.

### DIFF
--- a/charts/tezos/templates/baker.yaml
+++ b/charts/tezos/templates/baker.yaml
@@ -159,6 +159,36 @@ spec:
               name: var-volume
 {{- end }}
       initContainers:
+        - image: {{ .Values.tezos_k8s_images.key_importer }}
+          name: key-importer
+          imagePullPolicy: IfNotPresent
+          envFrom:
+            - secretRef:
+                name: tezos-secret
+            - configMapRef:
+                name: tezos-config
+          volumeMounts:
+            - mountPath: /var/tezos
+              name: var-volume
+        - args:
+            - "-c"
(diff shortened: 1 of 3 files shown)


harryttd:
related to #102

harryttd (approved):

PR #105, opened 2021-02-12
Fill in defaults aggressively in the containers to make mkchain optional

elric1:
We now import the keys in the config-generator container because
we use Python to construct the contents of /var/tezos/client by
doing a little crypto.  We generate deterministic keys based on
the hostname for keys that were not provided and intuit accounts
using default names where they are not provided.

We eliminate the key-importer container because we no longer use it.

We also provide for a numNodes: section in the provided _values.yaml
file so that users can simply specify how many of each kind of node
that they want.  We assign the first node to be a bootstrap if the
user forgets to specify one.


harryttd:
I'm going through this now. I don't think it has the latest from master. e.g. --number-of-nodes means `args.number_of_nodes - args.number_of_bakers`

charts/tezos/templates/node.yaml
```diff
@@ -88,23 +88,14 @@ spec:
...
-          volumeMounts:
-            - mountPath: /var/tezos
-              name: var-volume
         - image: {{ .Values.tezos_k8s_images.config_generator }}
           imagePullPolicy: IfNotPresent
           name: config-generator
           args:
             - "--generate-config-json"
           envFrom:
+            - secretRef:
```
harryttd, line 97: I believe tezos-secret is only needed when generating params.json as it contains the accounts. Not config.json.

charts/tezos/templates/baker.yaml
```diff
@@ -184,23 +175,14 @@ spec:
...
-          volumeMounts:
-            - mountPath: /var/tezos
-              name: var-volume
         - image: {{ .Values.tezos_k8s_images.config_generator }}
           imagePullPolicy: IfNotPresent
           name: config-generator
           args:
             - "--generate-config-json"
           envFrom:
+            - secretRef:
```
harryttd, line 184: Same comment from another spot: I believe tezos-secret is only needed when generating params.json as it contains the accounts. Not config.json.

charts/tezos/templates/node.yaml
elric1: I put this in here because it was needed by key-importer and I folded key-importers functionality into config-generator when I rewrote the calls to tezos-client to generating the keys in Python.

charts/tezos/templates/configs.yaml
```diff
@@ -4,6 +4,7 @@ data:
     {
       "bootstrap_mutez": "{{ .Values.protocol.bootstrap_mutez }}",
       "bootstrap_peers": {{ toJson .Values.bootstrap_peers }},
+      "numbakers": {{ .Values.numNodes.baking }},
```
harryttd, line 7: this needs `toJson` prepended like the above line. Otherwise value is nothing and config-generator breaks.

Also can you please keep property name format consistent like the others: e.g. `number_of_bakers` and `.Values.number_of_nodes`

charts/tezos/values.yaml
```diff
@@ -25,18 +24,16 @@ protocol:
...
+bootstrap_peers: []
 accounts:
-  - name: sample1
-    key: edsk...
-    type: secret
-  - name: sample2
-    key: edsk...
-    type: secret
+   - name: baker0
+numNodes:
```
harryttd, line 31: can you please make prop name similar to the others: `number_of_nodes`. thanks

charts/tezos/values.yaml
```diff
@@ -25,18 +24,16 @@ protocol:
...
+  baking: ""
+  regular: ""
```
harryttd, line 33: these two props should be `0` instead of `""`. Otherwise get in config-generator:
```if "numbakers" in CHAIN_PARAMS and CHAIN_PARAMS["numbakers"] > i:
TypeError: '>' not supported between instances of 'str' and 'int'```

Comment (nicolasochem) [reply charts/tezos/values.yaml:33]: they should be 1, actually, so it maps what's below.

File: config-generator/Dockerfile [hunk 6]
```diff
@@ -1,5 +1,8 @@
 FROM python:alpine
 ENV PYTHONUNBUFFERED=1
 COPY entrypoint.py /
+RUN apk add --no-cache --virtual .build-deps gcc musl-dev libffi-dev make \
```
harryttd, line 4: can you put the COPY under RUN so that every change to entrypoint doesn't need to apk add

nicolasochem (changes requested):
I merged master to this PR before testing it and it re-added the `import-keys` init container. Do not merge before syncing with latest master.

After that, I got it to work.

Per offline discussion, let's make `nodeNum` mandatory so there is less fluffy logic inside of helm. Also, make the default values be 1 baker and 0 node.

config-generator/entrypoint.py
```diff
@@ -2,11 +2,20 @@
 import json
 import os
 import socket
+import sys
+
+from datetime import datetime, timezone
```
harryttd, line 7: i think sys and datetime are unused imports

elric1: Yes, those were from an intermediate version of the code that has been superceded.

README.md
```diff
@@ -2,14 +2,18 @@
 
 This README will walk you through setting up a Tezos based private blockchain where you will spin up one bootstrap node as well as additional peer nodes if you'd like. Using `minikube`, these nodes will be running in a peer-to-peer network via a Zerotier VPN, inside of a Kubernetes cluster.
 
+Using mkchain is now optional as the docker images will fill in defaults
+for all of the values that mkchain can supply.  For more fine-grained
+control, you can still use mkchain to generate the values.yaml file and
+edit it by hand.
+
+It is also optional to setup a ZeroTier network.
```
harryttd, line 10: I believe that the one of the original points of tezos-k8s and the readme was to guide users in setting up a chain that they could expose and have other people join.

IMO:
- Based on the above assumption, Zerotier is therefore a necessary component and should not be made optional.
-  I would remove the "by hand" documentation and have them just use mkchain. Default nodes are 1 baker and 1 peer. If they want to adjust that, they can. Being that Zerotier is one of the main points here, they are using mkchain anyways to configure the ZT creds.
- Docker is a dependency (of minikube) regardless of using mkchain or not. It should be moved back to the top in prereqs. Also based on all this, python should also be moved back to the prereqs.

If everyone agrees with my points, I think no changes are necessary to the readme in this PR.

The only thing that could be of good benefit now, is like you mentioned Roland, that it is unclear that you don't need to clone the repo for this to all work. People seem to assume you do need to.

config-generator/entrypoint.py
```diff
@@ -254,6 +264,129 @@ def generate_parameters_config(parameters_argv):
...
+
+def import_keys():
+    print("Importing keys")
+    tezdir = "/var/tezos/client"
+    os.makedirs(tezdir, exist_ok=True)
+    os.chmod(tezdir, 0o777);
+    secret_keys = []
+    public_keys = []
+    public_key_hashs = []
+    for acct in fill_in_missing_accounts():
```
harryttd, line 335: What did we finally decide? Are we generating every key for every account on every node and importing all of them?

harryttd:
Reminder that this branch needs to be synced with master

README.md
nicolasochem: yes, mkchain will expand and I would not make it optional either.

elric1: I can revert the documentation changes, but I think that this should be revisited in the future.

It's definitely a good idea to have a walk-through of a specific use-case, but this is the main (and only?) documentation of the project---and so it should be general rather than only cover one use-case. I particularly noted this because the way that the documentation was written confused me and I had to ask @harryttd about it before I could make progress on firing up my first test network.

harryttd: Makes sense

config-generator/entrypoint.py
```diff
@@ -254,6 +264,129 @@ def generate_parameters_config(parameters_argv):
...
+    accounts = {}
+    for acct in ACCOUNTS:
+        accounts[acct["name"]] = acct
+    for i, node in enumerate(CHAIN_PARAMS["nodes"]["baking"]):
+        acct = node.get("bake_for", "baker" + str(i))
+        if acct not in accounts:
+            print("    Creating specified but missing account " + acct);
+            accounts[acct] = {"name" : acct}
+    if "number_of_bakers" in CHAIN_PARAMS:
+        if CHAIN_PARAMS["number_of_bakers"] > i:
```
harryttd, line 302: If no bakers are specified in values.yaml:
```yaml
number_of_nodes:
    baking: []
```
but we have:
```yaml
number_of_nodes:
    baking: 4
```
It throws an error:
```shell
Ensure that we have genesis_block
Importing keys
File "/entrypoint.py", line 392, in <module>
main()
File "/entrypoint.py", line 16, in main
import_keys()
File "/entrypoint.py", line 335, in import_keys
for acct in fill_in_missing_accounts():
File "/entrypoint.py", line 302, in fill_in_missing_accounts
if CHAIN_PARAMS["number_of_bakers"] > i:
UnboundLocalError: local variable 'i' referenced before assignment
```
Initializing `i = 0` above the `enumerate` loop seems to do the trick.

nicolasochem: I found another edge case.

If I have a values.yaml with `number_of_nodes` but I want to override it with new nodes:

```
[venv] nochem@peck ~/workspace/tezos-k8s () $ mkchain nico --number-of-nodes 2 --number-of-bakers 3
Found old values file. Some pre-existing values will remain the same, e.g. public/private keys. Please delete the values file to generate all new values.

  File "/home/nochem/workspace/tezos-k8s/mkchain/tqchain/mkchain.py", line 178, in main
    if len(old_create_values["nodes"]["baking"]) != args.number_of_bakers:
KeyError: 'nodes'
```

Also I think that if we go with `number_of_nodes`, mkchain should put it in the generated values.yaml

I am reverting to my position from yesterday. I don't want us to loose time fixing edge cases. There should be only one way to specify your topology in values.yaml, and `number_of_nodes` introduces another, which is not desirable.

values.yaml should be a fully exploded representation of the topology, which includes a full list of maps for every node. Helm should count the nodes["baking"] list to figure out how many items are in the replica set. If mkchain is not a good tool to generate values.yaml, users can write an alternative which suits their use cases.

elric1: Good catch.

nicolasochem (changes requested):
I tried mkchain with no parameters. I got a running private chain.

I tried mkchain with --number-of-bakers 3. I got a running private chain.

Then I tried to delete all accounts from values.yaml. I realized you still need the first one. So I deleted accounts 2 and 3. I got a running private chain. So account generation works.

Then I tried the same thing but gave random names for `bake_for` accounts. It worked, again, the missing keys were generated.

Then I tried to change the genesis block back to `YOUR_GENESIS_CHAIN_ID_HERE` and it worked, a valid genesis block was generated by the init container.

I tried invite with default values from mkchain, in two different namespaces in minikube. 

The chain created from the _invites_values file cannot validate blocks:

```
 19 01:18:53.823 - validator.block: validation of block BMUQbUio6veHF7AE45YsEkhPKMgy3qJTNqZXWDM5B2z9NqQK6Fp failed                                                                                                                                      │
│ Feb 19 01:18:53.823 - validator.block: Request pushed on 2021-02-19T01:18:53.779-00:00, treated in 34.968us, completed in 42.993ms ,                                                                                                                       │
│ Feb 19 01:18:53.823 - validator.block: Invalid block BMUQbUio6veH                                                                                                                                                                                          │
│ Feb 19 01:18:53.823 - validator.block:   Failed to validate the economic-protocol content of the block: Error:                                                                                                                                             │
│ Feb 19 01:18:53.823 - validator.block:                                                                    Invalid signature                                                                                                                                │
│ Feb 19 01:18:53.824 - validator.block: .                                                                                                                  
```

Looking at the output of config generator, it appears that the `genesis_pubkey` in the invited namespace is different from the one in the main namespace.

In invited namespace:

```
│   "network": {                                                                                                                                                                                                                                             │
│     "chain_name": "nico",                                                                                                                                                                                                                                  │
│     "sandboxed_chain_name": "SANDBOXED_TEZOS",                                                                                                                                                                                                             │
│     "default_bootstrap_peers": [],                                                                                                                                                                                                                         │
│     "genesis": {                                                                                                                                                                                                                                           │
│       "timestamp": "2021-02-19T01:09:36.665666+00:00",                                                                                                                                                                                                     │
│       "block": "BMYowQoSayQt24wGPnnotXPSAzK6yfhgGSgZGNP1HGu9LzDvaL4",                                                                                                                                                                                      │
│       "protocol": "PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex"                                                                                                                                                                                    │
│     },                                                                                                                                                                                                                                                     │
│     "genesis_parameters": {                                                                                                                                                                                                                                │
│       "values": {                                                                                                                                                                                                                                          │
│         "genesis_pubkey": "edpktmQaXn8WKfxEkiLsKZfHj7Fhe63nKcDUGCUxHMrxdXw7zNiVan"                                                                                                                                                                         │
│       }                                                                                                                                                                                                                                                    │
│     }                                                                                                                                                                                                                                                      │
│   }                 
```

In main namespace:

```│   "network": {                                                                                                                                                                                                                                             │
│     "chain_name": "nico",                                                                                                                                                                                                                                  │
│     "sandboxed_chain_name": "SANDBOXED_TEZOS",                                                                                                                                                                                                             │
│     "default_bootstrap_peers": [],                                                                                                                                                                                                                         │
│     "genesis": {                                                                                                                                                                                                                                           │
│       "timestamp": "2021-02-19T01:09:36.665666+00:00",                                                                                                                                                                                                     │
│       "block": "BMYowQoSayQt24wGPnnotXPSAzK6yfhgGSgZGNP1HGu9LzDvaL4",                                                                                                                                                                                      │
│       "protocol": "PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex"                                                                                                                                                                                    │
│     },                                                                                                                                                                                                                                                     │
│     "genesis_parameters": {                                                                                                                                                                                                                                │
│       "values": {                                                                                                                                                                                                                                          │
│         "genesis_pubkey": "edpkthTp4N4tfDmW4rjYAtsiqxR5qP1MDAzkCsXreZ2xYPdX6ka3J2"                                                                                                                                                                         │
│       }                                                                                                                                                                                                                                                    │
│     }                                                                                                                                                                                                                                                      │
│   }                               
```

So while the script appears to have no trouble importing private keys from values.yaml, it seems that import public keys from values.yaml is broken with this change.

mkchain/tqchain/mkchain.py
```diff
@@ -186,10 +186,8 @@ def main():
     else:
         # create new chain genesis params if brand new chain
         base_constants["genesis"] = {
-            "genesis_chain_id": get_genesis_vanity_chain_id(FLEXTESA),
-            "bootstrap_timestamp": datetime.utcnow()
-            .replace(tzinfo=timezone.utc)
-            .isoformat(),
+            "block": get_genesis_vanity_chain_id(FLEXTESA),
```
harryttd, line 189: - Do we want in this PR to remove the Flextesa dep that is creating the block hash, now that Roland's changes auto gen one?
- We might as well then also remove dep on tezos-client that gens keys. We don't have functionality anyways to import one's own keys yet and therefore are not loosing anything by removing this dep.

This would now completely remove docker as a dep of mkchain.

nicolasochem: we want these two things, but not in this PR. If invite works, I'm ok to merge as is (I won't be able to check in the next few hours)

nicolasochem (approved):
It works now, thanks!

elric1:
I rebased it up to head and squashed the two commits in the PR before I merged it, but the code should be what was tested.

PR #111, opened 2021-02-16
Invitee: don't wait for bootstrap node

### LINKED ISSUE #108: Invitee node broken
Invitee node is waiting for bootstrap node which it will never find as it is running on a separate machine. It does not need to wait for it anyways as the user is given an invite to what should be an already up and running chain.

### LINKED ISSUE #109: Remove install mkchain step for invitee in README
Invitee does not need mkchain

### LINKED ISSUE #110: mkchain: Should log out to user that an old values file was found
mkchain currently gives the impression that a new values file with new values was generated even when running subsequent mkchain commands with the same arguments. Users should be made aware that some values on subsequent will not have not changed (e.g. private/public keys). The only way to get all brand new values is to delete the values file and rerun mkchain.

harryttd:
closes #108
closes #109
closes #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,6 @@ The member needs to:
 
 1. Follow the [prerequisite installation instructions](#installing-prerequisites)
 2. [Start minikube](#start-minikube)
-3. [Install mkchain](./mkchain/README.md#install-mkchain)
 
 Then run:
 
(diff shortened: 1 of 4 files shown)


nicolasochem:
zerotier api is broken for me right now, the wait script is stuck at:

```
{"type":"internal","message":"Error updating member: json: cannot unmarshal string into Go struct field APIMember.hidden of type bool"}
```

If I remove `hidden: false` from the api call in the script it seems to work. Do you see the same?

nicolasochem (approved):
I verified by creating a zerotier chain and an invite chain and launching them in two namespaces on my machine. All containers came up properly and a full mesh was established.

PR #116, opened 2021-02-17
Remove reference to possible minikube bug (as per Andrew)

harryttd:


agtilden:
General point being that even if it is a bug in minikube we should probably avoid loose conjecture if we haven't gone as far as creating a [Short, Self Contained, Correct, Example](http://sscce.org/) - or better yet filing a ticket upstream.

PR #117, opened 2021-02-17
Public networks

nicolasochem:
This introduces a third mode of operation. We previously had zerotier-in-use true/false.

Now we have public/private/isolated networks. Description is in README.

The main addition is the capability to run mainnet or a testnet on tezos-k8s, and sync quickly thanks to snapshots.

Add a new container to import snapshot. Make the assmption of a public network will run in `rolling` mode.

The `config-generator` container is now built on top of the tezos container. The reason is that we need to extract the network data from the binary to put it in the `bootstrap_peers` section. Entrypoint is now a shell script which calls the previous python entrypoint.

To test on edo2net, the `master_328bc6c3_20210215194506` branch will work. Replace `v8-release` with this branch everywhere.

Then you need to submit `network: edo2net` in the values.yaml. We also add `--network` support to mkchain.

We switch the default protocol to edo2 because this is the one that runs in mainnet now.

Config-generator was previously using argparse to build config.json because it was originally part of mkchain. We replace it with a simple python object creation.

Black changed the formatting of some of Roland's crypto code for some reason.

If approved, I will be using "squash-and-merge" to merge.


charts/tezos/templates/baker.yaml
```diff
@@ -138,7 +138,7 @@ spec:
                 fieldPath: metadata.name
           - name: DAEMON
             value: endorser
-{{- if .Values.zerotier_in_use }}
+{{- if eq .Values.chain_type "private" }}
```
harryttd, line 141: IMO we should make the `chain_types` constants somehow. This way it is explicit what are acceptable values. And also so that we don’t have to keep referencing strings everywhere which can very easily introduce bugs. (e.g. hashs/hashes)

We can maybe have 3 (maybe just 2) vals: `is_private`, `is_isolated`, `is_public`. We can prob just go with the first 2. If NOT `private` AND NOT `isolated`, then it must be `public`. These vals get set by mkchain or whatever.

Or we can keep the `eq` checks in Helm templates but set “private” as a string Helm value. so `private: "private"`  and check based off of that constant.

config-generator/entrypoint.py
```diff
@@ -99,105 +99,34 @@ def get_node_config(
...
+                "listen-addr": ( net_addr + ":9732" if net_addr else "[::]:9732" )
+                }
+            }
+    if CHAIN_PARAMS["chain_type"] == "public":
+        node_config["network"] = CHAIN_PARAMS["network"]
+    else:
+        node_config["p2p"]["bootstrap-peers"] = bootstrap_nodes
+        node_config["network"] = {
+                "chain_name": chain_name,
+                "sandboxed_chain_name": "SANDBOZED_TEZOS",
```
harryttd, line 117: I doubt this matters one iota but just in case, it should be sandboxed with X not Z

charts/tezos/templates/baker.yaml
```diff
@@ -93,24 +93,6 @@ spec:
               name: config-volume
             - mountPath: /var/tezos
               name: var-volume
-        - image: "{{ .Values.tezos_k8s_images.baker_endorser }}"
```
harryttd, line 96: (Github did not put this comment the first time I made it on the file itself...)

From what I can see, there is no baker daemon running anymore. Was this intentional? Is it being run somewhere else? I couldn't tell where.

charts/tezos/templates/node.yaml
```diff
@@ -63,19 +63,37 @@ spec:
...
 {{- end }}
       initContainers:
-        - args:
-            - "-c"
-            - "[ -f /var/tezos/node/identity.json ] || (mkdir -p /var/tezos/node && /usr/local/bin/tezos-node identity generate 0 --data-dir /var/tezos/node --config-file /etc/tezos/config.json)"
-          command:
-            - /bin/sh
-          image: {{ .Values.images.tezos | quote }}
-          name: identity-job
+        - image: {{ .Values.tezos_k8s_images.config_generator }}
```
harryttd, line 66: I believe this will break Zerotier nets. As there will be a race condition between the bootstrap nodes and the other nodes, where if the other nodes get an ip before the Bnode, they won't be able to connect to the Bnode bec they won't know about the Bnode's ip.

snapshot-downloader/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 0
+else
+    echo "Did not find pre-existing data, importing blockchain"
+    mkdir -p ${node_dir}/data
+    echo '{ "version": "0.0.4" }' > ${node_dir}/version.json
+    cp -v /usr/local/share/tezos/alphanet_version ${node_dir}
+    snapshot_file=${node_dir}/chain.snapshot
+    curl -L -o $snapshot_file $snapshot_url
+    exec "${node}" snapshot import ${snapshot_file} --data-dir ${node_data_dir} --network $tezos_network --config-file /etc/tezos/config.json
+    find ${node_dir}
```
harryttd, line 28: does anything run after an `exec`?

nicolasochem: Duh! Thanks. You just solved this one too: https://github.com/midl-dev/tezos-on-gke/issues/44

charts/tezos/templates/baker.yaml
nicolasochem: I don't think we should have `is_isolated` and `is_private` because then they would not be mutually exclusive.

What do you recommend? Something like:

```
{{ $chain_type_private := "private }}
{{ $chain_type_isolated := "isolated" }}
{{ $chain_type_public := "public" }}
```

Then

```
{{- if eq .Values.chain_type $chain_type_private }}
```

?

harryttd: Yes something like that although I don't think that exact thing will work. Those variables won't be global to all templates.

I think there are a couple of options:
- Define global values: https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values
- Define helpers in a `.tpl` file in `charts/tezos/templates`. This will contain templates with something like this:
    ```yaml
    {{ define "tezos.public_chain" }}
    {{- "public" -}}
    {{ end }}
    ```
    Then we can compare on including the template. I think there is a way to treat the template like a function where you can pass a context in and let it spit back out a value like true/false that we can then just `if` on.
    
 Either way, probably not necessary for getting the F-beta net out the door.

charts/tezos/templates/node.yaml
nicolasochem: I moved the init containers again and fixed it.

charts/tezos/templates/baker.yaml
harryttd: let's iterate on this in another PR

PR #118, opened 2021-02-22
Pulumi eks2

elric1:
Let's discuss merging in the Pulumi code to ease deployment into cloud providers.
@nicolasochem: we started with your code. I took the liberty of changing "DO NOT MERGE" at the start of your commit message to "Initial" because it would look odd to have "DO NOT MERGE" prominently feature in the output of `git log --oneline`.


pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as fs from 'fs';
+import * as YAML from 'yaml'
+
+const { readdirSync } = require('fs')
+
+// Manual step: create a pulumi_values.yaml in the top level dir
+// with mkchain command:
+//   mkchain pulumi
+//
+const helm_values_file = fs.readFileSync('../pulumi_values.yaml', 'utf8')
```
harryttd, line 19: Running `pulumi` commands in top level of projects like mkchain is, results in error `no Pulumi project found in the current working directory`.

If you change dirs into pulumi, then it is looking for a values file a level up.

So right now you need mkchain in top level, then cd into pulumi, and then run pulumi commands.

Also, the values file that is being looked for should be the CHAIN_NAME env var, not hardcoded.

pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
...
+// For now, we set the number of EKS nodes to be the number of
+// Tezos bakers/nodes divided by 12.  We should fix this to
+// be a little more automagic.  Why 12?  It's a guess currently,
+// we should experiment a little to get a better number.  We add
+// three because that's the minimum.
+
+let numbakers = 0;
+let numnodes  = 0;
+
+if (helm_values["nodes"] != undefined) {
```
harryttd, line 48: This `if` block should be able to be simplified to:
```ts
const numBakers = helm_values.nodes?.baking?.length || 0
const numRegularNodes  = helm_values.nodes?.regular?.length || 0

const totalTezosNodes = numBakers + numRegularNodes
const desiredClusterCapacity = Math.round(totalTezosNodes / 8 + 3);
```

pulumi/package-lock.json
```diff
@@ -0,0 +1,807 @@
+{
```
harryttd, line 1: It seems there is both a package-lock (npm) and a yarn.lock (yarn). Should only need one. Depends on the default package mng we want to use.

pulumi/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+```shell
+mkchain --number-of-bakers 10 $CHAIN_NAME
+```
+
+## Use Pulumi to "make it so"
+
+```shell
+pulumi stack init my_name
+pulumi config set aws:region us-east-2
```
harryttd, line 85: Is this step needed being there is a `Pulumi.tqtezos.yaml` file that specifies it?

pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as awsx from "@pulumi/awsx";
+import * as aws from "@pulumi/aws";
+
+// Create a repository
+const repo = new awsx.ecr.Repository("tezos-k8s");
+
+import * as fs from 'fs';
+import * as YAML from 'yaml'
+
+const { readdirSync } = require('fs')
```
harryttd, line 13: I think this is unused

pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
...
+const helm_values_file = fs.readFileSync('../pulumi_values.yaml', 'utf8')
+const helm_values = YAML.parse(helm_values_file)
+
+const nginx_ingress_helm_values_file = fs.readFileSync('nginx_ingress_values.yaml', 'utf8')
+const nginx_ingress_helm_values = YAML.parse(nginx_ingress_helm_values_file)
+
+let images : {};
+images = helm_values['tezos_k8s_images'] || {};
+
+let imagelist = ["baker_endorser", "chain_initiator", "config_generator",
```
harryttd, line 28: For now this is prob fine, but we should prob have a dynamic way of getting the images.

@nicolasochem what do you say we make a docker dir now? :)

pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as eks from "@pulumi/eks";
+import * as k8s from "@pulumi/kubernetes";
+import * as awsx from "@pulumi/awsx";
+import * as aws from "@pulumi/aws";
```
harryttd, line 5: Seems unused import

pulumi/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+// Create a repository
+const repo = new awsx.ecr.Repository("tezos-k8s");
```
harryttd, line 8: I would personally put repo under all imports

harryttd:
Any formatting like prettier that could be applied to index.ts would be much appreciated.

harryttd:
Roland can you please add node_modules to the .gitignore. My editor keeps complaining the diff size is too huge. Maybe pulumi should get it's own .gitignore.

pulumi/index.ts
elric1: I'm using the CHAIN_NAME now in the current dir.

pulumi/index.ts
elric1: done, thanks.

pulumi/README.md
elric1: I've removed it, this is a generated file, we shouldn't check it in.

pulumi/index.ts
elric1: removed it.

pulumi/index.ts
elric1: I'd like to leave this for later?

pulumi/index.ts
elric1: did it.

harryttd:
In addition to formatting on the file which i mention above, does everyone agree we should keep the "case" of variables to camelCase which is the JS way? We could leave as it is but if users are meant to be able to contribute, perhaps we should do so.

elric1:
Okay, camelCase done.

pulumi/README.md
```diff
@@ -0,0 +1,111 @@
...
+- [kubectl](https://kubernetes.io/docs/reference/kubectl/kubectl/)
+- [helm](https://helm.sh/)
+- AWS CLI
+- nodejs
+- pulumi
+    - install pulumi
+    - create and account and tokens
+    - pulumi has some modules to install
+
+## Installing prerequisites
```
harryttd, line 31: Not needed if not specifying like top level README

pulumi/index.ts
harryttd: ☝️

pulumi/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "tezos-k8s",
+    "devDependencies": {
+        "@types/node": "^10.0.0"
+    },
+    "dependencies": {
+        "@pulumi/aws": "^3.0.0",
```
harryttd, line 7: can be removed

pulumi/package-lock.json
harryttd: ☝️

elric1: We went with yarn.lock.

pulumi/index.ts
```diff
@@ -0,0 +1,97 @@
...
+import * as YAML from 'yaml'
+
+const repo = new awsx.ecr.Repository("tezos-k8s");
+
+// Manual step: create a pulumi_values.yaml in the top level dir
+// with mkchain command:
+//   mkchain pulumi
+//
+
+let chainName = process.env.CHAIN_NAME || "pulumi";
```
harryttd, line 16: `chainName` is not changing so can be a `const`

pulumi/README.md
elric1: did it.

pulumi/index.ts
elric1: 13a7f40

pulumi/index.ts
```diff
@@ -0,0 +1,97 @@
...
+images = helmValues['tezos_k8s_images'] || {};
+
+let imagelist = ["baker_endorser", "chain_initiator", "config_generator",
+	         "rpc_auth", "wait_for_bootstrap", "zerotier"]
+
+for (let image of imagelist) {
+    images[image] = repo.buildAndPushImage("../" + image.replace(/_/g, "-"))
+}
+
+helmValues["tezos_k8s_images"] = images;
```
harryttd:
```suggestion
const tezosK8sImages = ["baker_endorser", "chain_initiator", "config_generator",  "rpc_auth", "wait_for_bootstrap", "zerotier"]
const taggedImages = {}

for (const image of tezosK8sImages) {
    taggedImages[image] = repo.buildAndPushImage("../" + image.replace(/_/g, "-"))
}

helmValues["tezos_k8s_images"] = taggedImages;
```

You don't need line 25 bec `helmValues` is coming from the values file that is gend by mkchain. And mkchain doesn't deal with the `tezos_k8s_images`. They are hardcoded in the chart's values.yaml.

pulumi/index.ts
```diff
@@ -0,0 +1,97 @@
...
+images = helmValues['tezos_k8s_images'] || {};
+
+const imagelist = ["baker_endorser", "chain_initiator", "config_generator",
+	           "rpc_auth", "wait_for_bootstrap", "zerotier"]
+
+for (let image of imagelist) {
+    images[image] = repo.buildAndPushImage("../" + image.replace(/_/g, "-"))
+}
+
+helmValues["tezos_k8s_images"] = images;
```
harryttd, line 34: I don't know where my previous comment here went, but this can be improved to:
```suggestion
const defaultHelmValuesFile = fs.readFileSync("../charts/tezos/values.yaml", 'utf8')
const defaultHelmValues = YAML.parse(defaultHelmValuesFile)

const helmValuesFile = fs.readFileSync(chainName + '_values.yaml', 'utf8')
const helmValues = YAML.parse(helmValuesFile)

const tezosK8sImages = defaultHelmValues["tezos_k8s_images"]
const pulumiTaggedImages = Object.entries(tezosK8sImages).reduce(
  (obj, [imageKey]) => {
    obj[imageKey] = repo.buildAndPushImage(`../${imageKey.replace(/_/g, "-")}`)
    return obj
  },
  {}
)

helmValues["tezos_k8s_images"] = pulumiTaggedImages
```

This gets all of the docker images specified in the default values.yaml file of the tezos chart. So no need to hardcode them in pulumi.

pulumi/index.ts
elric1: Your next suggestion overrides this one.

pulumi/index.ts
elric1: I've incorporated this suggestion.

nicolasochem (approved):
Thank you very much!

I have not tried yet but I see no downside on merging this so we all have a baseline to iterate on.

elric1:
The forced push only squashed and updated commit messages:
```
: arioch; git diff  fc1c4d6  69abb18 | cat
: arioch; 
```

Issue #46, opened 2020-11-17
move generate-constants to k8s jobs


nicolasochem:
We rely on locally installed docker to generate the constants by running short-lived commands on the tezos and flextesa container. This should mode to kubernetes jobs to reduce the HOWTO requirements.

Mkchain should then funnel the results out of kubernetes safely and back into a file, perhaps with `kubectl cp` or the python sdk equivalent.

brandisimus:
This is related to the docker issues, need to decide whether this is the right approach

nicolasochem:
I made the suggestion to funnel values out because I wanted to preserve the property that the same values.yaml should yield the same account/baker mapping with the same addresses.

Now that we have a mechanism to deterministically derive accounts from names, we can satisfy this property without extracting account keys, so we can drop this requirement.

brandisimus:
Closing the issue as per @nicolasochem above.

PR #121, opened 2021-02-24
Remove docker as a dependency of mkchain

### LINKED ISSUE #26: remove flextesa dependency
flextesa is used to generate the block value for config.json

this introduces an additional dependency and another docker image download. can the same thing be accomplished with tezos binaries?

### LINKED ISSUE #89: Research removing docker dependency in mkchain
Currently using docker to generate keys (tezos docker image) and creating a genesis block hash (in flextessa). This is an akward and unnecessary complication on the pipeline and ideally would be replaced with a simpler solution if possible. This ticket requires research wrt this possibility.

harryttd:
closes #89
closes #26 
Both of these closes were enabled by PR #105

Issue #46 has relation to this PR. Perhaps it should be closed and a new ticket opened regarding funneling generated constants inside k8s, outside to host land.

### DIFF
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -7,7 +7,7 @@ data:
       "nodes": {{ toJson .Values.nodes }},
       "chain_name": "{{ .Values.chain_name }}",
       "genesis_block": "{{ .Values.genesis.block }}",
-      "activation_account": "{{ (index .Values.accounts 0).name }}",
+      "activation_account": "{{ default "baker0" (first .Values.accounts).name }}",
       "timestamp": "{{ .Values.genesis.timestamp }}",
       "zerotier_in_use": {{ .Values.zerotier_in_use }},
       "proto_command": "{{ .Values.protocol.command }}",
--- a/mkchain/tqchain/mkchain.py
+++ b/mkchain/tqchain/mkchain.py
@@ -1,8 +1,5 @@
 import argparse
 import os
-import random
-import string
-import subprocess
 import sys
 from datetime import datetime, timezone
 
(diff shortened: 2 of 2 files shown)


elric1:
A general comment is: maybe we should have `mkchain` add comments to the values.yaml files to help people understand what they can edit?

harryttd:
I agree with Roland, but being as many things are changing bec of public nets, I think we should hold off until the values file format gets set in stone.

elric1:
I added #122 for the comments, it should be done later.

nicolasochem (changes requested):
I tried create with both a brand new values file and then overwriting an existing values files with default params and it worked.

I didn't try invite.

I have comment on the output itself.

Running with no parameters yields the result:

```
accounts: []
bootstrap_peers: []
chain_name: niconodocker
genesis:
  timestamp: '2021-02-26T18:41:05.485878+00:00'
images:
  tezos: tezos/tezos:v8-release
is_invitation: false
nodes:
  baking:
  - bake_for: baker0
    bootstrap: true
  regular:
  - {}
rpc_auth: false
zerotier_config:
  zerotier_network: null
  zerotier_token: null
zerotier_in_use: false
```

I feel like this is too much magic. It makes you assume that there are no accounts declared in your chain. but then what is "baker0"?

We should declare all accounts in values.yaml, and let the init container autogenerate an address for it if it's not specified.

Something like:

```
accounts:
  - name: baker0
bootstrap_peers: []
chain_name: niconodocker
genesis:
  timestamp: '2021-02-26T18:41:05.485878+00:00'
images:
  tezos: tezos/tezos:v8-release
is_invitation: false
nodes:
  baking:
  - bake_for: baker0
    bootstrap: true
  regular:
  - {}
rpc_auth: false
zerotier_config:
  zerotier_network: null
  zerotier_token: null
zerotier_in_use: false
```

If `accounts` is completely absent then we can let the magic happen:

```
bootstrap_peers: []
chain_name: niconodocker
genesis:
  timestamp: '2021-02-26T18:41:05.485878+00:00'
images:
  tezos: tezos/tezos:v8-release
is_invitation: false
nodes:
  baking:
  - bake_for: nochem0
    bootstrap: true
  regular:
  - {}
rpc_auth: false
zerotier_config:
  zerotier_network: null
  zerotier_token: null
zerotier_in_use: false
```

The above would generate an account named `nochem0` in the init container. This is a convenience for users that do not use mkchain. But mkchain should be explicit about which accounts it expects to exist.

Also README should be edited to drop docker as a requirement.

elric1:
I think that this PR is obsoleted by PR#205 which accomplishes the same thing in a slightly different and more comprehensive way.

harryttd:
Will be closed by #205

PR #128, opened 2021-03-02
Florence

harryttd:
- Uses specified bootstrap balances for every account
- Can set history mode of nodes. Defaults to rolling
- allows for non baking accounts to be part of genesis block
- puts chain parameters into values.yaml file allowing for direct injection of parameters.json into the node
- templatizes name of regular node and baker node in Helm templates

~At the present time, only florence_no_baking_accounts is working. florence_with_baking_accounts yields error on chain initiation:~
This was the old error:
```shell
<<<<2: 500 Internal Server Error
  [ { "kind": "permanent", "id": "proto.009-PsvJb9Pq.context.storage_error",
      "missing_key": [ "current" ], "function": "copy" } ]
Error:
  Storage error:
    Cannot copy undefined key 'current'.
  ```

### CODE REVIEW

File: config-generator/entrypoint.py [hunk 1]
```diff
@@ -212,15 +241,13 @@ def generate_node_config(node_argv):
     return node_config
 
 
-def get_parameters_config(bootstrap_accounts, bootstrap_mutez):
+def get_parameters_config(bootstrap_accounts):
+    default_balance = CHAIN_PARAMS["defualt_bootstrap_mutez"]
     parameter_config_argv = []
-    for bootstrap_account in bootstrap_accounts:
+    for account in bootstrap_accounts:
+        account_balance = account.get("balance") or default_balance
```
harryttd, line 248: If balance is 0, that should be set. If empty should default to default_balance

harryttd: does it need to validate if negative?

config-generator/entrypoint.py
```diff
@@ -295,28 +325,42 @@ def fill_in_missing_genesis_block():
...
+# 	baker<baker num>
 #
 # and fill in the details appropriately.
 
+
 def flatten_accounts():
     accounts = {}
-    for name, type, key in [[a["name"],a["type"],a["key"]] for a in ACCOUNTS]:
+    for account in ACCOUNTS:
+        name, type, key, balance, bootstrap_baker = itemgetter(
```
harryttd, line 336: hmm only realized that itemgetter will fail if missing any property. Should be the same though for previous code. I think values.yaml should always have 0 balance for nodes if no tez desired. Never just not having the property.

charts/tezos/values.yaml
```diff
@@ -38,3 +36,33 @@ nodes:
...
+  hard_storage_limit_per_operation: "60000"
+  test_chain_duration: "1966080"
+  quorum_min: 2000
+  quorum_max: 7000
+  min_proposal_quorum: 500
+  initial_endorsers: 1
+  delay_per_missing_endorsement: "1"
+  baking_reward_per_endorsement: ["200000"]
+  bootstrap_accounts:
+   - name: baker0
```
nicolasochem, line 68: is this expected to be here?

config-generator/entrypoint.py
nicolasochem: no

config-generator/entrypoint.py
```diff
@@ -147,15 +177,13 @@ def get_node_config(
...
+        account_balance = account.get("balance") or default_balance
         parameter_config_argv.extend(
-            [
-                "--bootstrap-accounts",
-                bootstrap_account,
-                bootstrap_mutez,
-            ]
+            ["--bootstrap-accounts", account["key"], str(account_balance)]
         )
     return generate_parameters_config(parameter_config_argv)
```
nicolasochem, line 189: Didn't we intend to get rid of all these `parser.add_argument(` below and replace it with a native pythoon object?

config-generator/entrypoint.py
nicolasochem: ok

baker-endorser/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM tezos/tezos:v8-release
+FROM registry.gitlab.com/nomadic-labs/tezos:amd64-florencenet-docker
```
nicolasochem, line 1: perhaps let's not put random branches on master? Instead, we should create a "deploy branch" if we want to record what we deployed for fbetanet/florencenet

charts/tezos/values.yaml
harryttd: It actually already exists in current code: https://github.com/tqtezos/tezos-k8s/blob/03edd3c5924426407a139f8c48320dabcf385f8f/charts/tezos/values.yaml#L31-L32

CI was failing on the Helm lint step, giving me a slice index range error. We are looking for the first item at index 0 for the activation account. So I think this would fail on master as well. My remove docker from mkchain branch fixes this bec it uses a diff Helm function to retrieve the first element.

nicolasochem (approved):
I tried isolated and private (zerotier) chains on my minikube, it worked.

We talked extensively about this PR and I also deployed fbetanet with some old version of this so I have high confidence the code is in a good state even though I didn't parse it all (especially not generate-config).

PR #130, opened 2021-03-03
make mkchain work in f-beta branch

nicolasochem:


mkchain/tqchain/mkchain.py
```diff
@@ -211,12 +212,17 @@ def main():
                         "name": account,
                         "key": keys[key_type],
                         "type": key_type,
+                        "bootstrap_baker": True,
+                        "bootstrap_balance": 50000000000000,
```
harryttd, line 216: the old default was "4000000000000". I assume it doesn't matter either way

mkchain/tqchain/mkchain.py
```diff
@@ -167,6 +167,7 @@ def main():
             "zerotier_network": args.zerotier_network,
             "zerotier_token": args.zerotier_token,
         },
+        "is_old_accounts_parameter_format": True,
```
harryttd, line 170: Maybe I should have made the default in Florence to be true, this way the florence with baking needs to specify this prop as False, and we wouldn't need the prop here in mkchain.

harryttd: If you agree with this, you can change the default value to True in configs.yaml. Then florence with baking should get `is_old_accounts_parameter_format: false`. And florence without baking doesn't need the prop. Then you can get rid of this addition in mkchain.

nicolasochem: It's fine this way, and easy enough to change later.

harryttd (approved):

PR #149, opened 2021-03-17
Add cloudwatch

elric1:


charts/tezos/templates/baker.yaml
```diff
@@ -80,6 +80,9 @@ spec:
             - /etc/tezos/config.json
           command:
             - /usr/local/bin/tezos-node
+          env:
+            - name: TEZOS_EVENTS_CONFIG
```
harryttd, line 84: - I think this env var should be put in values.yaml to be templatized, as it is being used multiple times and is the same value every time.
- Do we always want this env var set? Maybe just for development and not prod? Then we would need a flag. Or probably, if we templatize it, the var could just be an empty string.

I'm not sure how to structure it in values.yaml. You probably know better than me Roland as there might be other potential vars that you would like to set.

So maybe something along the lines of:
values.yaml
```suggestion
tezos_node_config:
  logging:
       name: TEZOS_EVENTS_CONFIG
       value: "file-descriptor-path:/dev/stdout?level-at-least=debug"
```
The value of logging could be empty str for user/production purposes.

And then in baker.yaml something like:
```yaml
env:
  env:
   - name: {{ .Values.tezos_node_config.logging.name }}
     value: {{ .Values.tezos_node_config.logging.value | toJson }}
```

logger/entrypoint.sh
```diff
@@ -0,0 +1,129 @@
...
+
+    if [ ! -z "$PREV_RUN" -a "$HASH" != "$PREV_RUN" ]; then
+	echo "$BLOCKS" | sed 's/^.* \([^ ]* [^ ]*\)/\1/'
+    else
+	echo "$BLOCKS"
+    fi
+}
+
+LAST=
+LAST_LEVEL=0
```
harryttd, line 49: not sure if you still want this here but it is unused

elric1: Yes, I've eliminated this one. I'm going to close this PR and open a new one to look at which is based on PR#177 which is hopefully close to complete.

PR #164, opened 2021-03-23
Readme updates 

harryttd:
Please review carefully

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,22 +130,12 @@ Set [unbuffered IO](https://docs.python.org/3.6/using/cmdline.html#envvar-PYTHON
 export PYTHONUNBUFFERED=x
 ```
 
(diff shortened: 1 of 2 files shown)


charts/rpc-auth/templates/rpc-auth.yaml
```diff
@@ -134,7 +134,7 @@ spec:
             - name: REDIS_PORT
               value: {{ .Values.redis.port | quote }}
             - name: TEZOS_RPC_SERVICE
-              value: tezos-node-rpc
+              value: tezos-baking-rpc
```
harryttd, line 137: This is why we need more templating. Will make a ticket

elric1 (approved):
Looks good.

PR #166, opened 2021-03-24
add Nomadic's and BCD tezos indexers

nicolasochem:
Both indexers are able to index a private chain.

~BCD indexer needs private images with env variables to set the hostname for elastic/db/rabbitmq. Next release of BCD hub will have them, until then it is necessary to build your own containers:~

~baking-bad/bcdhub#554~

The indexer need to be activated in values.yaml and does not run by default. The default values.yaml has an explanation in comments.


charts/tezos/templates/bcd_indexer.yaml
```diff
@@ -0,0 +1,146 @@
...
+    metadata:
+      labels:
+        app: {{ .Values.bcd_indexer_statefulset.name }}
+        appType: tezos
+    spec:
+      containers:
+        - image: bakingbad/bcdhub-api:3.5
+          name: api
+          env:
+            - name: BCD_ENV
```
harryttd, line 23: It's prob preferable to put all env vars in this file and nomadic indexer into config maps

charts/tezos/templates/bcd_indexer.yaml
```diff
@@ -0,0 +1,146 @@
...
+kind: StatefulSet
+metadata:
+  name: {{ .Values.bcd_indexer_statefulset.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  podManagementPolicy: Parallel
+  selector:
+    matchLabels:
+      app: {{ .Values.bcd_indexer_statefulset.name }}
+  serviceName: tezos-node
```
harryttd, line 12: IMHO, we should templatize serviceName. At the moment, we make the service name based on the baker/node stateful sets' name.

We should probably instead create a new prop in values.yaml called services that contains names for our services in static.yaml. Then the services and statefulset serviceName can reference those props. I'm recommending this because it would be a little weird to reference baker/node statefulset names for the serviceName in the indexers.

charts/tezos/templates/bcd_indexer.yaml
```diff
@@ -0,0 +1,146 @@
...
+            - name: RABBITMQ_HOSTNAME
+              value: localhost
+            - name: DB_HOSTNAME
+              value: 127.0.0.1
+            - name: RABBITMQ_DEFAULT_USER
+              value: guest
+            - name: RABBITMQ_DEFAULT_PASS
+              value: guest
+            - name: SANDBOX_NODE_URI
+              value: http://{{ .Values.index.target }}:8732
```
harryttd, line 44: The target can also be templatized in values.yaml instead of hardcoded `tezos-baking-node....`. Then using Helm [`tpl`](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) the value here could be templatized.

charts/tezos/templates/bcd_indexer.yaml
```diff
@@ -0,0 +1,146 @@
...
+            - name: RABBITMQ_HOSTNAME
+              value: localhost
+            - name: DB_HOSTNAME
+              value: 127.0.0.1
+            - name: RABBITMQ_DEFAULT_USER
+              value: guest
+            - name: RABBITMQ_DEFAULT_PASS
+              value: guest
+            - name: SANDBOX_NODE_URI
+              value: http://tezos-node-0.tezos-node:8732
```
harryttd, line 69: same templating here, etc. for all the diff places

charts/tezos/templates/bcd_indexer.yaml
harryttd: In fact the fqdn can also be part of the services prop. So you could have service's name and fqdn for all the places that need one or the other format.

charts/tezos/templates/static.yaml
```diff
@@ -6,8 +6,9 @@ metadata:
 spec:
   ports:
     - port: 8732
+      name: rpc
   selector:
-    app: tezos-baking-node
+    app: tezos-node
```
harryttd, line 11: what happens if there is only a baker in the cluster and no reg nodes?

chain-init needs to look for a baker, not a reg node. Hmmmm, well chain-init is a job and as of now every pod has all of the keys imported on it (for sk's this should change). So it would likely still activate the chain even if there is only a single reg node.

Shouldn't a baker be a more reliable source for RPC?

We should be able to select multiple apps if desired.

harryttd: Just tried on latest commit 02a4bb3 on branch indexer_new, and in minikube. It breaks bec job can't find the activation key.
```
>>>>1: http://tezos-node-rpc:8732/chains/main/blocks/genesis/protocols
<<<<1: 200 OK
{ "protocol": "PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex",
"next_protocol": "PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex" }
Error:
Erroneous command line argument 9 (tqtezos_baker_0).
no secret_key alias named tqtezos_baker_0
cannot read file (Unix.Unix_error(Unix.ENOENT, "open", "tqtezos_baker_0"))
Error while parsing URI: SK_URI needs a scheme
```

charts/tezos/templates/bcd_indexer.yaml
```diff
@@ -0,0 +1,129 @@
...
+        name: indexer-db
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: "300Gi"
+---
+apiVersion: v1
+kind: ConfigMap
```
harryttd, line 91: Unused configmap

Issue #161, opened 2021-03-22
wait-for-bootstrap: Should attempt to hit the endpoint of the bootstrap node before waiting


harryttd:
If you are scaling up nodes, the bootstrap nodes are already running. There is no point in delaying new nodes from spinning up. They should immediately check if a bootstrap node is up and running.

PR #168, opened 2021-03-24
wait-for-bootstrap: wait for a configured node

elric1:
Previously, we unconditionally waited for tezos-baking-node-0.  Now,
we take the configured list of nodes and wait until at least one of
them is up and running.

Also hit #161 as a drive-by.


harryttd (approved):

nicolasochem (approved):
Thanks. This removes the last hardcoding of tezos-bootstrap-node-0 in the code.

PR #169, opened 2021-03-24
refactor charts

elric1:


nicolasochem:
Chart refactoring looks OK. It's worth making a dedicated PR for it. I haven't tried it since I have reservations on the second commit.

Just one question, a lot of the newly added templates are wrapped into if statements ("if wait for bootstrap", "if download snapshot"). Did you consider leaving the if statements in the final yaml files for readability?

The second commit breaks the "known network" use case: if you want to launch an edo2net or mainnet node, config-generator will run `tezos-node config init` in order to extract data that is needed but trapped in the binary, such as the chain name and bootstrap peers. So it can't be merged as is. One solution is to move this step into its own init container that will be built on top of tezos. Private chains will never need this container so you can skip building it.

elric1:
I've removed the second commit. I only put it in this PR because it will conflict if I based it off head. Let's consider this one on its own. I'll keep on working on the image one separately.

I did consider putting the conditionals in the final YAML files and, in fact, that is how I originally did it. I found it to be less readable and also more repetitive because the same if statements would be repeated in both baker.yaml and node.yaml.

nicolasochem:
I tried a zerotier invite and:

```
Events:
  Type     Reason            Age                 From                    Message
  ----     ------            ----                ----                    -------
  Normal   SuccessfulCreate  65s                 statefulset-controller  create Claim var-volume-tezos-node-0 Pod tezos-node-0 in StatefulSet tezos-node success
  Warning  FailedCreate      24s (x14 over 65s)  statefulset-controller  create Pod tezos-node-0 in StatefulSet tezos-node failed error: Pod "tezos-node-0" is invalid: spec.initContainers[2].name: Duplicate value: "get-zerotier-ip"
```

nicolasochem:
The above happened on the invite chart and this is the event log for tezos-node statefulset.

if you run mkchain with --zerotier-network and --zerotier-token, it will give you a file whatever_invite_values.yaml

if you pass this file to helm template, I expect  that you will see a duplicate init container in the node staetefulset

elric1:
Yes, thanks. I did that and found the offending line. I included the zerotier template twice by default.

I then tested again with rpc_auth set to true---and also with the snapshot-downloader. These appear to generate the correct charts, now.

nicolasochem (approved):
Thank you. I created a zerotier chain with mkchain and was able to spin it up and join it from another workspace, and I have no reason to believe anything else is broken.

charts/tezos/templates/_containers.tpl
```diff
@@ -0,0 +1,180 @@
...
+    - containerPort: 9732
+      name: tezos-net
+  volumeMounts:
+    - mountPath: /etc/tezos
+      name: config-volume
+    - mountPath: /var/tezos
+      name: var-volume
+{{- end }}
+
+{{- define "tezos.container.baker" }}
```
harryttd, line 94: I know I'm very late to the party here but just wanted to throw out, that being only baker is is running baker/endorser, why not have this explicit there. It can still `include "tezos.localvars.pod_envvars"`.  

Also, was there any more discussion regarding having the `if` checks in the templates themselves?

PR #171, opened 2021-03-25
make storage size parametrizable

nicolasochem:
My immediate need is to deploy an archive node; I need 300Gi of storage.

### DIFF
--- a/charts/tezos/templates/baker.yaml
+++ b/charts/tezos/templates/baker.yaml
@@ -103,5 +103,5 @@ spec:
           - ReadWriteOnce
         resources:
           requests:
-            storage: "15Gi"
+            storage: {{ .Values.baker_statefulset.storage_size }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


elric1 (changes requested):
One comment that I have is that perhaps the size should be shared between the two types of node because I get the impression that it's really the `tezos-node` container that needs the space---and it's the size of the chain that drives the requirement.

nicolasochem:
Not always, for example you could have a sentry model where your important baker node is running in private node and protected by sentries. You may want an archive node in the core of your topology, but the sentries should be rolling since they are just passing p2p messages along.
You could even be defining custom node types and helm would create statefulsets. In doubt, I'd say let us give ourselves flexibility.

harryttd (approved):

elric1:
In this case, though, wouldn't you want to attach the storage size to the instance rather than the statefulset definition?

elric1 (approved):
We chatted offline and I'm convinced now.

PR #176, opened 2021-03-29
Add detail for development readme

orcutt989:
This PR updates `development.md` to add more instructions about building images with devspace needed for new version. Adds instructions for updating helm charts and running charts to point at local repo instead of public repo.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,8 +1,12 @@
-- [Prerequisites](#prerequisites)
+- [Helm Chart Development](#helm-chart-development)
+  - [Prerequisites](#prerequisites)
 - [Using devspace](#using-devspace)
   - [Notes](#notes)
+  - [Building images](#building-images)
 - [Helm Charts](#helm-charts)
   - [Creating Charts](#creating-charts)
+  - [Update charts for local mainnet](#update-charts-for-local-mainnet)
+  - [Run charts for local mainnet](#run-charts-for-local-mainnet)
   - [Notes](#notes-1)
 - [Creating Docker Images](#creating-docker-images)
 - [Releases](#releases)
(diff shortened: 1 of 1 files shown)


nicolasochem (changes requested):
the development already has instructions to do `devspace build`, you should update these or make a new section, but not have it mentioned in 2 places like now.

orcutt989:
> the development.md file already has instructions to do `devspace build`, you should update these or make a new section, but not have it mentioned in 2 places like now.

Fixed

nicolasochem:
But you should still add `--skip-push` ie.

> If you would like to build all of our images without using Devspace to deploy (you might want to do a helm install instead), you can run `devspace build -t dev --skip-push`.

orcutt989:
> But you should still add `--skip-push` ie.
> 
> > If you would like to build all of our images without using Devspace to deploy (you might want to do a helm install instead), you can run `devspace build -t dev --skip-push`.

Oops! Thanks. Fixed!

DEVELOPMENT.md
```diff
@@ -77,6 +80,18 @@ Regarding chart dependencies, Chart.yaml should not specify a dependency version
 
 Being that all charts are bumped to the same version on release, the parent chart will get the latest version of the dependency by default (which is the same as its own version) when installing via our Helm chart [repo](https://github.com/tqtezos/tezos-helm-charts).
 
+## Update charts for local mainnet
+
+```
+helm dependency update charts/tezos
+```
```
nicolasochem, line 87: This is also redundant as it is already mentioned in the Note just below.

DEVELOPMENT.md
```diff
@@ -77,6 +80,18 @@ Regarding chart dependencies, Chart.yaml should not specify a dependency version
...
+
+## Run charts for local mainnet
+
+```
+helm install tezos-mainnet charts/tezos --namespace tqtezos --create-namespace
+```
```
nicolasochem, line 93: This is not only for local mainnet. This is to install a chart from the filesystem instead of a helm repo. A more accurate description would be:

## Run local development chart

Instructions as per README install the latest release of tezos-k8s helm chart from a helm repository. To install a development version of a tezos chart in the `charts/tezos` directory instead, run:

nicolasochem (changes requested):

nicolasochem (approved):
thank you

PR #177, opened 2021-03-30
Use the provided image for all tezos binaries

elric1:
To do this, we switch from making containers based on
tezos/tezos:v8-release to making containers that either:

	1. use the provided tezos image and execute
	   a script which is passed in via the
	   statefulset configuration, or

	2. are not based on the tezos image and do not
	   contain tezos binaries.  These containers
	   prepare configuration, files, etc., for the
	   later containers.


nicolasochem (changes requested):
This would close https://github.com/tqtezos/tezos-k8s/issues/120 as well.

I tried running mkchain with no parameters and it failed with the following helm error:

```
[venv] nochem@peck ~/workspace/tezos-k8s () $ rm nico_values.yaml 
[venv] nochem@peck ~/workspace/tezos-k8s () $ mkchain nico
Generating vanity chain id
Generating keys for account baker0
Wrote chain creation constants to /home/nochem/workspace/tezos-k8s/nico_values.yaml
[venv] nochem@peck ~/workspace/tezos-k8s () $ helm install --replace nico charts/tezos --values nico_values.yaml --namespace tqtezos --create-namespace
Error: YAML parse error on tezos-chain/templates/baker.yaml: error converting YAML to JSON: yaml: line 22: mapping values are not allowed in this context
```

It's not immediately obvious what is causing this, line 22 looks fine to me.

utils/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# Note: we install build deps for pip, then remove everything after pip install.
+RUN	apk add --no-cache --virtual .build-deps gcc py3-pip		\
+				python3-dev musl-dev libffi-dev make	\
+     && pip install base58 pynacl					\
+     && apk del .build-deps						\
+     && apk add jq netcat-openbsd curl python3 py3-pip
+
+COPY commitment-params.json /
+COPY config-generator.py /
+COPY config-generator.sh /
```
nicolasochem, line 13: Did you forget to check this file in?

nicolasochem (changes requested):
I cannot build utils container due to missing shell script.

utils/Dockerfile
elric1: Yes, in fact, I did forget this one.

nicolasochem (changes requested):
Thanks for this.

I tried:

* mkchain with no parameters: ok
* mkchain with 2 bakers: ok
* boot florence from snapshot failed silently to import snapshot. Chain started syncing from scratch.

It looks like we are missing another file in source control `snapshot-importer.sh`

nicolasochem (changes requested):
with latest commit, I was able to launch florencenet and have it sync from bootstrap.

But now, I saw that invite is broken.

I created a zerotier chain: `mkchain nico --zerotier-network $ZT_NET --zerotier-token $ZT_TOKEN`

`nico_values.yaml` deployed fine but `nico_invite_values.yaml` gave me this error in the configure step:

```│ + CMD=config-generator                                                                                                                                                                                                                                     │
│ + chmod -R 777 /var/tezos/client                                                                                                                                                                                                                           │
│   Importing keys for account: baker0                                                                                                                                                                                                                       │
│     Public key specified                                                                                                                                                                                                                                   │
│     Appending public key: edpkunvwGbaZQx6u1qL6pdGg97BTjBHt9KLV8ysz6hupd8BauKE2DJ                                                                                                                                                                           │
│   Appending public key hash: tz1fymSaKG25y2dsdQGwPcY2uS2isjsNZtST                                                                                                                                                                                          │
│   Account key type: public                                                                                                                                                                                                                                 │
│   Account bootstrap balance: 4000000000000                                                                                                                                                                                                                 │
│   Is account a bootstrap baker: True                                                                                                                                                                                                                       │
│   Writing /var/tezos/client/secret_keys                                                                                                                                                                                                                    │
│   Writing /var/tezos/client/public_keys                                                                                                                                                                                                                    │
│   Writing /var/tezos/client/public_key_hashs                                                                                                                                                                                                               │
│ Starting config.json file generation                                                                                                                                                                                                                       │
│   File "/config-generator.py", line 490, in <module>                                                                                                                                                                                                       │
│   File "/config-generator.py", line 110, in main                                                                                                                                                                                                           │
│ + jq -r '.regular."tezos-node-0".bake_using_account'                                                                                                                                                                                                       │
│ + echo '{' '"baking":' '{},' '"regular":' '{' '"tezos-node-0":' '{' '"config":' '{' '"shell":' '{' '"history_mode":' '"rolling"' '}' '}' '}' '}' '}'                                                                                                       │
│ + my_baker_account=null                                                                                                                                                                                                                                    │
│ + '[' null '=' null ]                                                                                                                                                                                                                                      │
│ + my_baker_account=tezos-node-0                                                                                                                                                                                                                            │
│ + echo tezos-node-0                                                                                                                                                                                                                                        │
│ stream closed                                                                        
```

And the node started with the default config and started downloading mainnet from scratch.

I suspect something is broken with the is_invitation selector.

utils/snapshot-downloader.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+#
+# If we are snapshotting, we download the snapshot here because the snapshot
```
harryttd, line 4: This comment should be removed because there is no standalone snapshot downloader image. This script is being run via the utils image that has curl.

elric1: I've created Issue #190 to track writing the better documentation that we need. And, I've removed this comment which is misleading.

PR #186, opened 2021-04-28
Add cloudwatch on 177

elric1:
I rebased the PR onto PR#177 which is almost ready to merge. This will make it a bit easier to review.


elric1:
Hmmm, maybe not _that_ much easier! If you are interested, please have a look at the commits after 683cd58 because those are the only ones about which this PR actually is. I'll create a new PR once we merge 177 which is cleanly against master.

harryttd:
Hey @elric1 
What is the status of this PR? I'm going through our PR list to get the current lay of the land.

elric1:
Most of these commits were squshed into `1f8c4984d1312`. Only the latest 4 are not. Of those, `03dac19` is not general enough for the main branch and `d3e27c1` has been made part of the name-the-nodes PR#223. So, in this PR, we have `d3e27c1` and `9fd4ce3` which add cloudwatch support to the pulumi bits.
I think that this is useful code and I can make a new PR if we want to incorporate it?

harryttd:
I see that logger.sh is in #223 
I'm not sure that fluentbit should be part of tezos-k8s. It is aws specific. And not general like prometheus would be.

I know it is confusing to have pulumi in tez-k8s and tqinfra but we are working on modularizing it

orcutt989:
I believe I've already implemented this in TQinfra. All pod logs are shipped to cloudwatch.

elric1:
I've been using the pulumi code in tezos-k8s to deploy tests. If we think that tqinfra is a full replacement for the pulumi code in tezos-k8s, then maybe we should delete it? I can look into this, after playing a little bit with tqinfra, if we agree that it's a good idea. If I do this, then I'll make sure that tqinfra subsumes the use cases that I've been playing with.

orcutt989:
@elric1 could you check out the cloudwatch streams for the tqinfra stack and confirm that its doing what you need for all the chains already? Perhaps thats what you meant by subsume. 🤔

elric1:
I'm going to close this one. I'll submit a new one in a minute which is the current work minus what I've already put in and rebased onto master.

elric1:
I'm going to close this one. I'll submit a new one in a minute which is the current work minus what I've already put in and rebased onto master.

PR #187, opened 2021-05-03
Fix vms per node

elric1:
When we changed the format for representing nodes from a list to an object, my pulumi code to count them and scale the network appropriately broke.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -143,6 +143,12 @@ protocol:
   # tezos/tezos:v8-release docker image.
   command: 008-PtEdo2Zk
 
+#
+# Pulumi values:
+
+pulumi:
+  vms_per_node: 12
+
 # # Activate a new chain from genesis. Note, this is only for when your chain does
 # # not exist yet. For subsequent deploys, remove this.
 #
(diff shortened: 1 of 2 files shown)


nicolasochem:
cc @innov8r 

We are currently discussing a refactor of all the pulumi code between TQinfra, teztnets and @elric1 's use case here.

charts/tezos/values.yaml
```diff
@@ -143,6 +143,12 @@ protocol:
   # tezos/tezos:v8-release docker image.
   command: 008-PtEdo2Zk
 
+#
```
harryttd, line 146: I'm thinking that perhaps pulumi values should go into a diff file. Reason being, this values.yaml is specifying the details of the actual chain, not the infra.

Pulumi is a way to specify how to deploy your infra for your chain.

I think index.ts should pull in a separate values.yaml for the infra values. This file could be in `pulumi/`.

harryttd:
@elric1 If you could please address the comment then this PR could be merged. Regardless of our refactoring of pulumi across diff repos this can still be merged now.

charts/tezos/values.yaml
elric1: I'm implementing this in PR #236 and closing this PR.

elric1:
I'm implementing this in PR #236 and closing this PR.

Issue #175, opened 2021-03-26
Investigate/Implement rpc-auth chart being its own standalone chart


harryttd:
The instigating problem for this ticket is pulumi  seems broken and can't handle a chart that has dependencies.

Issue seems related to: https://github.com/pulumi/pulumi-kubernetes/issues/861

harryttd:
Chart has been standalone for a long time at this point

PR #188, opened 2021-05-03
remove RPC auth from dependencies of tezos-k8s

nicolasochem:
Pulumi won't deploy the tezos chart from submodule when the dependency exists
(even if its condition is set to false).

To enable rpc-auth, it is now necessary to explicitly install the chart.

Since the `rpc_auth` condition in tezos helm chart was only used for the
conditional rpc-auth dependency, we remove it everywhere.


harryttd:
Related to issue #175

harryttd (approved):

PR #189, opened 2021-05-03
Make tezos node's sts's and svc's conditional

harryttd:
This became necessary due to pulumi erroring out if the svc for the sts's existed without any sts replicas.
Also rpc svc can select bakers and regular nodes.


nicolasochem (approved):
thanks, it works for me

PR #191, opened 2021-05-04
User activated upgrade support

nicolasochem:
It is possible to force the tezos node to switch protocol at a given
level. This is part of the network config section.

Example:
```
   "network": {
     "chain_name": "nico",
     "genesis": {
       "block": "BMFonSBNMfDf6MFRAw25xsCKhxx6f1gUsPr6KWnNZaviKu6Ak5D",
       "protocol":
"PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex",
       "timestamp": "2021-05-02T01:25:17.944261+00:00"
     },
     "user_activated_upgrades": [
       {
         "level": 5,
         "replacement_protocol":
"ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK"
       }
     ],
     "sandboxed_chain_name": "SANDBOXED_TEZOS",
     "default_bootstrap_peers": [],
     "genesis_parameters": {
       "values": {
         "genesis_pubkey":
"edpkvFcxq7NYu2PCYD9pjABSZMK3ypVPnYwCPpRZCpwpsRuYvHA8ZU"
       }
     }
   }
 }
```

To make it work in tezos-k8s, the following was done:

1. support for running baker and endorser binaries for several protocols
   at once. This is done by converting the `protocol` value into
   `protocols`. This is great, since it also enables support of seamless
   transition between protocol during real mainnet upgrades.
2. replace the ad-hoc build of the network config with a pass-through of
   every key/value pair passed in values.yaml `node_config_network`
   value into the node's network config. This enables support of any
   network feature that we don't know of yet, in addition to the
   user-activated upgrade which is of interest here. [Here is an example of this feature being used in galpha2net](https://github.com/tqtezos/teztnets/pull/12/files#diff-887d7eefee11c072829a52ea5d64be0d91e6d20cbdb07553846485caa763cbc6R9-R11).

Maybe also `sandboxed_chain_name` should be brought into this
passthrough instead of being hardcoded?

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -143,15 +143,21 @@
 {{- end }}
 
 {{- define "tezos.container.baker" }}
-- image: "{{ .Values.images.tezos }}"
+{{- range .Values.protocols }}
+- image: "{{ $.Values.images.tezos }}"
   command:
     - /bin/sh
   args:
     - "-c"
     - |
-{{ tpl (.Files.Get "scripts/baker-endorser.sh") . | indent 6 }}
+{{- /*
+Below set is a trick to get the range and global context. See:
+https://github.com/helm/helm/issues/5979#issuecomment-518231758
+*/}}
+{{- $_ := set $ "command_in_tpl" .command }}
+{{ tpl ($.Files.Get "scripts/baker-endorser.sh") $ | indent 6 }}
   imagePullPolicy: IfNotPresent
-  name: baker
+  name: baker-{{ lower .command }}
   volumeMounts:
     - mountPath: /etc/tezos
       name: config-volume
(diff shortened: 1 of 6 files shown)


harryttd (approved):

charts/tezos/templates/_containers.tpl
```diff
@@ -143,15 +143,21 @@
 {{- end }}
 
 {{- define "tezos.container.baker" }}
```
elric1, line 145: We might s/baker/bakers/ because the template now makes N bakers.

elric1: (same for endorsers.)

charts/tezos/values.yaml
```diff
@@ -131,10 +131,13 @@ expected_proof_of_work: 26
 node_config_network:
   chain_name: mainnet
 
-protocol:
-  # Set the protocol binary. The protocol must be recognized by the
-  # tezos/tezos:v8-release docker image.
-  command: 008-PtEdo2Zk
+protocols:
```
elric1, line 134: Maybe mention here the "user-activated-upgrades" config?

elric1 (approved):
These comments are suggestions, none of them is important enough to hold up the review. I will approve regardless of whether you decide to take action or not.

charts/tezos/values.yaml
nicolasochem: This really belongs to tezos documentation. It looks like there are plenty of hidden ways to tune the node's `network` config that ought to be documented.

PR #195, opened 2021-05-12
Add ability to deploy a tzkt indexer

### LINKED ISSUE #5: provide tzkt indexer
provide a deployment description to run [tzkt](https://github.com/baking-bad/tzkt) using the private chain config.

this deployment should be an optional feature for each cluster participating in the chain. every participant in the chain has the option of running their own indexer.

harryttd:
closes #5

### DIFF
--- a/charts/tezos/templates/tzkt_indexer.yaml
+++ b/charts/tezos/templates/tzkt_indexer.yaml
@@ -35,9 +35,10 @@ spec:
           - "-c"
           - |
             CHAIN_NAME="{{ .Values.node_config_network.chain_name }}"
+            SNAPSHOT_URL="{{ .Values.indexer.config.db_snapshot_url }}"
             if [ ! -d "/tmp/indexer-db/pgdata" ]; then
-              echo "Downloading $CHAIN_NAME db snapshot..."
-              curl -Lfs -o "/tmp/db-snapshot/$CHAIN_NAME.backup" {{ .Values.indexer.config.db_snapshot_url }}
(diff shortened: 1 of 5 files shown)


charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -0,0 +1,183 @@
...
+              touch "$ENV_FILE"
+
+              # Get the chain id
+              CLIENT="/usr/local/bin/tezos-client -E {{ .Values.indexer.rpc_url }}"
+              until $CLIENT rpc get /version; do sleep 2; done
+              CHAIN_ID=$($CLIENT rpc get /chains/main/chain_id)
+              echo "Got chain id: $CHAIN_ID"
+              echo "export TZKT_TezosNode__ChainId=$CHAIN_ID" >> "$ENV_FILE"
+      containers:
+        - image: bakingbad/tzkt-api:latest
```
harryttd, line 67: Is it ok to be pulling `latest` for tzkt images? Or i should pin it

nicolasochem (changes requested):
It works fine on my machine.

For our most immediate tqinfra use case, we will probably want both tzkt and nomadic indexer side by side, so downstream users can pick their favorite indexer. Both of them should be set up against the same archive node.

`.Values.indexer` should be changed into a list `.Values.indexers` everything else being equal. Each item of the list deploys a statefulset of one.

nicolasochem (approved):
I verified again and it still works. Thank you!

PR #196, opened 2021-05-13
genesis contracts and commitment URL support + remove hard-coded commitments

nicolasochem:
Commitments are precursor to faucet accounts. we had previously committed them in here for the sake of getting florencenet out, but they really belong in teztnets (and in fact the faucet itself belongs there too).

It is not possible to pass such a large object with helm values, so we implement a mechanism to download them from a random http location. On teztnet's side, I am putting this in source code and it gets deployed in a [s3 bucket whose url is then passed to helm](https://github.com/tqtezos/teztnets/blob/main/index.ts#L86-L113)

A mechanism to inject contracts at genesis is also included. It works very similarly to the commitments. [example](https://github.com/tqtezos/teztnets/tree/main/bootstrap_contracts)

Also changed logging behavior a bit: when passing commitments, parameters.json is so large than debugging is impossible. I am no longer printing it to the logs, and I also limit the output of activation to 200 lines.

Granadanet will be deployed this way so it's a good idea to merge soon.

(I will squash and merge since commit history is messy)


harryttd:
I did not test but the code looks good to me

harryttd (approved):

nicolasochem:
Yes it's a bit hard to test since you need http endpoints. I don't have a good minikube story for this feature, I suppose in this case it's necessary to fall back to injecting files inside the docker build. Thanks for reviewing.

PR #199, opened 2021-05-13
preliminary work for private mode support

nicolasochem:
* allow arbitrary config from values.yaml to be set in the node. This
  allow for setting private mode from values.yaml and this setting is
  transparently passed to the node config

* do not fail silently when generate-config.py crashes

* scour both regular and baking nodes for bootstrap nodes in
  wait-for-bootstrap. If you have only non-baking bootstrap nodes, the
  init script would otherwise fail.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -1,4 +1,5 @@
 import argparse
+import collections.abc
 import json
 import os
 import socket
@@ -445,6 +446,19 @@ def get_genesis_pubkey():
             raise Exception("ERROR: Couldn't find the genesis_pubkey")
         return genesis_pubkey
 
+def recursive_update(d, u):
+    '''
+    Recursive dict update
+    Used to merge node's config passed as chart values
+    and computed values
+    https://stackoverflow.com/a/3233356/207209
+    '''
+    for k, v in u.items():
+        if isinstance(v, collections.abc.Mapping):
+            d[k] = recursive_update(d.get(k, {}), v)
+        else:
+            d[k] = v
+    return d
 
 def create_node_config_json(
     bootstrap_peers,
@@ -452,7 +466,8 @@ def create_node_config_json(
 ):
     """ Create the node's config.json file """
 
-    node_config = {
+    values_node_config = MY_NODE.get("config", {})
+    computed_node_config = {
         "data-dir": "/var/tezos/node/data",
         "rpc": {
             "listen-addrs": [f"{os.getenv('MY_POD_IP')}:8732", "127.0.0.1:8732"],
@@ -461,9 +476,9 @@ def create_node_config_json(
             "bootstrap-peers": bootstrap_peers,
             "listen-addr": (net_addr + ":9732" if net_addr else "[::]:9732"),
         },
-        "shell": MY_NODE.get("config", {}).get("shell", {}),
         # "log": {"level": "debug"},
     }
+    node_config = recursive_update(values_node_config, computed_node_config)
 
     if THIS_IS_A_PUBLIC_NET:
         node_config["network"] = NETWORK_CONFIG["chain_name"]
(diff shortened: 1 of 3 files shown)


harryttd:
Whatever the structure or concept `links` will turn out to be, can we please give it a more obvious name? Something like `trusted_peers`? This is how the docs seem to refer to private node connections. https://tezos.gitlab.io/user/node-configuration.html?highlight=private%20mode

nicolasochem:
These are not trusted peers, these are links between 2 peers. What's not obvious about it?

harryttd:
Why need a standalone list of node links? With Roland's restructuring of the `nodes` definition, why not have a list of node links per node instance, per node class?
```yaml
nodes:
  tezos-baking-node:
    instances:
    - bake_using_accounts:
      - tezos-baking-node-0
      config:
        shell:
          history_mode: archive
        p2p:
          private-mode: true
      is_bootstrap_node: true
      links:
         - tezos-node-0
         - tezos-node-1
         - tezos-baking-node-1
    - bake_using_accounts:
      - tezos-baking-node-1
      config:
        shell:
          history_mode: archive
      is_bootstrap_node: true
      links:
         - tezos-node-0
         - tezos-node-1
    runs:
    - baker
    - endorser
    storage_size: 15Gi
```

This centralizes the node's links inside the node itself. You don't need an explicit 2 way mapping from the instanceA to instanceB as the links are being defined in instanceA.

Also, I don't think one should need to specify the FQDN of each node. Why can't the code handle that for the user? All you need is the name of the node itself.

Personally I still think we need a better name than `links`. Why not something like `link_to_nodes`, `link_with_nodes`, `peer_with_nodes`? Or `links_to_nodes`/`links_with_nodes`/`peers_with_nodes`? These names are much more explicit as to what this field is for. 

(The former names suggestion is in imperative/verb form. The latter names suggestion is in active/descriptive form. The reason I am making this grammatical distinction is because I just realized that we haven't been consistent with these forms in our structures. Examples being `runs` as the adjective, `bake_using_accounts` as the imperative. I think if something is a function you would use the imperative/verb form. If it is a property you would use the active/adjective/descriptive. `node.runs.baker = true` is describing a property of the instance. It isn't a function telling the node to run baker. The function in code would be `shouldRunBaker` and that would look up the prop. Sorry for this whole long craziness. Just want to express clearly what I'm thinking. So to sum this all up, I think we should use `peers_with_nodes`.)

nicolasochem:
I agree with replacing the fqdns with just the hostnames. Also replacing `links` with `peering_links` is fine.

I think defining links separately from nodes is cleaner. I briefly searched for reasons why and came across this:

https://stackoverflow.com/questions/43052290/representing-a-graph-in-json

itkach:
I think it would be really cool to define topology as  in dot format (https://graphviz.org/) instead of a custom syntax - it could then be immediately visualized and compared with actual topology reported by https://github.com/oxheadalpha/tezos-net-viz

harryttd:
@nicolasochem 

Does the link array structure imply that it is non-directed and therefore `A -> B` and and also `B -> A` ? Or is the graph directed and therefore just `A -> B`?

If the former then I understand why a separate structure makes sense, as you wouldn't want to define the edge in each node instance. You only want to define the edge once. If we really do require a separate structure and we are talking about DAGs, why not have a map of all the nodes, where each node has a list of every other node they point to? Why should each element in the list be a pair instead of a list of nodes?

The SO question you referenced is talking about a DAG. However, the DAG the user is talking about is more complicated than our use case here. His use case requires that the nodes and edges each have associated data. So the DAG representation is more complicated. We don't need that. Just a string being the hostname of a node pointing to another hostname string.

Here is a SO question about DAGs in json that is just like our use case: https://stackoverflow.com/a/9899057/6379084
The answer is like I'm suggesting here.

PR #201, opened 2021-05-13
Allow any node to be a bootstrap node.

elric1:


nicolasochem (changes requested):
I tried from `mkchain --number-of-nodes 2 --number-of-bakers 2`, it works.
I also tried  `mkchain --number-of-nodes 2 --number-of-bakers 2`, making all baker nodes non-bootstrap and all non-baker nodes bootstrap, and it worked. The activation happened on the regular nodes, and the baking nodes bootstrapped and started baking.

But `mkchain` with no arguments is broken, because it gives the following values

```
nodes:
  baking:
    tezos-baking-node-0:
      bake_using_account: baker0
      config:
        shell:
          history_mode: archive
      is_bootstrap_node: true
  regular: null
```

And it crashes `ALL_NODES = { **NODES["baking"], **NODES["regular"] }`. Until now, all the values.yaml files always had baking nodes, but you can't make the same assumption with regular nodes. Can you replace this with `NODES.get("baking", {})` (same with regular)?

elric1:
I've made the requested update.

nicolasochem (approved):
I retried everything I did prior, plus zerotier. It works. Thank you!

PR #202, opened 2021-05-14
add nomadic indexer, following the model of tzkt

nicolasochem:
hardcode to latest version for now

### DIFF
--- a/charts/tezos/templates/static.yaml
+++ b/charts/tezos/templates/static.yaml
@@ -47,3 +47,17 @@ spec:
     app: {{ .Values.tzkt_indexer_statefulset.name }}
   type: NodePort
 {{- end }}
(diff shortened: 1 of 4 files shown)


charts/tezos/templates/static.yaml
```diff
@@ -47,3 +47,17 @@ spec:
...
+{{- if (include "tezos.shouldDeployNomadicIndexer" .) }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Values.nomadic_indexer_statefulset.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  ports:
+    - port: 5000
```
harryttd, line 59: I don't think this svc is needed. No nomadic container is being exposed. Not sure if a container should be exposed.

harryttd:
Documentation should mention nomadic as another indexer option

PR #205, opened 2021-05-18
mkchain: eliminate dependency on docker by using pytezos for keys.

### LINKED ISSUE #26: remove flextesa dependency

### LINKED ISSUE #89: Research removing docker dependency in mkchain

elric1:
Research and code for #89.

closes #89
closes #26

### DIFF
--- a/mkchain/tqchain/keys.py
+++ b/mkchain/tqchain/keys.py
@@ -1,0 +1,98 @@
+import random
+import string
+import subprocess
+
+use_docker = False
+
+try:
+    from secrets import token_bytes
+
+    from pytezos import pytezos
+    from pytezos.crypto.encoding import base58_encode
+except (ImportError, NotImplementedError):
+    use_docker = True
+
+
+def set_use_docker(b):
+    global use_docker
+    if b != None:
+        use_docker = b
+
+
+loaded_images = {}
+
+
+def pull_docker_image(image):
+    if image not in loaded_images:
+        loaded_images[image] = 1
+        print(f"Checking for docker image {image}")
+        has_image_return_code = subprocess.run(
+            f"docker inspect --type=image {image} > /dev/null 2>&1", shell=True
+        ).returncode
+        if has_image_return_code != 0:
+            print(f"Pulling docker image {image}")
+            subprocess.check_output(
+                f"docker pull {image}", shell=True, stderr=subprocess.STDOUT
+            )
+            print(f"Done pulling docker image {image}")
+
+
+def run_docker(image, entrypoint, *args):
+    # We pull the image separately and before the "docker run" to
+    # simplify parsing the output.  If we didn't, then "docker run"
+    # might pull the image and emit the fact that it did so.
+    pull_docker_image(image)
+    return subprocess.check_output(
+        "docker run --entrypoint %s --rm %s %s" % (entrypoint, image, " ".join(args)),
+        stderr=subprocess.STDOUT,
+        shell=True,
+    )
+
+
+def extract_key(keys, index: int) -> bytes:
+    return keys[index].split(b":")[index].strip().decode("ascii")
+
+
+def gen_key(image):
+    if not use_docker:
+        key = pytezos.key.generate(export=False)
+        return {"public": key.public_key(), "secret": key.secret_key()}
+
+    keys = run_docker(
+        image,
+        "sh",
+        "-c",
+        "'/usr/local/bin/tezos-client "
+        + "--protocol PsDELPH1Kxsx gen keys mykey && "
+        + "/usr/local/bin/tezos-client "
+        + "--protocol PsDELPH1Kxsx show address mykey -S'",
+    ).split(b"\n")
+
+    return {"public": extract_key(keys, 1), "secret": extract_key(keys, 2)}
+
+
+def get_genesis_vanity_chain_id(seed_len=16):
+    print("Generating vanity chain id")
+
+    if not use_docker:
+        return base58_encode(token_bytes(32), b"B").decode("utf-8")
+
+    seed = "".join(
+        random.choice(string.ascii_uppercase + string.digits) for _ in range(seed_len)
+    )
+
+    return (
+        run_docker(
+            "registry.gitlab.com/tezos/flextesa:01e3f596-run",
+            "flextesa",
+            "vani",
+            '""',
+            "--seed",
+            seed,
+            "--first",
+            "--machine-readable",
+            "csv",
+        )
+        .decode("utf-8")
+        .split(",")[1]
+    )
--- a/mkchain/tqchain/mkchain.py
+++ b/mkchain/tqchain/mkchain.py
@@ -1,12 +1,11 @@
 import argparse
 import os
-import random
-import string
-import subprocess
 import sys
 from datetime import datetime, timezone
 
 import yaml
+
+from tqchain.keys import gen_key, get_genesis_vanity_chain_id, set_use_docker
 
 from ._version import get_versions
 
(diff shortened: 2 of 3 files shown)


elric1:
The force push was just to remove a couple of imports that were no longer necessary and made lint fail.

mkchain/tqchain/mkchain.py
```diff
@@ -192,10 +132,6 @@ def main():
     # pull it. This is to avoid parsing extra output. Preferably, we want to get
     # rid of docker dependency from mkchain.
     FLEXTESA = "registry.gitlab.com/tezos/flextesa:01e3f596-run"
-    images = [args.tezos_docker_image]
-    if not args.should_generate_unsafe_deterministic_data:
-        images.append(FLEXTESA)
-    pull_docker_images(images)
```
harryttd, line 198: the comments here should be removed and added in the new file created

harryttd: Also, `FLEXTESA` should be removed. The function that it is passed to below should now have the image hardcoded inside of it.

harryttd:
It may be that pytezos should go somewhere in setup.py.

Likely the readme should also be updated. You shouldn't need docker and minikube when using pytezos. Helm would still work if you are deploying the chart with gend values to another cluster.

mkchain/tqchain/keys.py
```diff
@@ -0,0 +1,95 @@
+import random
```
harryttd, line 1: Can this file be renamed to something more explicit about what its purpose is? Say `generation_utils` or maybe even just `helpers`/`utils`? `keys` doesn't give obvious context unless it was nested in something like `utils/`

mkchain/tqchain/keys.py
```diff
@@ -0,0 +1,95 @@
...
+    from pytezos import pytezos
+    from pytezos.crypto.encoding import base58_encode
+except ImportError:
+    use_pytezos = False
+
+from ._version import get_versions
+
+sys.path.insert(0, "tqchain")
+
+__version__ = get_versions()["version"]
```
harryttd, line 20: I'm not certain, is this stuff needed? In regards to `__version__`, it seems to only be used in mkchain.py.

harryttd:
Pytezos has external dependencies: https://pytezos.org/quick_start.html#requirements
The readme should mention. Maybe they can be listed in setup.py and their necessity be dependent upon the OS running?

harryttd (changes requested):

mkchain/tqchain/keys.py
elric1: Yes, you're correct, it doesn't appear to be necessary...

harryttd (approved):

nicolasochem:
I am getting an error with mkchain, but it's probably due to some missing lib on fedora 33.

```
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/crypto/key.py", line 279, in from_mnemonic                                                                                                                          
    _, secret_exponent = pysodium.crypto_sign_seed_keypair(seed=seed[:32])                                                                                                                                                                                    
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/crypto/key.py", line 38, in __getattr__                                                                                                                             
    raise NotImplementedError(                                                                                                                                                                                                                                
NotImplementedError: Please, install packages libsodium-dev, libsecp256k1-dev, and libgmp-dev, and Python libraries pysodium, secp256k1, and fastecdsa                      
```

If I understand correctly, we are not making pytezos an explicit dependency, but if it is there, we use it. Isn't it better to make it explicit with a `--no-docker` flag?

I could confirm that the docker method still works, at least.

elric1:
I added two flags: `--use-docker` and `--no-use-docker`. If you specify one, then you get the behaviour that you asked for. If you don't, then we still autodiscover what's there prefering pytezos.

harryttd:
Thanks Roland! can you please add the flags to the readme?

nicolasochem:
`--use-docker` still crashes for me because I suppose it crashes during lib import

```
[venv] nochem@peck ~/workspace/tezos-k8s () $ mkchain nico --use-docker
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/bin/mkchain", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib64/python3.9/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 855, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/nochem/workspace/tezos-k8s/mkchain/tqchain/mkchain.py", line 8, in <module>
    from tqchain.keys import gen_key, get_genesis_vanity_chain_id, set_use_docker
  File "/home/nochem/workspace/tezos-k8s/mkchain/tqchain/keys.py", line 10, in <module>
    from pytezos import pytezos
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/__init__.py", line 13, in <module>
    from pytezos.client import PyTezosClient
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/client.py", line 5, in <module>
    from pytezos.block.header import BlockHeader
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/block/header.py", line 14, in <module>
    from pytezos.sandbox.parameters import sandbox_params
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/sandbox/parameters.py", line 53, in <module>
    Key.from_faucet(sandbox_commitment).blinded_public_key_hash(),
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/crypto/key.py", line 304, in from_faucet
    key = cls.from_mnemonic(
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/crypto/key.py", line 279, in from_mnemonic
    _, secret_exponent = pysodium.crypto_sign_seed_keypair(seed=seed[:32])
  File "/home/nochem/workspace/tezos-k8s/mkchain/venv/lib/python3.9/site-packages/pytezos/crypto/key.py", line 38, in __getattr__
    raise NotImplementedError(
NotImplementedError: Please, install packages libsodium-dev, libsecp256k1-dev, and libgmp-dev, and Python libraries pysodium, secp256k1, and fastecdsa
```

I still have not spent enough time to figure why pytezos does not work on fedora 33 but I guess other people could run into this so more protection is needed.

nicolasochem (approved):
it is not trying and failing (for me) to import pytezos, so uses docker. So it does not break my workflow. Approved, thanks.

PR #213, opened 2021-06-15
Run publicnet node from config url

### LINKED ISSUE #212: can't run public testnet from URL
passing

```network:
   chain_name: https://teztnets.xyz/granadanet
```

results in wrong config file causing the node to not start.

config-init runs and populates the config file, but then config-generator ovewrites it with `"chain_name": "https://..."` which the tezos node cannot parse.

Originally, `config-init` was used to extract the bootstrap peers from the network config and put them in the p2p config. That was only needed for public chains. Private chains have their bootstrap peers configured internally.

But now, there is a new usage of `config-init` which is to use the tezos node to retrieve the url and validate it, for when the network is defined with a url.

The solution is to modify config-generator to always populate the network portion from `config-init`.

harryttd:
closes #212

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -118,9 +118,8 @@ bootstrap_peers: []
 # Identity generation difficulty.
 expected_proof_of_work: 26
 
-# Create a custom network using a config structure that is similar to a node's
-# config.json.
-# Run `tezos-node config --help` for more info.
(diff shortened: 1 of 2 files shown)


nicolasochem (approved):
I tried:

* florencenet
* granadanet with url
* private network with `mkchain`

It all works. Thanks!

PR #219, opened 2021-06-24
remote signer deployment

nicolasochem:
This creates a new values.yaml object `signers` deploying `tezos-signer`
daemons in pods in dedicated k8s deployment, and its associated service.

It is combined with a new annotation to the `baker` object called
`bake_using_signer`. It takes a value of the signer name. When set,
instead of native signing, the baker and enroser processes will connect
to the signing endpoint for every bake and endorse operation.

Example values.yaml config:

```
nodes:
  baking:
    tezos-baking-node-0:
      bake_using_account: baker0
      bake_using_signer: signer0
signers:
  signer0: {}
```

For now it does not limit the key deployment. Once we agree, in a later
PR, we can limit the private key exposure to the containers where they
are needed. But for now, every signer can see every key (just like every
baker can see every key in master).

mkchain will deploy one signer by default, and every baker will connect
to it. This is made standard as this is good practice.

Q: why signers are deployments while bakers are a stateful set?

A: tezos-signer is a very light process and a good fit for a deployment:
k8s will ensure there is always one (and just one) running at a time. If
there are several signers, we want similar guarantees for each since
(later) they will only sign for one given key.

Note: linting commit is separate since linting is apparently broken in master. Please check the first commit.

### DIFF
--- a/charts/tezos/scripts/remote-signer.sh
+++ b/charts/tezos/scripts/remote-signer.sh
@@ -1,0 +1,11 @@
+set -ex
+
+TEZ_VAR=/var/tezos
+TEZ_BIN=/usr/local/bin
+CLIENT_DIR="$TEZ_VAR/client"
+NODE_DIR="$TEZ_VAR/node"
+NODE_DATA_DIR="$TEZ_VAR/node/data"
+
+CMD="$TEZ_BIN/tezos-signer -d $CLIENT_DIR launch http signer -a 0.0.0.0 -p 6732"
+
+exec $CMD
--- a/charts/tezos/templates/signer.yaml
+++ b/charts/tezos/templates/signer.yaml
@@ -1,0 +1,65 @@
+{{- range  $key, $val := .Values.signers }}
+apiVersion: apps/v1
+kind: Deployment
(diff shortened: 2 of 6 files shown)


charts/tezos/scripts/remote-signer.sh
```diff
@@ -0,0 +1,11 @@
...
+
+TEZ_VAR=/var/tezos
+TEZ_BIN=/usr/local/bin
+CLIENT_DIR="$TEZ_VAR/client"
+NODE_DIR="$TEZ_VAR/node"
+NODE_DATA_DIR="$TEZ_VAR/node/data"
+
+CMD="$TEZ_BIN/tezos-signer -d $CLIENT_DIR launch http signer -a 0.0.0.0 -p 6732"
+
+exec $CMD
```
elric1, line 11: `exec $CMD` looks a little lonely by itself.
How about, either:
```
CMD="$TEZ_BIN/tezos-signer"

exec $CMD -d $CLIENT_DIR ...
```
or just:
```
exec $TEZ_BIN/tezos-signer ...
```

PR #223, opened 2021-07-08
Name the nodes

elric1:


charts/tezos/values.yaml
```diff
@@ -69,50 +70,46 @@ accounts: {}
...
+#      - endorser
+#    instances:
+#      - bake_using_account: baker0
+#        is_bootstrap_node: true
+#        config:
+#          shell:
+#   tezos-node:
+#     instances:
+#       - {}
+#       - {}
```
harryttd, line 112: My comment from the closed PR:
> Schema looks good.
> 
> This should still mimic the current values.yaml. So baking node should not be spun up by default. Only a regular node with rolling history mode. The default values.yaml deploys a node connecting to mainnet.
> 
> Also the comments here in the `nodes` section should still be used.

mkchain/tqchain/mkchain.py
```diff
@@ -212,36 +227,35 @@ def main():
...
-                "config": {
-                    "shell": {"history_mode": "archive" if n < 2 else "rolling"}
-                },
-            }
-            for n in range(args.number_of_bakers)
+        BAKER_NODE_NAME: {
+            "runs": ["baker", "endorser"],
+            "storage_size": "15Gi",
+            "instances": [
+                node_config(BAKER_NODE_NAME, True, n)
```
harryttd, line 235: Comment from closed PR: https://github.com/tqtezos/tezos-k8s/pull/209/files#r655653774
Magic booleans are being used

utils/config-generator.py
```diff
@@ -343,60 +367,26 @@ def import_keys(all_accounts):
...
-        pubkey_and_balance_pairs.append(
-            [account_values["key"], account_values["bootstrap_balance"]]
-        )
+    for v in accounts.values():
+        if "bootstrap_balance" in v and v["bootstrap_balance"] != "0":
+            key = v.get("pk")
+            if key == None:
+                key = v.get("pkh")
+            if not v.get("is_bootstrap_baker_account", True):
+                key = v.get("pkh")
```
harryttd, line 388: This would give an account baking rights if the `is_bootstrap_baker_account` field was not set on the dict. That should not be by default.

Why can't this be simplified to:
```suggestion
        if "bootstrap_balance" in v and v["bootstrap_balance"] != "0":
           if v.get("is_bootstrap_baker_account", False):
               key = v.get("pk")
           else:
                key = v.get("pkh")
```

elric1: Interestingly, though, it _is_ the default when we create accounts that have not been specified. And, I think that this does make sense because we only really know about baking accounts in the code---we only create accounts for bakers. And extra account that is created, is inserted into `*_values.yaml` by hand.
Now, what we _can_ do is heuristically determine if we believe an account is attached to a baker and set it that way. But until then, I do think that the True default likely makes the most sense. I can look at the heuristic, but it might be a little complicated, not sure.

elric1: The simplification that you propose is subtly different, I think. My code installs the pkh if the pk doesn't exist even in the bootstrap baker version. That said, this is dealing with a distinction that, perhaps, is not possible in practice as we can't currently specify just a public key hash. We might want to future-proof the code by tossing our toys out of the pram if we only have a pkh when we expect a pk at this stage. We can add that later, if you think that it's a good plan. Otherwise, I'll take your patch as committed in ebc3c33.

mkchain/tqchain/mkchain.py
elric1: See a2adf12

utils/config-generator.py
```diff
@@ -21,15 +20,27 @@
...
 MY_POD_NAME = os.environ["MY_POD_NAME"]
 MY_POD_TYPE = os.environ["MY_POD_TYPE"]
+
 MY_POD_CONFIG = None
+ALL_NODES = {}
+BAKING_NODES = {}
+
+for cl, val in NODES.items():
+    if val != None:
+        i = 0
```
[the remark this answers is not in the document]
elric1: See a2adf12

charts/tezos/values.yaml
elric1: see 9058020

nicolasochem (changes requested):
I did `mkchain` without any parameters and observed 2 issues:

* it deploys one baking node and one regular node. Previously it deployed one baking node only.

[This line](https://github.com/tqtezos/tezos-k8s/blob/master/mkchain/tqchain/mkchain.py#L20) cause mkchain to add `tezos-node: null` to the `nodes` section of values.yaml, but it looks like it is now named `tezos-regular-node`

* it didn't actually work. It activates properly, but subsequent baking fails. By default, mkchain deploys a signer. The secret key in the baking node is supposed to point to the signer. With mkchain with no parameters, I found that `secret_keys` was an empty list:

```~ $ cat /var/tezos/client/public_key_hashs 
[
    {
        "name": "tezos-baking-node-0",
        "value": "tz1N19m7XWZs67ChRBq34nXT6sE1b9EwE2ZX"
    }
~ $ cat /var/tezos/client/public_keys
[
    {
        "name": "tezos-baking-node-0",
        "value": {
            "locator": "unencrypted:edpkuhaZJ39XnfdfAALgZNHBDMEmiddmbh1LpAqmsB6iXLj88h4Lzh",
            "key": "edpkuhaZJ39XnfdfAALgZNHBDMEmiddmbh1LpAqmsB6iXLj88h4Lzh"
        }
    }
~ $ cat /var/tezos/client/secret_keys 
[]~ 
```

utils/config-generator.py
harryttd: It is true we only create accounts for bakers. However, it is the default only when bakers specify they have an account and it doesn't exist, and `SHOULD_GENERATE_UNSAFE_DETERMINISTIC_DATA` is true. And the defaulting is happening on the account creation. Not at the point in time where after all the keys have been accounted for, we now get all of the bootstrap account keys and balances. It shouldn't be determined at this point in the code if an account will be given baking rights or not. It should just be getting their key and balance. 

Imagine we are spinning up a new test chain and for one account we forgot to do `is_bootstrap_baker_account: false`. And the owner of the account does not want to be a baker (right away). Their account would still be given baking rights. And this is without gening unsafe data.

harryttd: I think your second comment makes sense as we can't specify a pkh yet. and the patch is good

elric1: Actually, mkchain also creates the accounts and sets `is_bootstrap_baker_account: true`. That and the filling in of defaults are the only places where our code generates accounts. The heuristic, as I mentioned, is likely superior, but we're pretty close to it as all generated accounts are marked off as true already. So, I think that we're in agreement.

elric1:
@nicolasochem: the second one turns out to be an indentation error. Not sure how it slipped through, but I do remember testing. I may have fixed the indentation---and therefore messed up the semantics---late in the process. I fixed this in `1450a76` adding a print output saying differentiating between adding a key vs a URL in the logs.
Your first point should be addressed by `ec9341b`, I just got the name of the tezos-node wrong in `charts/tezos/values.yaml` which, of course, caused an extra node to be brought up.

charts/tezos/values.yaml
```diff
@@ -69,50 +70,55 @@ accounts: {}
...
+#
+# defaults are filled in for most values.
+#
+# E.g.:
 #
-# The `config` property should mimic the structure of a node's config.json.
-# Run `tezos-node config --help` for more info.
+# nodes:
+#  tezos-baking-node:
+#    storage_size: 15Gi
```
nicolasochem, line 100: storage_size is defined here but also above in the file in `baker_statefulset` and `regular_node_statefulset`. Shall it not be in just one place?

nicolasochem (changes requested):
After most recent changes, it bakes. I also tried with 2 bakers and 2 regular nodes and it works. Zerotier is broken, I can't use mkchain with `--zerotier-network`. It's probably broken in the "invite" code path.

```
[venv] nochem@peck ~/workspace/tezos-k8s () $ mkchain nico --zerotier-network $ZT_NET --zerotier-token $ZT_TOKEN
Generating vanity chain id
Checking for docker image registry.gitlab.com/tezos/flextesa:01e3f596-run
Generating keys for account tezos-baking-node-0
Checking for docker image tezos/tezos:v9-release
Wrote chain creation constants to /home/nochem/workspace/tezos-k8s/nico_values.yaml
  File "/home/nochem/workspace/tezos-k8s/mkchain/tqchain/mkchain.py", line 297, in main
    REGULAR_NODE_TYPE: {f"{REGULAR_NODE_NAME}-0": regular_node_config},
NameError: name 'REGULAR_NODE_TYPE' is not defined
```

charts/tezos/values.yaml
elric1: Yes, I forgot to remove those two definitions. see `da8fa99`. (testing now, but it almost certainly didn't break things...)

nicolasochem:
:confused: still not ok. The invited node is complaining that it can't find a bootstrap peer.

```
│                                                                                                                                                                                                                                                            │
│ + CMD=config-generator                                                                                                                                                                                                                                     │
│ + chmod -R 777 /var/tezos                                                                                                                                                                                                                                  │
│ + set -e                                                                                                                                                                                                                                                   │
│   Importing keys for account: tezos-baking-node-0                                                                                                                                                                                                          │
│     Appending public key: edpkuiNn7Qrk19hfy1ci54t7DEWnQWo8SXBPhqDqrMw26eL7EExpVA                                                                                                                                                                           │
│   Appending public key hash: tz1an1ZS515uMyxQ8UEoYW64ud4XkbNoRELt                                                                                                                                                                                          │
│   Account key type: public                                                                                                                                                                                                                                 │
│   Account bootstrap balance: 4000000000000                                                                                                                                                                                                                 │
│   Is account a bootstrap baker: True                                                                                                                                                                                                                       │
│   Writing /var/tezos/client/secret_keys                                                                                                                                                                                                                    │
│   Writing /var/tezos/client/public_keys                                                                                                                                                                                                                    │
│   Writing /var/tezos/client/public_key_hashs                                                                                                                                                                                                               │
│ Starting config.json file generation                                                                                                                                                                                                                       │
│  -- is tezos-node-0 a bootstrap peer?                                                                                                                                                                                                                      │
│   File "/config-generator.py", line 528, in <module>                                                                                                                                                                                                       │
│   File "/config-generator.py", line 120, in main                                                                                                                                                                                                           │
│ stream closed                                                                        
```

Looking at the diff, I see changes in this area, so I guess it broke this scenario. As a reminder, when you have a zerotier invite, your bootstrap peer list comes from a file left by the zerotier init container (which gets it from the metadata on zerotier server), it's not supposed to be generated based on the node list.

nicolasochem (approved):
zerotier works now. Thank you!

harryttd (approved):

elric1:
I squashed a few of the commits into each other so that the commit history tells a more concise story. No changes resulted:
```
: arioch; git diff bb5165e 1d741c3
: arioch;
```

PR #230, opened 2021-07-20
Minor tezedge updates

elric1:


harryttd (approved):

nicolasochem (approved):
Thank you! I was able to have a tezedge pod establish peering with native client pods. I think we'll have to document how to run tezedge.

Issue #214, opened 2021-06-16
Support for multiple keys for each baker in a YAML file


hai-nguyen-van:
As discussed with @elric1, it is not currently supported to give several baker keys to a baker in a `.yaml` file. For instance, the following snippet with trigger an expected behavior (without warnings) and will not be able to use `bootstrap1` and `bootstrap2` keys for the baker process `tezos-baking-node-0`.
```
nodes:
  baking:
    tezos-baking-node-0:
      bake_using_account: bootstrap1 bootstrap2
```
Testing that a baker is able to handle several baker keys is part of our testing targets. I believe this feature request is a matter of high interest.

nicolasochem:
@hai-nguyen-van is this a new feature of tenderbake?

Currently you need to deploy one baker binary per baking address:

https://tezos.gitlab.io/introduction/howtorun.html#baker

```
Let’s launch the daemon pointing to the standard node directory and baking for user bob:

tezos-baker-alpha run with local node ~/.tezos-node bob
```

hai-nguyen-van:
I don't think this is related to Tenderbake. This was possibly already implemented even for Emmy.

nicolasochem:
@hai-nguyen-van so it's undocumented? what is the `tezos-baker` command for it?

hai-nguyen-van:
In the man page (https://tezos.gitlab.io/active/cli-commands.html), it is indeed specified that several names can be mentionned (given by dots `...`):
```
tezos-baker-009-PsFLoren run with local node context_path [name...]
```

hai-nguyen-van:
@nicolasochem Are there any progress on this side? I'm afraid we will need this very very soon. The idea is to do resilience testing on Tenderbake but with a committee of 7000. However, spawning a testnet of 7000 baking nodes is probably not possible. The only way is to have this feature so that we can share the 7000 baker keys to a smaller and reasonnable number of baking nodes (let's say 700).

PR #231, opened 2021-07-20
Allow for multiple baking accounts per baker

elric1:
As requested in Issue #214, we implement multiple baker accounts per
baker.  We update the definition of instances in the node section
to allow:

  tezos-baking-node:
    runs:
    - baker
    - endorser
    storage_size: 15Gi
    instances:
    - bake_using_accounts:
      - baker0.0
      - baker0.1
      - baker0.2
      config:
        shell:
          history_mode: full
      is_bootstrap_node: true

If "bake_using_accounts" is provided, then it is interpreted as a list
of accounts.  The will be automatically created if they don't exist when
"should_generate_unsafe_deterministic_data" is true.

"bake_using_account" is still honoured.  It is an error to specify both.

mkchain still makes parameters files using "bake_using_account" so that
we can use a newer mkchain with slightly older charts.  We expect to
deprecate this and update in the future.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -153,6 +153,21 @@ def fill_in_missing_genesis_block():
         genesis_config["block"] = gbk_b58
 
 
+def get_baking_accounts(baker_values):
+    acct = baker_values.get("bake_using_account")
+    accts = baker_values.get("bake_using_accounts")
+
+    if acct and accts:
+        raise ValueError(
+            'Mustn\'t specify both "bake_using_account" and "bake_using_accounts"'
+        )
+
+    if acct:
+        accts = [acct]
+
+    return accts
+
+
 # Secret and public keys are matches and need be processed together. Neither key
 # must be specified, as later code will fill in the details if they are not.
 #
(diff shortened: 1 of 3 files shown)


nicolasochem (approved):
Thank you. It works: I did `mkchain --number-of-bakers 2` then deleted the second node and configured the first one with `bake_using_accounts` and the baker came up, baked for both accounts and used the remote signer correctly.

I tried a few more things just for fun:

* same than above, but with 2 remote signers
* one account using a remote signer and one local

It worked.

elric1:
I just finished the `verify_this_bakers_account()` logic and did a quick force push.

nicolasochem (changes requested):
I think you broke non-baking nodes?

This is what I get in the regular nodes after `mkchain --number-of-bakers 2 --number-of-nodes 1`:

```
│ + CMD=config-generator                                                                                                                                                                                                                                     │
│ + chmod -R 777 /var/tezos                                                                                                                                                                                                                                  │
│ + set -e                                                                                                                                                                                                                                                   │
│   Importing keys for account: tezos-baking-node-0                                                                                                                                                                                                          │
│   File "/config-generator.py", line 541, in <module>                                                                                                                                                                                                       │
│   File "/config-generator.py", line 60, in main                                                                                                                                                                                                            │
│     import_keys(all_accounts)                                                                                                                                                                                                                              │
│   File "/config-generator.py", line 340, in import_keys                                                                                                                                                                                                    │
│     if expose_secret_key(account_name):                                                                                                                                                                                                                    │
│   File "/config-generator.py", line 281, in expose_secret_key                                                                                                                                                                                              │
│     return account_name in MY_POD_CONFIG.get("bake_using_accounts")                                                                                                                                                                                        │
│ TypeError: argument of type 'NoneType' is not iterable                                                                                                                                                                                                     │
│ stream closed                                                                                                                                                                                                                                              │
│                                                                                                                                                                                                                                                            │
│                                                                                                                                                                                                                                                            │
│                        
```

nicolasochem (approved):
It works again. Thanks!

I won't test zerotier, I don't anticipate any breakage there.

elric1:
I squashed them manually to avoid adding "fixup to 4dfcea9" to the commit. They're the same, though:
```
: arioch; git push -f chapeltech multiple-baker-accounts
Enumerating objects: 15, done.
Counting objects: 100% (15/15), done.
Delta compression using up to 2 threads
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 1.70 KiB | 1.71 MiB/s, done.
Total 8 (delta 7), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (7/7), completed with 7 local objects.
To github.com:chapeltech/tezos-k8s
 + 4dfcea9...919196e multiple-baker-accounts -> multiple-baker-accounts (forced
update)
: arioch; git diff  4dfcea9 919196e
: arioch;
```

utils/config-generator.py
```diff
@@ -160,47 +175,43 @@ def fill_in_missing_genesis_block():
...
+        raise Exception("ERROR: No baker accounts specified")
+
+    for acct in accts:
+        if not accounts.get(acct):
+            raise Exception(f"ERROR: No account named {acct} found.")
+
+        # We can count on accounts[acct]["type"] because import_keys will
+        # fill it in when it is missing.
+        if accounts[acct]["type"] != "secret":
+            raise Exception(f"ERROR: Either a secret key was not provided for {acct}")
```
harryttd, line 214: Is there supposed to be more to this sentence? From the "either" it would seem so.

harryttd (approved):
I left one little comment. Otherwise code looks good.

utils/config-generator.py
```diff
@@ -267,7 +278,7 @@ def expose_secret_key(account_name):
         return account_name in MY_POD_CONFIG.get("sign_for_accounts")
 
     if MY_POD_TYPE == "node":
-        return MY_POD_CONFIG.get("bake_using_account") == account_name
+        return account_name in MY_POD_CONFIG.get("bake_using_accounts", {})
```
harryttd, line 281: What if the non plural accounts was specified in the config of the node? If baker accounts are not filled in then the plural list is not added to the nodes config. So this check would return false in the case of singular account specified. 

Forgive me if this makes no sense. I'm reviewing from my phone and did not test.

elric1: The was fixed up earlier---but not in every code branch. I've fixed it, rebased onto master, squashed, and force pushed.

elric1:
I rebased in my local git repo. The commit "Provide default value for activation account" was mistakenly pushed and I've taken it out---it belongs in another pull request.

PR #234, opened 2021-07-26
Default activation account

elric1:


harryttd (approved):

nicolasochem (approved):
I tried on minikube with a default file generated by mkchain without parameters. It worked. Then I removed the activation account name from my values.yaml and it still worked.

Thanks for the pr.

PR #235, opened 2021-07-27
Add cloudwatch on multibake

elric1:
I'm refiling this one. After discussion with @nicolasochem and @harryttd, we've decided that I'll continue to use this pulumi code for the time being. That being the case, I'm going to refile the cloudwatch code.
This also includes a commit to fix the pulumi code after the recent values file schema restructure.


elric1:
Hmmm, this branch isn't quite right. Let me try again.

PR #259, opened 2021-08-09
utils/entrypoint.sh must exit with a error on missing command

### LINKED ISSUE #258: utils/entrypoint.sh should error at the end because that should never be reached.

elric1:
Fixes #258


harryttd (approved):
thank you!

PR #262, opened 2021-08-17
Add ability to run nodes with different tezos images/implementations

### LINKED ISSUE #113: Ability to run nodes with different docker images
A user should be able to chose a number of nodes running specific versions of protocols and simulate the a protocol upgrade.

### LINKED ISSUE #251: The account the baking node is referencing doesn't exist
The account the  baker and signer are referencing is `baker0`. The baking account in `accounts` is `oxheadalpha_baker_0`. Let's fix that account to be `baker0`.

### LINKED ISSUE #257: Doc update: comments in charts/tezos/values.yaml
We need to update the comments in here to reflect some additional features:

- [x] The commented out baker node’s instance has config.shell but doesn’t have any other subprops. This breaks the config.json because the node does not allow config.shell to be empty. The old values.yaml had history_mode set here to rolling.
- [x] values.yaml needs to mention all possible containers that can run for node. Currently only lists baker/endorser.
- [x] For people that are using multiple accounts for baker, should the values.yaml point out that you can do that.

### LINKED ISSUE #276: References to tezos images should now be named octez

harryttd:
closes #113 
closes #251
closes #276
closes #257


charts/tezos/templates/_containers.tpl
```diff
@@ -118,15 +118,31 @@
 {{- end }}
 {{- end }}
 
+{{- define "tezos.getNodeImplementation" }}
+{{- $node_implementation := $.node_vals.node_implementation }}
+  {{- if and (hasKey $node_implementation "octez") (hasKey $node_implementation "tezedge") }}
```
elric1, line 123: I wonder if you might want to error out if there is more than one element to the dictionary instead? Oh, hmm, I just thought of something. I think that we've missed the baker/endorser images in here! I think that we need to be able to change those. For tezedge, you still use the NL baker/endorser.

elric1 (changes requested):

charts/tezos/templates/_containers.tpl
```diff
@@ -143,13 +159,13 @@
 {{- end }}
 
 {{- define "tezos.container.tezedge" }}
-{{- if has "tezedge" $.node_vals.runs }}
+{{- if eq (include "tezos.getNodeImplementation" $) "tezedge" }}
```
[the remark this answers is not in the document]
harryttd: makes sense. so you can set the key to indicate the implementation desired but not the specific version of said implementation.

charts/tezos/templates/_containers.tpl
elric1: We've discussed online with @nicolasochem changing the schema in a way that invalidates my request. I shall, therefore, resolve this thread.

harryttd:
@elric1 ready for your review

elric1 (approved):
Looks good to me, I like the schema and I've taken a look at the code changes. I've also started a basic setup on minikube with:
```
nodes:
  eu:
    images:
      octez: tezos/tezos:v10-release
    instances:
    - bake_using_accounts: [tezos-baking-node-0]
      config:
        shell: {history_mode: archive}
      is_bootstrap_node: true
    - is_bootstrap_node: true
    - {}
    - {}
    - {}
    runs: [octez_node, baker, endorser, logger, metrics]
    storage_size: 15Gi
  us:
    instances:
    - {}
    - {}
    - {}
    runs: [octez_node, baker, endorser, logger, metrics]
    storage_size: 15Gi
  af:
    instances:
    - {}
    - {}
    runs: [tezedge_node, baker, endorser, logger, metrics]
    storage_size: 15Gi
  tezos-node: null
```
It comes up as expected and starts baking.

PR #268, opened 2021-09-11
Tezos node pods can select a cluster node + fix for snapshot import on v10

nicolasochem:
Combination of 2 small things:

* I need the ability to choose which node of my cluster my tezos node is running on. This is for an external project leveraging tezos-k8s - this is pretty standard thing for a helm chart to support
* snapshot import is broken on v10. While it's being fixed, I am adding a workaround. [issue](https://gitlab.com/tezos/tezos/-/issues/1663)


charts/tezos/values.yaml
```diff
@@ -246,3 +246,7 @@ protocols:
 #       db_password: bar
 #       # api_log_level: Debug
 #       # indexer_log_level: Debug
+#
+# To set a node-selector key-value pair, uncomment the following:
+# node_selector:
+#   key: value
```
harryttd, line 252: Why is this here? Shouldn't this be associated with the tezos nodes config? Each teznode class could then be assigned individually

harryttd: If we go this way, we wouldn't need the `nodeSelectorConfig` helper

nicolasochem:
@harryttd I changed the node selector setting to go at `nodes` level instead of globally. It works for me. Please re-review.

harryttd:
I changed the pr description a little. I hope that's ok

harryttd (approved):

PR #277, opened 2021-09-15
Restore archive node filesystem archive from s3

orcutt989:
Downloads an archive node filesystem snapshot that was created with [snap-to-s3](https://github.com/thenickdude/snap-to-s3) and expands it to `/var/tezos` during the `snapshot-downloader` init step of a new node.  To download and expand a 15GB snapshot takes about 5-10 minutes.

A filesystem snapshot of an EC2 volume in the form of an LZ4 level 1 compressed archive is done manually beforehand. Compression at this level adds virtually no time to decompression versus flat archive.   These snapshots live in the S3 bucket [tezos-snapshots](https://s3.console.aws.amazon.com/s3/buckets/tezos-snapshots?region=us-east-2&tab=objects#).

This branch has been updated to use a mainnet v10 filesystem snapshot.


nicolasochem:
Thank you! Now that POC is working, please turn this into a feature.

Let's call the new thing a "tarball".

Create a new init container called `tarball-downloader.sh`

This init container should be conditionally injected in the pod when a `tarball_url` is given as helm value.

You can accomplish this with a new helper `shouldDownloadTarball` which will look like [`shouldDownloadSnapshot`](https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/templates/_helpers.tpl#L66-L77).

Care should be taken to make sure tarball and snapshot download are mutually exclusive.

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ -d /var/tezos ] ; then
+  curl $TARBALL_URL | lz4 -d | tar -x -C /var/tezos
```
nicolasochem, line 4: What happens if there is already chain data? It looks like your script will always download and unpack the tarball when the value is given. You need to check somehow. Snapshot downloader has a similar logic where it does not download the snapshot if chain data is already here.

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,7 @@
...
+else
+  echo "/var/tezos does not exist."
```
nicolasochem, line 6: throw an error here

charts/tezos/templates/_helpers.tpl
```diff
@@ -63,6 +63,17 @@
 {{- end }}
 {{- end }}
 
+{{/*
+  Checks if filesystem archive should be downloaded.
+*/}}
+{{- define "tezos.shouldDownloadTarball" -}}
+{{- if or (.Values.tarball_url)}}
```
nicolasochem, line 70: why "if or"?

nicolasochem (changes requested):
See inline comments: please make sure it downloads the tarball at first pod start on an empty volume, not subsequent restarts. Also, please document this new capability in `charts/tezos/values.yaml` comments.

charts/tezos/templates/_helpers.tpl
orcutt989: I'm not sure. I copied the definition from shouldDownloadSnapshit. I thought it was the same syntax.

orcutt989: Ah I see now nevermind.

nicolasochem:
Looks good now. Does it work? Did you try starting a granadanet node from scratch? And when you restart the pod, it does not download again?

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,24 @@
...
+node_dir="$data_dir/node"
+node_data_dir="$node_dir/data"
+
+
+if [ -d /var/tezos ] ; then
+  if [ ! -d $node_data_dir/context ]; then
+	echo "Did not find pre-existing data, importing blockchain"
+	mkdir -p $node_data_dir
+	echo '{ "version": "0.0.4" }' > $node_dir/version.json
+	cp -v /usr/local/share/tezos/alphanet_version $node_dir
```
harryttd, line 13: this should be removable. I don't know why it's here. it just errors

orcutt989: All 4 lines or just the last 2? @nicolasochem wanted the same check in there from `snapshot-downloader.sh`

orcutt989:
> Looks good now. Does it work? Did you try starting a granadanet node from scratch? And when you restart the pod, it does not download again?

I just tried and it does not! Wahoo! 🎉

utils/tarball-downloader.sh
harryttd: Just line 13. I select all lines usually when I my comment is on multiple 🙂

nicolasochem (approved):
thank you

charts/tezos/values.yaml
```diff
@@ -30,6 +30,15 @@ chain_initiator_job:
...
 
+# You can download a public filesystem LZ4-compressed filesystem tar of an archive 
+# node by putting the URL to the archive file in tarball_url.
+
+# NOTE full_snapshot_url and rolling_snapshot_url need to be set to null.
+
+# tarball_url: 
+# full_snapshot_url: null
+# rolling_snapshot_url: null
+
```
nicolasochem, line 41: Move this new section to line 135 where snapshots are defined.

Write
`NOTE: tarballl_url and snapshot_url are mutually exclusive. When `tarball_url` is configured, `full_snapshot_url` and `rolling_snapshot_url` must be null.`

nicolasochem (changes requested):

nicolasochem (approved):
thank you!

charts/tezos/values.yaml
```diff
@@ -127,6 +127,13 @@ signers: {}
 full_snapshot_url: https://mainnet.xtz-shots.io/full
 rolling_snapshot_url: https://mainnet.xtz-shots.io/rolling
 
+# Alternatively to a snapshot, you can download a public LZ4-compressed filesystem tar
+# of a node's /var/tezos/node folder by putting the URL to the archive file in tarball_url:
+# tarball_url: 
+
+# NOTE: tarballl_url and *_snapshot_url are mutually exclusive. When tarball_url is 
+# configured,full_snapshot_url and rolling_snapshot_url must be null.
+
```
harryttd, line 136: This condition should be enforced in Helm

utils/tarball-downloader.sh
harryttd: I see this is marked as resolved. Was the code just not pushed yet?

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,24 @@
...
+
+data_dir="/var/tezos"
+node_dir="$data_dir/node"
+node_data_dir="$node_dir/data"
+
+
+if [ -d /var/tezos ] ; then
+  if [ ! -d $node_data_dir/context ]; then
+	echo "Did not find pre-existing data, importing blockchain"
+  rm -rf $node_data_dir
```
harryttd, line 11: please indent

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,24 @@
...
+	mkdir -p $node_data_dir
+	cp -v /usr/local/share/tezos/alphanet_version $node_dir
+	curl $TARBALL_URL | lz4 -d | tar -x -C /var/tezos
+  rm $node_data_dir/identity.json
```
harryttd, line 15: please indent

charts/tezos/values.yaml
harryttd: Also please put this NOTE  block right after the other "alternatively" block and before the `tarball_url`. The property should not be in between 2 paragraphs both describing things about it.

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,24 @@
...
+node_data_dir="$node_dir/data"
+
+
+if [ -d /var/tezos ] ; then
+  if [ ! -d $node_data_dir/context ]; then
+	echo "Did not find pre-existing data, importing blockchain"
+  rm -rf $node_data_dir
+	mkdir -p $node_data_dir
+	cp -v /usr/local/share/tezos/alphanet_version $node_dir
+	curl $TARBALL_URL | lz4 -d | tar -x -C /var/tezos
```
harryttd, line 14: The tarball script will continue to execute even if a command in the pipeline fails. example of where i deliberately made the code fail and you can see the script continues:
```shell
│ Did not find pre-existing data, importing blockchain                                                        │
│   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                             │
│                                  Dload  Upload   Total   Spent    Left  Speed                               │
│   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Error 44 : 100   219  100   2 │
│ Unrecognized header : file cannot be decoded                                                                │
│ tar: short read                                                                                             │
│ rm: can't remove '/var/tezos/node/data/identity.json': No such file or directory                            │
│ /var/tezos:                                                                                                 │
│ total 8                                                                                                     │
│ drwxrwxrwx    2 utmp     root          4096 Sep 20 18:12 client                                             │
│ drwxr-xr-x    3 utmp     root          4096 Sep 20 18:12 node
```

orcutt989: i have added `set -e`

utils/tarball-downloader.sh
orcutt989: yeah for some reason my editor was forcing tabs isntead of spaces ONLY for sh files...

utils/tarball-downloader.sh
orcutt989: hmm weird

charts/tezos/values.yaml
orcutt989: > This condition should be enforced in Helm

in _containers.tpl?

utils/tarball-downloader.sh
```diff
@@ -0,0 +1,26 @@
...
+
+
+if [ -d /var/tezos ] ; then
+  if [ ! -d $node_data_dir/context ]; then
+    echo "Did not find pre-existing data, importing blockchain"
+    rm -rf $node_data_dir
+    mkdir -p $node_data_dir
+    cp -v /usr/local/share/tezos/alphanet_version $node_dir
+    curl $TARBALL_URL | lz4 -d | tar -x -C /var/tezos
+    rm $node_data_dir/identity.json
```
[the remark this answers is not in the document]
harryttd: Also (based off talking via slack), if we would like some semblance of keeping track of progress, why not have `curl`, `lz4`, and `tar` all run on separate lines and have their output stored as a variable? Instead of running them as a pipeline. Then we can log things in between each command. We can also use `set -x` to log when each command is began.

charts/tezos/values.yaml
harryttd: I would probably do it in _helpers.tpl.

This is the code that checks if it should download a regular snapshot:

https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/templates/_helpers.tpl#L66-L77
```yaml
{{/*
  Checks if a snapshot should be downloaded. Either full_snapshot_url or
  rolling_snapshot_url must not be null.
  Returns a string "true" or empty string which is falsey.
*/}}
{{- define "tezos.shouldDownloadSnapshot" -}}
{{- if or (.Values.full_snapshot_url)  (.Values.rolling_snapshot_url) }}
  {{- "true" }}
{{- else }}
  {{- "" }}
{{- end }}
{{- end }}
```

I would validate that if tarball is also set after checking the snapshot urls are set, then fail. Then the same logic could be done for if the tarball should be downloaded. So checking if tarball exists, if yes and also snapshot urls, then fail.
It may be possible/preferable to do this assertion as its own standalone check instead of in both the snapshot and tarball logic.

utils/tarball-downloader.sh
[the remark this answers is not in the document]
orcutt989: I'm implementing the better pipe command the better rm. Although breaking up the commands is a good idea, its going to involve us preempting a file name for the downloaded archive, and then the removal of the temporary file. Since we can't get useful logging out of this anyways, I am going to keep the pipe for now to avoid having to manage big temporary files. Thanks for the suggestion. 👍

nicolasochem (approved):
Thank you! I fixed conflicts, and tried to set tarball and snapshot at the same time. It failed as expected.

I tried to sync a node from a granadanet snapshot,  there was an issue because you added `set -e` and there was an alphanet_version file that does not exist anymore. I removed this line so we can keep set -e.

PR #282, opened 2021-09-15
Add option to start node with local storage

nicolasochem:
Useful for faster IO for baking and RPC use cases.

Must sync from snapshot or tarball.

### DIFF
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -44,6 +44,11 @@ spec:
           name: dev-net-tun
         - emptyDir: {}
           name: config-volume
+{{- $local_storage := $val.local_storage | default false }}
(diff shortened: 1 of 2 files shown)


charts/tezos/templates/nodes.yaml
```diff
@@ -44,6 +44,11 @@ spec:
           name: dev-net-tun
         - emptyDir: {}
           name: config-volume
+{{- $local_storage := $val.local_storage | default false }}
+{{- if $local_storage }}
```
harryttd, line 48: Why need to set a variable? The default value can be set on line 48

harryttd (approved):

harryttd (approved):
thanks

elric1 (approved):
Looks good. I ran it with and without local storage in minikube.

PR #283, opened 2021-09-17
Readiness probe for tezos-node

nicolasochem:
We already had a readiness probe for tzkt, this PR adds it for the node
itself.

This is done by a python sidecar. When it receives a get request on the
/is_synced endpoint, it will query the node head block header and verify
timestamp is less than 10 minutes old.

We are using a http probe served by a flask/werkzeug server. I had to
add flask to the utils container.

When the chain is not bootstrapped, it is considered healthy so the
activation call can reach it.

This is important for a RPC server with several endpoints as they may
not all be synced. For example, when you just created the statefulset,
they sync in sequential order.

More importantly, in combination with #282 health check becomes more
important. As the chain syncs from snapshot or tarball each time it is
launched, it is important to not direct RPC traffic to nodes that are
still syncing.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -154,6 +154,10 @@
       name: config-volume
     - mountPath: /var/tezos
       name: var-volume
+  readinessProbe:
+    httpGet:
+      path: /is_synced
+      port: 31732
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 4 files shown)


elric1:
I had a concern w.r.t. readiness probes because they cause DNS names to be unpublished. I did a bit of research and have solved this issue in #294. This removes this potential objection.

charts/tezos/templates/_containers.tpl
```diff
@@ -154,6 +154,10 @@
       name: config-volume
     - mountPath: /var/tezos
       name: var-volume
+  readinessProbe:
+    httpGet:
+      path: /is_synced
```
elric1, line 159: There's an interesting conversation to have here. The default failureThreshold is three. This means that the probe must fail three times in a row for the service to be taken offline. I think that this is interesting, because if you _know_ that the head is too old then maybe you should take the pod offline immediately whereas if you fail to contact either the sidecar or the tezos-node, then you want to retry a few times.
How about periodSeconds? Is the default (10s) good for our purposes? I'd assume so, because our threshold above is 600s so an extra 30s is only 5% extra.

nicolasochem:
emergency merge since we have a production issue and the readiness probe will likely solve it

elric1 (commented):
Looks good.

PR #287, opened 2021-09-20
Fix tzkt port binding issue

harryttd:
tzkt hosts were not able to be resolved.

The issue is in regards to not being able to bind to localhost inside of a docker container. So we bind the port to `0.0.0.0`.


nicolasochem (approved):
it looks straightforwards. lgtm

PR #289, opened 2021-09-23
Few small fixes

### LINKED ISSUE #288: race in chain-initiator.sh
I think that I've seen a race in chain initiator where:
```
if ! $CLIENT rpc get /chains/main/blocks/head/header | grep '"level": 0,'; then
    echo "Chain already activated, considering activation successful and exiting"
    exit 0
fi
```
Behaves as if the RPC doesn't return a result which contains "level": 0, but the prior RPC:
```
until $CLIENT rpc get /version; do
    sleep 2
done
```
has already succeeded.

elric1:
Fixes #288 and a couple of things that I noticed.

### DIFF
--- a/bin/helm-compare-versions
+++ b/bin/helm-compare-versions
@@ -16,13 +16,15 @@ CUR_BRANCH=
 OLD_BRANCH=
 COMPARE_STASHED=0
 I_STASHED_IT=0
+DIFF_ARGS=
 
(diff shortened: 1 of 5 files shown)


nicolasochem (approved):
I tried with solo baker, 2 bakers+ 2 nodes, and zerotier. It works. Thank you!

PR #294, opened 2021-09-27
Pub dns not ready

### LINKED ISSUE #216: wait-for-bootstrap races w.r.t. bootstrap nodes.
So, I think that we still have a race when starting the nodes which is that the bootstrap nodes don't wait for each other to come up and so they might not find each other. This has been observed in the case that all of the nodes are bootstrap nodes.
A workaround is to make sure that you have at least one non-bootstrap node. I think that this does conclusively solve the issue because if one of the bootstrap nodes is delayed until after the non-bootstrap node comes up then that delayed bootstrap node will be able to at least find the bootstrap node that the non-bootstrap node found.
Also: https://gitlab.com/tezos/tezos/-/issues/1382 would fix this issue.

elric1:
We improve two things here. First, we cause coredns to publish the DNS names of our pods regardless of whether they are ready. And then, we use this to make `wait-for-bootstrap.sh` both simpler and more robust. We fix #216.


nicolasochem:
You are waiting for all bootstrap peers. What happens if we run a public network such as hangzhounet or mainnet and one of the bootstrap peers is offline forever? It looks like the init container will never terminate.

elric1:
Hmmm, well, yes. This is a hard one because the node won't retry any DNS names that it can't look up. So, if I only wait for one then I run the risk of having a node which doesn't quite work properly. Especially later with the topology patch, nodes may not be able to discover the other configured peers from their connected peers.
The thing is that this only requires that the peer has a DNS entry, it doesn't have to be up...
Maybe the right answer here is to filter the list for peers that are in kubernetes? Or maybe we just defeat the logic if we're using a public chain?

nicolasochem (changes requested):
> Or maybe we just defeat the logic if we're using a public chain?

yes, please do this.

nicolasochem:
I tried a topology of just one node (output of mkchain with no parameter) and wait_for_bootstrap fails with:


```
│ + CMD=wait-for-bootstrap                                                                                                                                                                                                                                   │
│ + exec /wait-for-bootstrap.sh                                                                                                                                                                                                                              │
│ No bootstrap peers were provided                                                                                                                                                                                                                           │
│ stream closed                                                                                                                                                                                                                                              │
│                      
```

elric1:
Oh. Yes, that is valid which means my assertion that we have peers is invalid in that case. I think that we can probably remove the check then and assume that the `config-generator` container will error out if the configuration is invalid?
Another thought is that maybe I should rename the container, because it is now not waiting for bootstrap but rather name service.

elric1:
I have removed the check for at least one bootstrap peer. I also renamed the container to wait-for-dns to be more descriptive of what it actually does. That is in a seperate commit so that we can take it out easily if we don't want to do it.

nicolasochem:
I think zerotier wants to be before config-generator but you moved it afterwards

```
│  -- is tezos-node-0 a bootstrap peer?                                                                                                                                                                                                                      │
│   File "/config-generator.py", line 572, in <module>                                                                                                                                                                                                       │
│   File "/config-generator.py", line 124, in main                                                                                                                                                                                                           │
│ stream closed     
```

that's only for the invited node. the main cluster is working fine.

Shall we just remove zerotier from the codebase?

nicolasochem (approved):
you fixed zerotier now. Thank you for making the code simpler!

Ideally `wait-for-dns` would go away altogether once octez bootstrap mechanism resolves dns periodically instead of just one at process start. Let's maintain the pressure on this bugfix.

elric1:
The force push that I just did squashes some but not all of the commits to tell a more informative story. There is no code difference:
```
: arioch; git diff ee22421 49e9cc9
: arioch;
```

PR #295, opened 2021-09-27
switch genesis proto to mainnet-genesis

nicolasochem:
It looks like since the beginning of tezos-k8s we had been using the
`PtYUEN` protocol which is carthage-bootstrap.

Nomadic is going to remove this protocol from tezos source so they asked
that we switch to tezos mainnet activation. Doing this here for our
mkchain-generated values.yaml files and the commented out example in the
default one.


harryttd (approved):
I'm assuming this works

nicolasochem:
Yes, I tried locally and I also configured mondaynet and dailynet to use this. If dailynet gets created fine today, I'll merge.

PR #296, opened 2021-09-27
remove commitments file mechanism

nicolasochem:
Since hangzhounet, we are no longer using the old file from
faucet.tzalpha.net, instead we are generating deterministic faucet
accounts from a seed. We no longer need this mechanism to download this
large commitment file from s3, so I am removing it as a feature from
tezos-k8s.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -225,12 +225,8 @@ protocols:
 #   # This data is typically too large to pass it directly inside helm chart.
 #   bootstrap_contract_urls: []
 #
-#   # Pass url pointing to a json file listing commitments (precursor to faucet accounts).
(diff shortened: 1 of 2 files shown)


harryttd:
Is there any potential use case for someone to want to be able to download a commitments file? For us creating public chains i get that we are just going to generate the addresses.

nicolasochem:
Not really. I don't think anyone else in the world is creating faucet accounts for tezos testnets. And if they do, they can use the deterministic mechanism that we created.

harryttd (approved):

elric1 (approved):
Looks good, I'll approve.
Just a comment: I think that in the future, it might be an idea to mash this together with our existing `should_generate_unsafe_deterministic_data` (if it isn't already).

PR #303, opened 2021-10-05
utils/logger.sh: log pending operations in the mempool

elric1:


harryttd (approved):

nicolasochem (approved):
It looks reasonable. Also, it's the longest jq script I've ever seen :)

elric1:
Yes, I'm going a little overboard with the jq. The eventual goal, though, I think should be to replace all of the functionality of this logger script with the metrics container.

PR #305, opened 2021-10-07
Tzkt statefulset annotations + don't re-download snapshot on pod restart

harryttd:
PR stops re-downloading a snapshot on pod restart when `keep_snapshot_after_import: false`

### DIFF
--- a/charts/tezos/templates/tzkt_indexer.yaml
+++ b/charts/tezos/templates/tzkt_indexer.yaml
@@ -59,6 +59,12 @@ data:
 
     echo "Snapshot imported"
     echo "$SNAPSHOT_URL" > "$MY_IMPORTED_SNAPSHOTS_URL"
+{{/*
+TIP: When developing, you can choose to persist the snapshot file after it's
+imported. This will save time as the pod would not have to re-download it. To
+enable this, add `keep_snapshot_after_import: true` in the tzkt section of
+values.yaml. File name format is "{db_name}-{indexer_image_tag}.backup".
+*/}}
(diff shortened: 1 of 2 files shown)


harryttd:
> is there already a mechanism in place for not re-importing snapshot at pod restart when tzkt is already synced?

yup https://github.com/oxheadalpha/tezos-k8s/blob/e84aec172f5d13de97cb5873c89f321bc8486f7d/charts/tezos/templates/tzkt_indexer.yaml#L167-L173

> PR title "don't redownload snapshot on pod restart" implies that it's currently something that's happening?

will update the description. This means that when `keep_snapshot_after_import: false` the snapshot is deleted after download. On pod restart it used to download again (without importing).

> values.yaml comment says "for development, don't delete snapshot after import": do you imply tezos-k8s development? so when you are working on tzkt spinup logic, you want to keep the file?

Yes. When developing you shouldn't have to wait a long time for snapshots to keep downloading. The snapshot should be re-used.

> annotations: what are they for? We don't have a mechanism for annotations for tezos node itself.

My intent right now is primarily to allow setting pulumi annotations. e.g. timeouts and skipAwait via values.yaml. As tzkt takes a long time to sync after being deployed, we don't the pulumi deployment to crash. Users can set whatever annotations they want though. We should have an annotations field and many others for tezos node. In general most k8s resource fields should be configurable.
An example would be the labels of a pod. This is important for our node module in development. The p2p svc in Teztnets has its selector hardcoded. Pods should be able to have custom labels set and a user should then be able to pass to the ingress function the label(s) they want it to select off of.

PR #307, opened 2021-10-16
Faucet refactor

nicolasochem:
We are adressing several challenges:

* code duplication between the faucet website itself and tezos-k8s
* teztnets deploys always build the faucet containers from submodule
* faucet website usability is low. In particular, there are no instructions.
* faucet has state: there is a redis container that keeps track of which faucet accounts have been given

A solution is to put the faucet containers in tezos-k8s repo. This way,
they will be built by the release build system.

We introduce a new chart: tezos-faucet. This chart deploys a faucet website. It can be deployed alongside a tezos chain. Both tezos and tezos-faucet share one script (`account-gen`): this script generates public/private key pairs for faucet accounts. This script runs inside the `utils` container, which is now used by both charts. The public key pair is passed to the activation job, which makes sure these accounts are added into block 1. The private keys are passed to the website, and delivered one by-one. This is recapctha-protected. We use a seed that's shared between charts, so we know that the accounts are the same.

We also add argparse to the account-gen script to provide helpful help messages, and more comments.

In teznets, we were previously using a fork of the legacy tezos-faucet repo as submodule: this is now gone. teztnets will be able to deploy without building any container once we merge & release this.

We also change the faucet app itself: it's no longer bothering with ui and rendering. It is a simple POST endpoint that 1. verifies captcha has been validated and 2. returns the next account in the list, in json format. The teztnets jekyll frontend is now in charge of faucet frontend, calls the POST endpoint, and renders the result. We also added a http redirect on GET / so if people use the old url, they will be redirected to the new one. [Example of new faucet page](https://teztnets.xyz/hangzhounet-faucet).

At boot, the account-gen init container determines which faucet is the first unused on in the list by querying balances: if it finds 10 empty accounts in a row, it assumed that it reached the end of used faucets. It's not ideal (it would be better to do an activation dry-run for every account) but it works.

I will be squasing this one.


charts/tezos-faucet/values.yaml
```diff
@@ -0,0 +1,8 @@
+tezos_k8s_images:
+  faucet: tezos-k8s-faucet:dev
+  utils: tezos-k8s-utils:dev
+recaptcha_keys:
+  siteKey: # fill in recaptcha site key
+  secretKey: # fill in recaptcha secret key
+number_of_accounts: 1000
```
elric1, line 7: This is duplicated in `charts/tezos`, but in a different location. I think that we should re-use the key names from the main chart s.t. we can share the values file. So, change `number_of_accounts` and `seed` to:
```
activation:
  deterministic_faucet_seed: askjdcnasdfasd
  deterministic_faucet_number_of_accounts: 1000

elric1: Also, I think that you should eliminate the word "deterministic" from these keys, because if you are providing a seed then it's obvious that you are seeding a PRNG (hopefully a CPRNG) to generate a set of values.

elric1 (changes requested):

charts/tezos-faucet/values.yaml
nicolasochem: Thanks. I've done that. Your suggestions are already deployed on teztnets and it works: https://github.com/oxheadalpha/teztnets/commit/7a0babc94c85e42b6cfc9e13c47ca07305cab778

elric1 (approved):

PR #313, opened 2021-11-05
Make tezos node and rpc service labels and selector configurable

### LINKED ISSUE #310: Configurable Tezos node and tezos-node-rpc service labels/selector
This is important for say the p2p svc nlb that a user may create via https://github.com/oxheadalpha/tezos-pulumi. The user needs to be able to specify a pod selector based off of which traffic will be forwarded to.

The selctor for the rpc service should also be configurable, as the user may only want to forward rpc traffic to a subset of nodes in the namespace.

harryttd:
closes #310

### DIFF
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -1,42 +1,43 @@
 {{- range $key, $val := .Values.nodes }}
 {{- if $val }}
-{{- $_ := set $ "node_class" $key }}
-{{- $_ := set $ "node_vals" $val }}
+  {{- $_ := set $ "node_class" $key }}
+  {{- $_ := set $ "node_vals" $val }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
-  name: {{ $key }}
+  name: {{ $.node_class }}
   namespace: {{ $.Release.Namespace }}
 spec:
   podManagementPolicy: Parallel
-  replicas: {{ $val.instances | len }}
-  serviceName: {{ $key }}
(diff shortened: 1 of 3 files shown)


orcutt989:
Please rebase. I just upgraded hangzhounet to RC2.

orcutt989:
Awesome thank you! 🙏

charts/tezos/templates/nodes.yaml
```diff
@@ -1,42 +1,43 @@
...
     spec:
       containers:
-{{- include "tezos.container.node"      $ | indent 8 }}
-{{- include "tezos.container.tezedge"   $ | indent 8 }}
-{{- include "tezos.container.bakers"    $ | indent 8 }}
-{{- include "tezos.container.endorsers" $ | indent 8 }}
-{{- include "tezos.container.logger"    $ | indent 8 }}
-{{- include "tezos.container.metrics"   $ | indent 8 }}
-{{- include "tezos.container.zerotier"  $ | indent 8 }}
-{{- include "tezos.container.sidecar"   $ | indent 8 }}
```
nicolasochem, line 31: why change indentation here?

harryttd: Easier to read the yaml. Cleaner instead sublocks still being under the same level as their parent.

nicolasochem (approved):
LGTM, didn't try. I assume you tried the indentation change and it works.

harryttd:
Thanks. yes i've been working on tezos-pulumi using this code in my local chart

PR #320, opened 2021-11-16
Add a way to define env vars in values.yaml

elric1:


elric1:
I think that I may use the PR to group a few small but unrelated changes on which I am working.

charts/tezos/templates/nodes.yaml
```diff
@@ -39,10 +39,10 @@ spec:
         {{- include "tezos.init_container.config_init"         $ | indent 8 }}
         {{- include "tezos.init_container.zerotier"            $ | indent 8 }}
         {{- include "tezos.init_container.config_generator"    $ | indent 8 }}
-        {{- include "tezos.init_container.wait_for_dns"        $ | indent 8 }}
         {{- include "tezos.init_container.snapshot_downloader" $ | indent 8 }}
         {{- include "tezos.init_container.snapshot_importer"   $ | indent 8 }}
         {{- include "tezos.init_container.tarball_downloader"  $ | indent 8 }}
+        {{- include "tezos.init_container.wait_for_dns"        $ | indent 8 }}
```
nicolasochem, line 45: moved by accident?

nicolasochem (commented):
great, I assume the primary need is for debug flags like ` TEZOS_LOG="* -> debug"` ?

charts/tezos/templates/nodes.yaml
harryttd: There is a weird space encoding issue. I don't see it moved in github but i do in the email github sent me.

harryttd:
> I think that I may use the PR to group a few small but unrelated changes on which I am working.

```yaml
## `global_node_config` defines global values which will be
## set for every node type in the `nodes` section below.
global_node_config:
    # Environment variables for every nodes' containers
    env: {}
```
?

charts/tezos/templates/nodes.yaml
elric1: Debugging flags and also for things like `LD_PRELOAD` or `TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=Y`.

elric1: The move isn't an accident, I'm using this PR for a number of small changes. It's in a separate commit which tells you why I did it.

charts/tezos/templates/_containers.tpl
```diff
@@ -11,6 +11,10 @@
   value: node
 - name: MY_NODE_CLASS
   value: {{ .node_class }}
+{{- range $key, $val := .Values.node_globals.env }}
```
harryttd, line 14: Not every container has `localvars.pod_envvars` applied. So global vars aren't technically applied to all containers. And there are containers that do have it applied but don't want additional env vars that they don't need like `TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=Y`

Perhaps there should be a separate templating function of the global vars aside from `localvars.pod_envvars`? Maybe the global node env vars should only be applied to the necessary containers instead of to all containers in a pod who don't need the var pollution.

elric1: I fixed that, all the pods now have it. But, I think that people might very well want to apply variables to all of the containers. I'm not sure that env var pollution is too terrible, is it?

elric1: Or to put it differently, only tezedge didn't include the pod_envvars. So, I fixed that in this commit. If we are to streamline the env vars, maybe that belongs in an issue which we can address separately?

nicolasochem: From what I read on the I network test channel, debug flag on the baker is useful, but debug flag on the node is very vebose (8 Mb logs a minute!) so it may be useful to have env vars per container, indeed. But that can come later.

charts/tezos/templates/nodes.yaml
nicolasochem: OK. Every time we touch this, zerotier breaks :P I'll give it a try.

charts/tezos/templates/_containers.tpl
elric1: I'm not convinced that we need to avoid env var pollution, especially env vars that the user is specifically setting themselves. However, I am convinced that we need to be able to set the same env var to different values in different containers. I'll look into it.
Keep in mind the next commit will be:
```
node_globals:
  config:
     key: val
```
to pass through global config settings. And that will not be per container type.

elric1:
Okay, I've split out all of the ancilliary work that I was putting in the PR into separate PRs. In order to write the functionality that has been requested vis a vis allowing environment variables to be set on a container by container basis---which I agree is necessary and will help us with many use cases, I thought that it would be a good idea to refactor the code even more than I did last time. I did this to minimise repetition of what will be some fun templating code for every single container that we mainain.
I'm not finished, yet, but please have a look at the existing refactoring and provide feedback.

orcutt989:
> Also, I don't like the names that I used for the config section "node_global" is a bit odd. Suggestions?

What if we just called them `node_var` or `node_envs`. Dont we know they are global just from using them?

elric1:
I'm closing this to create a new PR.

PR #322, opened 2021-11-18
Ability to specify both archive and rolling tar urls

harryttd:
PR merges the snapshot and tarball downloader scripts into a single snapshot-downloader.sh script. ("snapshot" referring to both a tezos exported snapshot and also a tar being an fs snapshot)

We allow the specification of archive and rolling tar urls as opposed to a single tar url. Helm logic checks that only either rolling tar or snap is specified. Both are not allowed.


elric1:
Having a look, I think that it might be better to merge snapshot-{downloader,importer}. The tezos image includes wget rather than curl, but they should be functionally equivalent. Question is: can we speed things up by parallelising the download and import as we do for tarballs? If that is possible, then we should definitely do it as it would speed things up.

harryttd:
Good point that we can use wget. However I didn't really want to write regex to parse the node's history mode because Octez doesn't have jq. This PR was the easiest way for me to get the desired results as we need it for Oxheadinfra. The download + import flow is already not ideal. This PR makes it a little more ideal by having the different file types download flow all in one script.

Regarding our discussion via Slack, that is a great point to ask NL to implement snapshot import via a URL. I can make an issue on Gitlab.

elric1:
I'm not sure how difficult it would be for them to implement, i.e. I'm not sure if the `fseek(2)` during the snapshot import. But, my guess is that it is a purely linear process of adding blocks and validating the signatures, in which it should be relatively straight--foward.

charts/tezos/templates/configs.yaml
```diff
@@ -13,7 +13,8 @@ data:
     }
   FULL_SNAPSHOT_URL: "{{ .Values.full_snapshot_url }}"
   ROLLING_SNAPSHOT_URL: "{{ .Values.rolling_snapshot_url }}"
-  TARBALL_URL: "{{ .Values.tarball_url }}"
+  ARCHIVE_TARBALL_URL: "{{ .Values.archive_tarball_url }}"
```
elric1, line 16: I wonder if a slightly better schema might be interesting at some point, something like:
```
snapshot:
  type: tarball
  full_url: ...
  rolling_url: ...
```

utils/snapshot-downloader.sh
```diff
@@ -7,26 +7,62 @@ node_dir="$data_dir/node"
...
+snapshot_url=""
+tarball_url=""
 case "$my_nodes_history_mode" in
-        full)           snapshot_url="$FULL_SNAPSHOT_URL"       ;;
-        rolling)        snapshot_url="$ROLLING_SNAPSHOT_URL"    ;;
-        *)              echo "No snapshot URL provide for node, exiting"
-                        exit 1;;
+  full)     snapshot_url="$FULL_SNAPSHOT_URL";;
+
+  rolling)  snapshot_url="$ROLLING_SNAPSHOT_URL"
```
elric1, line 34: I'd like to see this case as:
```
   rolling*) ...
```
as you can specify the number of cycles to store by specifying something like `rolling:2`.

harryttd: our conclusion over slack is to make a new issue

charts/tezos/templates/configs.yaml
elric1: I created issue #325 , we can discuss it later.

charts/tezos/templates/nodes.yaml
```diff
@@ -19,12 +19,12 @@ spec:
       labels:
         appType: tezos-node
         node_class: {{ $.node_class }}
-        {{- if has "baker" $.node_vals.runs }}
+      {{- if has "baker" $.node_vals.runs }}
```
elric1, line 22: Is this part of this commit or did it sneak in somehow?

elric1 (changes requested):

charts/tezos/templates/nodes.yaml
harryttd: I nested `baking_node` and labels under the `if`s so you can see the conditional structure more easily instead of it all being stacked on top of eachother

elric1: Ah, sorry, I didn't notice that it was whitespace only.

elric1 (approved):

elric1:
I opened [#2138: tezos-node snapshot import should be able to take a remote URL](https://gitlab.com/tezos/tezos/-/issues/2138).

orcutt989 (approved):
Awesome thanks!

PR #327, opened 2021-11-22
Allow the config pass-through to be used at all three levels.

elric1:
This commit allows for config to be set globally and per statefulset rather than just at the instance level.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -15,13 +15,15 @@ from base58 import b58encode_check
 
 ACCOUNTS = json.loads(os.environ["ACCOUNTS"])
 CHAIN_PARAMS = json.loads(os.environ["CHAIN_PARAMS"])
+NODE_GLOBALS = json.loads(os.environ["NODE_GLOBALS"]) or {}
 NODES = json.loads(os.environ["NODES"])
 SIGNERS = json.loads(os.environ["SIGNERS"])
 
 MY_POD_NAME = os.environ["MY_POD_NAME"]
 MY_POD_TYPE = os.environ["MY_POD_TYPE"]
 
-MY_POD_CONFIG = None
+MY_POD_CLASS = {}
+MY_POD_CONFIG = {}
 ALL_NODES = {}
 BAKING_NODES = {}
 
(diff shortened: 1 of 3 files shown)


utils/config-generator.py
```diff
@@ -525,7 +526,10 @@ def create_node_config_json(
         },
         # "log": {"level": "debug"},
     }
-    node_config = recursive_update(values_node_config, computed_node_config)
+    node_config = NODE_GLOBALS.get("config", {})
+    node_config = recursive_update(node_config, MY_CLASS.get("config", {}))
+    node_config = recursive_update(node_config, MY_POD_CONFIG.get("config", {}))
+    node_config = recursive_update(node_config, computed_node_config)
```
harryttd, line 532: my comment copy paste from origin PR:
```
Thanks for adding this feature 😄 Nicolas and I were recently talking about having a config on the class level.

Can you please document this functionality in values.yaml?
```

utils/config-generator.py
```diff
@@ -31,6 +32,7 @@
             name = f"{cl}-{i}"
             ALL_NODES[name] = inst
             if name == MY_POD_NAME:
+                MY_CLASS = val
```
harryttd, line 35: Can you please initialize this variable on line 24 where we initialize other global vars? Also if you agree, to please name it MY_NODE_CLASS so the meaning of class is clear that it is referring to the particular node's statefulset name/class.

Actually, MY_NODE_CLASS is an env var on the pod, so it can just be set on line 21 with the other pod env vars.

utils/config-generator.py
elric1: Oh, this got lost somewhere when I split the one PR into separate PRs for each feature. Let me add this back in.

utils/config-generator.py
elric1: I used `MY_POD_CLASS` to be consistent with `MY_POD_CONFIG` and initialised it.

nicolasochem (approved):
I tried mkchain with 2 bakers and 1 node. I edited the config to the following:

```
node_globals:
  config:
    shell:
      history_mode:
        rolling:
          additional_cycles: 7
nodes:
  tezos-baking-node:
    instances:
    - bake_using_accounts:
      - tezos-baking-node-0
      is_bootstrap_node: true
      config:
        shell:
          history_mode:
            rolling:
              additional_cycles: 4
    - bake_using_accounts:
      - tezos-baking-node-1
      is_bootstrap_node: true
    runs:
    - octez_node
    - baker
    storage_size: 15Gi
  tezos-node:
    config:
      shell:
        history_mode:
          rolling:
            additional_cycles: 2
    instances:
    - is_bootstrap_node: false
    storage_size: 15Gi

```

The configuration was applied as expected. Thank you for this.

PR #330, opened 2021-11-22
Fixed up some things when the activation account is missing.

elric1:
We deal with two cases: (1) we're generating accounts, in which case we just add it to the list of accounts to generate, and (2) we aren't in which case we make the error message actually tell you what is wrong.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -54,7 +54,7 @@ def main():
 
     if SHOULD_GENERATE_UNSAFE_DETERMINISTIC_DATA:
         fill_in_missing_genesis_block()
-        all_accounts = fill_in_missing_baker_accounts()
+        all_accounts = fill_in_missing_accounts()
         fill_in_missing_keys(all_accounts)
 
     fill_in_activation_account(all_accounts)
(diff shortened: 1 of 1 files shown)


nicolasochem (approved):
I did `mkchain --number-of-bakers 2` then changed the activation account name to an account that didn't exist.

Saw this error:

```
│     raise Exception("ERROR: Couldn't find the genesis_pubkey. " +                                                                                                                                                │
│ Exception: ERROR: Couldn't find the genesis_pubkey. This generally happens if you forgot to define an account for the activation account 
```

Then I added the setting `should_generate_unsafe_deterministic_data: True` to my values.yaml and I saw the following in the init container of the activation account:

```
│ Creating activation account: tezos-baking-node-999                                                                                                                                                               │
│ Fill in missing keys                                                                                                                                                                                             │
│   Deriving secret key for account tezos-baking-node-999 from genesis_block                                                                                                                                       │
│ Importing keys                                                                                                                                                                                                   │
│   Importing keys for account: tezos-baking-node-999                                                                                                                                                              │
│     Appending secret key                                                                                                                                                                                         │
│     Appending public key: edpkttvGKGy8GrUzG7hSmtptPBkYjpaoSHxC9KZLsgaBFCMVm4ZBfB                                                                                                                                 │
│   Appending public key hash: tz1d8cnv63ZxWr454SZiNeGnuukwUEknGbRJ                                                                                                                                                │
│   Account key type: secret                                                                                                                                                                                       │
│   Account bootstrap balance: 4000000000000                                                                  
```

then the chain activated and bakers started baking. I believe that's the intent. Thank you.

PR #332, opened 2021-11-23
add an optional fitness parameter to activation

nicolasochem:


nicolasochem:
Closing: reactivating with a higher fitness is currently broken

PR #337, opened 2021-11-29
private chains should not download archive tarballs

nicolasochem:
I tried mkchain with no parameters, and realized that the chain was
trying to download a mainnet snapshot. This is not correct behvior,
fixing.

### DIFF
--- a/mkchain/tqchain/mkchain.py
+++ b/mkchain/tqchain/mkchain.py
@@ -161,6 +161,7 @@ def main():
         "full_snapshot_url": None,
         "rolling_snapshot_url": None,
         "archive_tarball_url": None,
+        "rolling_tarball_url": None,
     }
 
     # preserve pre-existing values, if any (in case of scale-up)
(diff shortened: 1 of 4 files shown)


mkchain/tqchain/mkchain.py
```diff
@@ -157,9 +157,10 @@ def main():
             "zerotier_network": args.zerotier_network,
             "zerotier_token": args.zerotier_token,
         },
-        # Custom chains should not pull snapshots
+        # Custom chains should not pull snapshots or tarballs
         "full_snapshot_url": None,
         "rolling_snapshot_url": None,
+        "archive_tarball_url": None,
```
harryttd, line 163: thanks for catching. THis should also include `rolling_tarball_url`

nicolasochem:
but it's none by default. shall I still add it?

harryttd:
I think it's a good idea to. It definitely doesn't hurt.

harryttd (approved):
Thanks for handling this.

PR #340, opened 2021-12-03
We musn't download a snapshot if we are activating a private chain

elric1:


nicolasochem:
IIUC you are checking whether the `activation` node is present in config, then assuming it's a private chain and not downloading any snapshot even if it's passed as config.

See a typical values.yaml for teztnets: https://github.com/oxheadalpha/teztnets/blob/main/hangzhounet/values.yaml

The trouble there is that I am leaving the `activation` parameter in place (because activate is idempotent: if the chain is already activated, the job does nothing). But I may want to download a snapshot (or more likely a tarball since teztnets platform runs archive nodes). For protocol testnets and long-running testnets, we will eventually build archive tarballs (and actually it will happen pretty soon).

We expect users to use mkchain when they want a private chain. mkchain sets snapshots/tarballs to null. I recently fixed a bug related to this: https://github.com/oxheadalpha/tezos-k8s/commit/bdc709664363b042c6f222c76d4e695f3d4d979c

I think your PR introduces under-the-hood magic. I like it better the way it is.

One idea that could reconcile us would be to have "per chain snapshots" in values: `rolling_snapshot_url` could be a string, in which case it's always honored. Or it could be a map of strings with network such as `mainnet`, `hangzhounet`, `https://teztnets.xyz/kaizen` as keys. The default values.yaml would have such a map pointing to the various snapshots and tarballs that we make available. But when you run a private chain, you don't match any of these so you don't download a snapshot or tarball by default.

harryttd:
I'm thinking out loud....

I understand that the activate is idempotent but we still do check if the job itself should be created in the first place based on if the activation values are specified. You can keep the activate values in the values.yaml but just comment them out on subsequent deploys. However, expecting a user to comment them out is probably not very realistic.

I think a general pattern that people should use is to have base values.yaml files for different scenarios which fill in and override any of the default chart values.yaml. This way you don't have to remember to set tar/snap urls to null, for example.

In regards to the map of nets <-> urls, i don't think we should be maintaining this and hardcoding the different test networks as they come and go.

elric1:
Hmmm, and we have to consider what to do the first time. If you have a snapshot configured and activate set, then the first time you start there will be an error downloading the snapshot. So, you really want an activate-or-snapshot kind of functionality.

nicolasochem:
> I think a general pattern that people should use is to have base values.yaml files for different scenarios which fill in and override any of the default chart values.yaml. This way you don't have to remember to set tar/snap urls to null, for example.
> 
> In regards to the map of nets <-> urls, i don't think we should be maintaining this and hardcoding the different test networks as they come and go.

Good points, yes. I still like that starting the helm chart with no parameters whatsoever will give you a mainnet node syncing from a tarball. I don't think there is a way to have everything we want.

elric1:
Also, maybe snapshots should be gathered into a map a la:
```
snapshots:
  rolling:
    type: tarball
    url: http://the.location.or/the/snapshot
  full:
    type: snapshot
    url: http://...
  archive:
    type: tarball
    url: ...
```
Now, there _is_ still an issue because rolling snapshots include a number of additional cycles and you can't mismatch those without providing an extra flag. I'm not sure if we want to worry about that or not...

elric1:
I think that if we make sync'ing from a tarball the default, then we should really sign the tarball and make tezos-k8s check the sigs. Otherwise, I'd recommend a snapshot as the default even if it is slower.

elric1:
I've made a new proposal in #394.

PR #351, opened 2022-01-05
Pyrometer

nicolasochem:
This is a chart I'm using to deploy pyrometer in teztnets: https://pyrometer.teztnets.xyz

So, it's known working.

### DIFF
--- a/charts/pyrometer/templates/configmap.yaml
+++ b/charts/pyrometer/templates/configmap.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+data:
+  config.json: |
+{{ .Values.config | mustToPrettyJson | indent 4 }}
+kind: ConfigMap
+metadata:
+  name: pyrometer-config
+  namespace: {{ .Release.Namespace }}
--- a/charts/pyrometer/templates/deployment.yaml
+++ b/charts/pyrometer/templates/deployment.yaml
@@ -1,0 +1,27 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pyrometer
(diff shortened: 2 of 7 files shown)


charts/pyrometer/templates/deployment.yaml
```diff
@@ -0,0 +1,36 @@
...
+    matchLabels:
+      app: pyrometer
+  template:
+    metadata:
+      labels:
+        app: pyrometer
+      annotations:
+        # ensure that the pod bounces each time configmap changes
+        # https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
harryttd, line 17: very cool

charts/pyrometer/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+data:
+  config.json: |
+{{ .Values.config | mustToPrettyJson | indent 4 }}
```
harryttd, line 4: can we default to `{}` so if `config` value is empty/null it won't break pyrometer. A null value breaks it

harryttd (changes requested):

charts/pyrometer/templates/configmap.yaml
nicolasochem: why would you launch pyrometer without config thou?

harryttd: ah right. good point. There are no defaults it would fall back on.

harryttd (approved):

PR #353, opened 2022-01-07
Rename tezos nodes in various places

### LINKED ISSUE #352: Change "tezos-node" to name that uses the node's history mode
The current naming scheme of `tezos-node` doesn't make it obvious what the node is doing and being used for. We have more explicit naming in oxheadinfra, like `rolling-node`. We have to set `tezos-node: null` to override the default chart node values so that it isn't created. And then create the nodes we want, e.g. `archive-node`, `rolling-node`, etc.

 With this change, we won't have to worry about a `tezos-node` statefulset being created. However, if one doesn't want a rolling node, they would still need to set `rolling-node: null`.

harryttd:
closes #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,44 +326,45 @@ E.g.:
 
 ```
 nodes:
-  full-baking-node:
+  baking-node:
     storage_size: 15Gi
     runs:
-    - baker
-    - endorser
-    - logger
+      - baker
+      - endorser
+      - logger
     instances:
-    - bake_using_account: baker0
-      is_bootstrap_node: true
-      config:
-        shell:
+      - bake_using_account: baker0
+        is_bootstrap_node: true
+        config:
+          shell:
+            history_mode: rolling
   full-node:
     instances:
-    - {}
-    - {}
-  full-tezedge-node:
+      - {}
+      - {}
+  tezedge-full-node:
     runs:
-    - baker
-    - endorser
-    - logger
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -138,24 +138,16 @@ Connecting to a public net is easy!
...
 
-If you'd like to spin up a node with history mode full, run:
+Running this results in:
 
-```shell
-helm install tezos-mainnet oxheadalpha/tezos-chain \
---namespace oxheadalpha --create-namespace \
---set nodes.regular.tezos-node-0.config.shell.history_mode=full
-```
-
```
harryttd, line 155: I got rid of creating a full node and just left the user with creating the default rolling node. We would need to introduce needing a values.yaml here and referencing it, and I think it is not necessary at this point. We can't just `--set tezos-node....` anymore, bec the node names are not uniform. They reflect their history modes.

nicolasochem (changes requested):
Looks good except some of your readme changes. I don't like `full-baking-node`, it sounds awkward. Plus the recommendation is to use rolling nodes for baking. Can you just name it `baking-node` and set it as rolling in the config example?

Also `full-tezedge-node` should be named `tezedge-full-node` I think.

nicolasochem (approved):
thanks!

harryttd:
Thank you Sir

PR #354, opened 2022-01-11
Document config pass-through at statefulset level.

### LINKED ISSUE #345: values.yaml docs is missing comment regarding config on the node class level
Currently only the config on node_globals and the instance level are referred to. The node class config level needs to be commented about as well.

I would add this comment in https://github.com/oxheadalpha/tezos-k8s/blob/953037199b6674fd78cf002397a747231566f2c6/charts/tezos/values.yaml#L74

elric1:
Fixes #345

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -74,6 +74,8 @@ accounts: {}
 ## Params at the statefulset level:
 ## - "config": The "config" property should mimic the structure of a node's
 ##             config.json. Run `tezos-node config --help` for more info.
+##             If present at the statefulset level, it overrides it in
+##             node_globals.
 ## - "storage_size": the size of the PV
 ## - "images": Optional specification of images to use for the tezos node and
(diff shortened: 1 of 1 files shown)


charts/tezos/values.yaml
```diff
@@ -88,8 +90,7 @@ accounts: {}
 ##   - "bake_using_account": Account name that should be used for baking.
 ##   - "bake_using_accounts": List of account names that should be used for baking.
 ##   - "is_bootstrap_node": Is this node a bootstrap peer.
-##   - "config": The "config" property should mimic the structure of a node's
-##             config.json. Run `tezos-node config --help` for more info.
+##   - "config": same as the outer lavel, pass through to config.json.
```
harryttd, line 93: "lavel" typo. I think it might also be a good idea to mention here that it will override the top level config.

harryttd (changes requested):

harryttd (approved):

PR #355, opened 2022-01-11
snapshot-downloader.sh now checks for history_mode at all levels.

### LINKED ISSUE #341: Snapshot downloader: Check for node's history mode only looks at the config on the node instance level
The snapshot-downloader script checks the history mode of the node to determine what snapshot/tarball it should download. The script only checks the instance level config. If no history mode is defined there, but say the mode is defined on the class level or on the global level, the script fails. `Invalid node history mode: 'null'`

elric1:
Fixes #341

### DIFF
--- a/utils/snapshot-downloader.sh
+++ b/utils/snapshot-downloader.sh
@@ -19,10 +19,8 @@ fi
 
 echo "Did not find a pre-existing blockchain."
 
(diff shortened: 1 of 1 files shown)


harryttd (approved):
I tested using class level config and node_global config

nicolasochem (approved):
looks good, and it simplifies the code. Note: I didn't try it myself.

PR #360, opened 2022-01-31
add the accuser

elric1:


harryttd:
Please add accuser to the values.yaml comments: https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/values.yaml#L85-L86

harryttd (changes requested):
Please add accuser to the values.yaml comments:
https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/values.yaml#L85-L86

charts/tezos/templates/_containers.tpl
```diff
@@ -250,6 +250,22 @@ Also start endorser for protocols that need it.
...
 
+{{- define "tezos.container.accusers" }}
+{{- if has "accuser" $.node_vals.runs }}
+{{ $node_vals_images := $.node_vals.images | default dict }}
+{{- range .Values.protocols }}
+- name: accuser-{{ lower .command }}
+  image: "{{ or $node_vals_images.octez $.Values.images.octez }}"
+  imagePullPolicy: IfNotPresent
+  command:
+    - /usr/local/bin/tezos-accuser-{{ .command }}
```
harryttd, line 261: Do we need to provide a path to the client directory? 

Maybe the baker-endorser.sh script can be utilized if all of its internal code can and should be used for accuser.

elric1: Hmmm, ISTR that I have some code around somewhere that refactors all of this. I wonder what I did with it, but it would have done this.
But to the point at hand: we could put those directories in place and it would be nice for symmetry (and I will during the refactor code which is lying around somewhere), however, interestingly, the accuser doesn't need any of it. It's config is separate, it only speaks HTTP to localhost and it doesn't require access to any accounts. In essence, the accuser just generates an anymous accusation and the proceeds of any penalties accrues to the first baker to include said accusation in the block that they are baking.

nicolasochem (approved):
I tried it on my minikube and it works. Thank you

Issue #319, opened 2021-11-16
Pin the python base version in utils container


harryttd:
As part of our pulumi infra, we often build the utils image. We have attempted a docker caching solution so that the image does not need to be built from scratch every time. It doesn't always work.

After some investigation: if the code of the underlying base layer of an image is updated, and the base you specify in your image is tagged so that it would try and pull the updated base version, your entire image would need to be built from scratch. So even though it looks like your base is fully pinned, i.e. `python:3.9-alpine`, it isn't really, and python code updates could all be tagged with the same tag, and hence cause image builds from scratch.

related to #317

Issue #317, opened 2021-11-15
Utils docker image build is really slow


harryttd:
The amount of time it takes is especially a problem with deploying code to our infra. We have tried to utilize a caching mechanism in pulumi. It doesn't completely work as of yet. But I wanted to put some articles in here for future reference that have some ideas about how to make our python base image build faster.

We should investigate:
- different python base images
- parallel pip package installs. We currently have many lines all running separate pip install cmds.

https://pythonspeed.com/articles/alpine-docker-python/
https://stackoverflow.com/questions/26228136/pip-build-option-to-use-multicore

elric1:
I think that PR#362 and its fix PR#366 address this issue?

harryttd:
Closing issue. Removing numpy significantly reduced the build time.

PR #361, opened 2022-02-08
utils/Dockerfile: make image smaller

elric1:


nicolasochem (approved):
This PR adds `--no-cache` to every pip command. Previously we were doing it inconsistently, resulting in a large image.

I tried mkchain with default parameters on this build and it worked. I also tried activation with faucet and it worked.

#317 and #319 can be addressed in another PR. Let's merge. ty

PR #362, opened 2022-02-08
utils/: remove numpy as a dependency

elric1:
This takes more than 5 minutes off the build.


nicolasochem:
I can't activate with a faucet because of [this line](https://github.com/oxheadalpha/tezos-k8s/blob/4c75d8c03fd92580e49616a4ff17f1f7b214b084/utils/faucet-gen.py#L107)

```
│                                                                                                                                                                                   │
│ + CMD=faucet-gen                                                                                                                                                                  │
│ + shift                                                                                                                                                                           │
│ + exec /faucet-gen.py --seed 'deterministic faucet seed' --number-of-accounts 100 --write-commitments-to /faucet-commitments/commitments.json                                     │
│ Seed is deterministic faucet seed                                                                                                                                                 │
│ numpy seed is 846951416                                                                                                                                                           │
│ Traceback (most recent call last):                                                                                                                                                │
│   File "/faucet-gen.py", line 107, in <module>                                                                                                                                    │
│     np.random.seed(seed=numpy_seed)                                                                                                                                               │
│ NameError: name 'np' is not defined                                                                                                                                               │
│                                         
```

elric1:
I've fixed your issue and also added another commit which removes the bitcoin lib dependency replacing it with pytezos.

nicolasochem (approved):
This PR improves on the faucet python code base in 2 ways:

* removes the numpy dependency by using the paretovariate function from python `random` library
* remove the bitcoin git dependency by:
  * using the mnemonic library from trezor which seems to be well maintained
  * using base64 encode and decode functions from pytezos. A generic base64 method may have worked as well, but we already use pytezos in the config-gen script, so...

I launched a private chain with a faucet on minikube and it activated fine. The activation job also had a faucet init container and the logs said `Faucet commitment file found, adding faucet commitments to protocol parameters` so I assume it is working.

Roland claims that the files were identical after introducing the mnemonic package and pytezos. I didn't check.

Numpy removal results in different balances for faucet accounts. But it matters only to the activate job. When upgrading faucet for a chain that has already been activated (hangzhou, ithacanet) the faucet will hand out json with `amount` being off. I think that won't matter. Let's see. If it matters, there is always the option to run an old helm chart for these faucets.

Thank you for making this code more maintainable and the container smaller.

PR #364, opened 2022-02-10
Snapshot Engine

orcutt989:
Adds Snapshot Engine Helm chart and Docker image.


snapshot_maker/volumeFromSnap.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: problem-volume
+  namespace: mainnet-shots-2
```
nicolasochem, line 5: This should not be hardcoded.

snapshot_maker/mainJob.yaml
```diff
@@ -0,0 +1,165 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "zip-and-upload"
+  namespace: mainnet-shots
```
nicolasochem, line 5: This should not be hardcoded.

snapshot_maker/snapshot-website-base/_config.yml
```diff
@@ -0,0 +1,55 @@
...
+    color_theme: blue # Available themes: blue (default), green, purple, red and yellow
+    header:
+        logo:
+            text: XTZ-Shots
+            image:
+            logo_url: 'https://xtz-shots.io/index.html'
+        nav:
+            - item_name: Mainnet
+              item_url: 'https://mainnet.xtz-shots.io/'
+            - item_name: Granadanet
```
nicolasochem, line 22: should be hangzhou and ithaca here? I think this is what creates the links on the top right.

snapshot_maker/snapshotMakerJob.yaml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        app: snapshot-maker
+    spec:
+      serviceAccountName: snapshot-maker-sa
+      containers:
+        - name: snapshot-maker
+          image: 082993323195.dkr.ecr.us-east-2.amazonaws.com/snapshot-maker:latest
```
nicolasochem, line 15: should not be hardcoded

snapshot_maker/mainJob.yaml
```diff
@@ -0,0 +1,165 @@
...
+    spec:
+      serviceAccountName: snapshot-maker-sa
+      initContainers:
+        - name: init-tezos-filesystem
+          image: ""
+          command: ["/bin/sh"]
+          args:
+            - "-c"
+            - |
+              #!/bin/sh
```
nicolasochem, line 20: can we put that in a separate file and use Roland's method?

nicolasochem (commented):
Thank you for this, a few initial comments.

harryttd:
Please rename the PR to reflect what it is actually accomplishing

harryttd (changes requested):
This is awesome work!

I'm not done reviewing yet. I've mostly made templating comments.

I still need to go through the snapshot_maker code and understand it better. I'd also like to see if we can perhaps make some values more templatizeable. Bec many values are defined as templatizeable or hardcoded in one chart, but then they are expected to be static in other charts / templates, as they are hardcoded.

charts/snapshot-warmer/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+The reason for this is the longer you wait to take an EBS snapshot, the longer it takes to actually create the snapshot of an AWS EBS Volume.  
+
+If we continuously take snapshots, this assures that the artifact creation process does not have to wait for a snapshot to complete and also always has the newest data to work with.
+
+## Dependencies & Testing
+
+This Helm Chart is dependent on AWS IAM Roles, AWS OIDC Trust Policy, the AWS EKS CSI driver, and AWS IAM policy to perform actions on AWS resources from Kubernetes Pods.
+
+All of these resources can be deployed to a real AWS EKS cluster with https://github.com/oxheadalpha/oxheadinfra. However testing in minikube without AWS mocking tools would not be possible.
```
harryttd, line 21: I don't think we need to mention oxheadinfra in this repo. We can just say you need an EKS cluster. 
 
Also I'm thinking that we should have a full list of all the IAM roles, policies, configurations, etc that we know we need, and provide them in this readme or something. Otherwise it is very unclear how to fully deploy a working chart. 

```suggestion
This Helm Chart is configured to work only with an AWS EKS cluster. It requires AWS IAM Roles and Policies, an AWS OIDC provider, and the EKS CSI driver to perform necessary actions on AWS resources from Kubernetes Pods.

Testing in minikube without AWS mocking tools is not possible. You should create your own EKS cluster to test.
```

harryttd (changes requested):
more comments

snapshot_maker/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+#
+# Installs lz4, jq, yq, kubectl, and awscliv2
+#
+RUN apk --no-cache add \
+        binutils \
+        curl \
+        lz4 \
+        jq  \
+        bash \
+    && wget -q -O /usr/bin/yq $(wget -q -O - https://api.github.com/repos/mikefarah/yq/releases/latest \
```
harryttd, line 14: we might want to pin yq? He's had breaking major changes in the past

charts/snapshot-warmer/scripts/snapshot-warmer.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/sh
+
+cd /
+
+## Snapshot Namespace
+NAMESPACE="${NAMESPACE}" yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
```
harryttd, line 6: I don't think we should need to prefix with `NAMESPACE="${NAMESPACE}"`. It is assigning a variable's value to the value it already is. It is a global env var passed in to the container via the pod specification and should be available everywhere. This comment applies to multiple places in each of the snapshotting bash scripts we have.
```suggestion
yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
```

charts/snapshot-warmer/scripts/snapshot-warmer.sh
```diff
@@ -0,0 +1,101 @@
...
+PERSISTENT_VOLUME_CLAIM=var-volume-snapshot-"${HISTORY_MODE}"-node-0
```
harryttd, line 7: FYI, If we were to pass the node class value the current warmer is deployed for as an env var, this works and we wouldn't have to hardcode the volume's full name:
```shell
PERSISTENT_VOLUME_CLAIM=$(kubectl get -n "$NAMESPACE" po -l node_class="$NODE_CLASS" \
-o jsonpath='{.items[0].spec.volumes[?(@.name=="var-volume")].persistentVolumeClaim.claimName}')
```

It might also be nice combined with the idea of flattening the `nodes` structure in values.yaml to add a `volumeName` field to each node class. That would be in addition to `history_mode` being the only field so far. This would templatize code even more instead of making the snapshot code hardcoded and tightly dependent upon what is currently the default value in tezos-k8s.

```yaml
nodes:
  snapshot-archive-node:
    history_mode: archive
    volumeName: var-volume
```
```shell
kubectl get -n "$NAMESPACE" po -l node_class="$NODE_CLASS"  \
-o jsonpath="{.items[0].spec.volumes[?(@.name==\"$VOLUME_NAME\"})].persistentVolumeClaim.claimName}"
```

charts/snapshot-warmer/scripts/snapshot-warmer.sh
```diff
@@ -0,0 +1,101 @@
...
+
+HISTORY_MODE="${HISTORY_MODE}" yq e -i '.metadata.labels.history_mode=strenv(HISTORY_MODE)' createVolumeSnapshot.yaml
+PERSISTENT_VOLUME_CLAIM="${PERSISTENT_VOLUME_CLAIM}" yq e -i '.spec.source.persistentVolumeClaimName=strenv(PERSISTENT_VOLUME_CLAIM)' createVolumeSnapshot.yaml
+
+while true; do
+
+  # Remove unlabeled snapshots
+  while [ "$(kubectl get volumesnapshots -o jsonpath='{.items[?(.status.readyToUse==true)].metadata.name}' --namespace "${NAMESPACE}" -o go-template='{{len .items}}' --selector='!history_mode')" -gt 0 ]; do
```
harryttd, line 15: This kubectl cmd is run 8 times in this file. It can be made into a function to shorten it a good bit. Something like:
```shell
getSnapshotNames() {
  local readyToUse="${1##readyToUse=}"
  shift
  kubectl get volumesnapshots -o jsonpath="{.items[?(.status.readyToUse==$readyToUse)].metadata.name}" \
    --namespace "$NAMESPACE" "$@"
}
getSnapshotNames readyToUse=false -o go-template='{{len .items}}' --selector='!history_mode'


```

harryttd:
Question on the kubectl commands we use in regards to the sleeping we do: Why not just use `kubectl wait` cmd? 

https://mrkaran.dev/posts/kubectl-wait/
https://kb.novaordis.com/index.php/Kubectl_wait

There is also a `--wait` flag you can pass to cmds. But it may be that the `kubectl wait` is the better way to do it. Could be because you can pass multiple resources to it.

`kubectl wait --help`

snapshot_maker/snapshotMakerJob.yaml
harryttd: Hmmm I just realized that for our comments about hardcoding images, I think we would actually want to hardcode it to `tezos-k8s-snapshot-maker:dev`

orcutt989:
Never seemed to work right unfortunately.

snapshot_maker/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM jekyll/jekyll:builder
```
harryttd, line 1: I think we will have a problem with publishing the snapshot-maker container to docker hub for 2 reasons:

- I don't believe we updated our docker hub credentials in github secrets when we moved from TQ to OA. Looking at the CI logs for dockerfile publish step, I see that we are getting an unauthorized error. The CI code will pass the publish step if the curl command failed to create the docker image repo. It should really check the response from docker hub. If it gets an error, only then error, but if the repo exists we should be able to check for that in the response and then exit successfully.
Screen Shot 2022-02-15 at 1 22 11 AM

Relevant code:https://github.com/oxheadalpha/tezos-k8s/blob/d8b7336d9a4de7098fe35c21c48bc75653e2721a/.github/workflows/ci.yml#L135-L140 

https://github.com/oxheadalpha/tezos-k8s/blob/d8b7336d9a4de7098fe35c21c48bc75653e2721a/.github/workflows/ci.yml#L166-L179


 
- I _think_ this will be an issue for us: The format for docker image naming is to use `-`. Not `_`. The CI will publish the snapshot-maker image as `tezos-k8s-snapshot_maker`, bec `snapshot_maker` is the name of the folder containing the dockerfile. The CI looks at folders containing dockerfiles and uses the name of the folder for the image name. I think we need to keep the format consistent and should rename the file using `-`. We can if we decide to have the CI substitute `_` for `-` and leave the inconsistency. IF we decide to change the folder name, we should likely do so after all the comments have been addressed. Changing the name before hand might mess with the github comments as the files are now in a new directory. I should have made it more explicit in DEVELOPMENT.md to use `-`: https://github.com/oxheadalpha/tezos-k8s/blob/publish-to-xtz-shots/DEVELOPMENT.md#creating-docker-images

harryttd:
>Question on the kubectl commands we use in regards to the sleeping we do: Why not just use kubectl wait cmd?
>> Never seemed to work right unfortunately.

It does seem actually to be considered an experimental command:
```shell
❯ k wait --help
Experimental: Wait for a specific condition on one or many resources.
```
Cool you tried though. Now we know :) 

>

snapshot_maker/snapshot-website-base/_config.yml
```diff
@@ -0,0 +1,57 @@
+title: Tezos Snapshots
```
harryttd, line 1: If I understand correctly how the website is created, the zip and upload job will bundle the Jekyll project and upload it to s3. The Jekyll site is bundled together with the docker image in tezos-k8s.

It is likely more ideal to create a separate repo to manage the website instead of building it inside the docker image and committing it in tezos-k8s. The site is very customized for xtz-shots and we are making xtz-shots and the website implementation a dependency of tezos-k8s. I don't think we should be managing our front end for snapshots inside tezos-k8s. I think there is a line to draw here for separation of concerns. IMO.

We can create a separate repo and then pass a url to the repo as a variable. The job can git clone it and bundle and upload the project. This way a user can optionally have a front end if they want. If they do, they can create it however they want and host it themselves wherever they choose, pointing to the tars in S3. I guess they can also pass a url to their Jekyll site and we can bundle and upload it for them. If they don't want to host a public site, they can just store their tars in S3 and download them with an S3 url.

snapshot_maker/snapshotMakerJob.yaml
orcutt989: this is being overridden by pulumi fyi. pulumi feeds in a value for the image at run time. I think we can make this `""` if we want.

charts/snapshot-warmer/scripts/snapshot-warmer.sh
orcutt989: This is yq specific syntax. Check it out -

```bash
$ NAMESPACE=test
$ yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
$ cat createVolumeSnapshot.yaml
metadata:
  namespace: ""
$ NAMESPACE="${NAMESPACE}" yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
$ cat createVolumeSnapshot.yaml
metadata:
  namespace: "test"
```

snapshot_maker/Dockerfile
orcutt989: I don't know why I used underscore. I think there was a reason, but I can't remember right now.

snapshot_maker/mainJob.yaml
nicolasochem: right, an approach would be to run helm inside the container and deploy the snapshot job using helm. but that's probably a big refactor.

snapshot_maker/snapshotMakerJob.yaml
nicolasochem: ideally, when we do templating we should have a template file, with the sections to replace delineated with `{{}}` or equivalent.

charts/snapshot-warmer/scripts/snapshot-warmer.sh
orcutt989: This is a great suggestion. I have spent some time staring at the code and seeing multiple places where I need to do this, but it would require some refactoring.  Every time I put a loop it, in seems to move faster/more unpredictably than I can put waits in, or sleeps. Results are unpredictable so I stuck with the copy pastes for now.

snapshot_maker/snapshot-website-base/_config.yml
nicolasochem: We are already doing this - it's the jekyll remote theme. 

What we can do is pass the remote theme `oxheadalpha/xtz-shots-website` as a helm variable instead of hardcoding it in the container.

The annoying thing is all the header structured data. It should not be part of _config.yml. This can probably be moved to the remote theme repository.

charts/snapshot-warmer/scripts/snapshot-warmer.sh
harryttd: I just tested it with using `export NAMESPACE` without prepending the env var to the `yq` cmd and it works. Export makes it a global var. The env vars are being passed into the container from the pod as global env vars, so this should work. Was this tested before and it didn't work?
```shell
❯  NAMESPACE=test
❯  yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
❯  cat createVolumeSnapshot.yaml | head -5
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: "2021-11-07-01-13-02-node-snapshot"
  namespace: ""
❯ export NAMESPACE=test
❯ yq e -i '.metadata.namespace=strenv(NAMESPACE)' createVolumeSnapshot.yaml
❯ cat createVolumeSnapshot.yaml | head -5
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: "2021-11-07-01-13-02-node-snapshot"
  namespace: "test"

File: snapshot_maker/Dockerfile (hunk 11 above)
Comment (harryttd) [reply snapshot_maker/Dockerfile:1]: > We did update the creds this summer when we migrated tezos-k8s. I think the issue is that there isn't a repo created for snapshot_maker.

This error is occurring for releases of tezos-k8s that do not have snapshot maker yet.

File: snapshot_maker/snapshot-website-base/_config.yml (hunk 12 above)
Comment (harryttd) [reply snapshot_maker/snapshot-website-base/_config.yml:1]: Oh very cool. Yes we should (optionally) be able to pass a theme in and have the project bundle it. 

We don't need to commit the .gitignore. The gem files I guess we need to be able to build the remote theme.

Do we need to be creating the index.md file in zip and upload? Can't that be moved as well?

Comment (orcutt989) [reply snapshot_maker/snapshot-website-base/_config.yml:1]: I think the separate repo is a great idea, but is a bit of a lift for this PR.  Right now zip and upload pulls the latest artifact info and pushes it that way we prevent any old information getting put up there. I put some time into thinking about it and that's the best thing I could come up with for the time being.  If we had a separate repo we'd need to set up remote API git commits/CI runs.

Comment (nicolasochem) [reply snapshot_maker/snapshot-website-base/_config.yml:1]: @harryttd we can remove gitignore. Yes, the markdown file needs to be created here. It's like a metadata file, but for humans.

Comment (harryttd) [reply snapshot_maker/snapshot-website-base/_config.yml:1]: > @harryttd we can remove gitignore. Yes, the markdown file needs to be created here. It's like a metadata file, but for humans.

From slack discussion: we should be able to add the necessary data to a config file that can be passed to Jekyll to render and not need to create the index.md file in zip&upload.

File: charts/snapshot-scheduler/README.md [hunk 13]
```diff
@@ -0,0 +1,205 @@
...
+
+This Helm Chart is dependent on the AWS EKS CSI driver to create Kubernetes custom resources (VolumeSnapshots and VolumeSnapshotContents) and facilitate the creation of AWS EC2 EBS Volume Snapshots.
+
+Subsequent jobs and pods are dependent on AWS S3 buckets, AWS ACM Certificates, Route 53 DNS Records, and AWS Route 53 domains.
+
+All of these resources can be deployed to a real AWS EKS cluster with https://github.com/oxheadalpha/oxheadinfra. However testing in minikube without AWS mocking tools would not be possible.
+
+## Setup
+
+If your `values.yaml` needs at least 2 Tezos nodes, 1 with **archive** history mode, and another **rolling** history mode.
```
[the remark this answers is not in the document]
orcutt989: Ah you fixed this with your refactor of the chart. Thanks!

charts/snapshot-scheduler/README.md
```diff
@@ -0,0 +1,205 @@
...
+      - ...
+  snapshot-rolling-node:
+  ...
+    instances:
+      - ...
+        config:
+          shell:
+            history_mode: rolling
+      - ...
+```
```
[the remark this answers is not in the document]
orcutt989: Great suggestion. I actually think I was able to simplify this further by just using the `$config.labels.node_class_history_mode` that is already being used.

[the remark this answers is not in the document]
orcutt989: On second thought I think there may be a lot of refactoring involved in this as much of tezos-k8s relies on this structure https://github.com/oxheadalpha/tezos-k8s/search?q=%22config%22+%22history_mode%22.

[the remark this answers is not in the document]
harryttd: Why does this require a lot of refactoring? This is separate from tezos-k8s. it's its own chart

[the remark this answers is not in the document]
orcutt989: My understanding is that tezos-k8s needs `instances[].config.shell.history_mode` for the node config as seen in the link I provided. 

If we get rid of that the tezos node wont know what history mode to start as right? We can add what you suggested, but then we're doubling up history mode information when now we are using the same information for the snapshot maker that tezos-k8s is using.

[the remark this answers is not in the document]
harryttd: It needs that struct when defining nodes in the tezos-k8s chart. The snapshot-engine chart is not defining nodes that should be created. It is targeting nodes that are created by the former chart. Hence you only need a simple flat structure.

I think your point:
> Great suggestion. I actually think I was able to simplify this further by just using the $config.labels.node_class_history_mode that is already being used.

is good and we don't need to specify the history mode in values.yaml. The warmer job (and whatever else that needs to) can `kubectl get statefulset` and check the `node_class_history_mode` to determine the history mode, and fail if the label isn't specified. So instead of https://github.com/oxheadalpha/tezos-k8s/pull/364/files#diff-0aa45a47facc9d1c918bd191fdcd2ef0f6144606426f8c3dbecabed30b9eacedR40, you would just get the label via kubectl. And no more need to also write the hist mode in values.y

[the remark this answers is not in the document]
orcutt989: Thanks for the clarification, but I'm still not understanding something.

If I remove `instances[].config.shell.history_mode` from namespace/values.yaml then the nodes don't create.  Why would we add a duplicate set of values when we could use the ones that are already there for tezos-k8s?

IE
```
nodes:
  rolling-node: null
  snapshot-archive-node:
    >>history_mode: archive
    labels:
      node_class_history_mode: archive
    storage_size: 100Gi
    storageClassName: ebs-sc
    runs:
      - octez_node
    instances:
      - config:
          shell:
            history_mode: archive
  snapshot-rolling-node:
    >>history_mode: rolling
    labels:
      node_class_history_mode: rolling
    storage_size: 80Gi
    storageClassName: ebs-sc
    runs:
      - octez_node
      - metrics
    instances:
      - config:
          shell:
            history_mode: rolling

node_config_network:
  chain_name: hangzhounet

images:
  octez: tezos/tezos:v12.2

protocols:
  - command: 010-PtGRANAD

full_snapshot_url: null
rolling_snapshot_url: null
archive_tarball_url: https://hangzhounet.xtz-shots.io/archive-tarball
rolling_tarball_url: https://hangzhounet.xtz-shots.io/rolling-tarball
```

Snapshot-maker won't be deployed without tezos-k8s, so I am not understanding how it "doesnt need" the tezos k8s values?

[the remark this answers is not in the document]
orcutt989: It was suggested we separate out the values files for tezos-k8s and snapshot-engine, but my concern is that this will involve refactoring all the tezos-k8s places that expects `values.yaml`.

charts/snapshot-scheduler/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+The Helm template loops over **the first** in each type array `snapshot-archive-node` and `snapshot-rolling-node`. If you have multiple rolling/archive nodes only the first one of each type is targeted by the snapshot scheduler.
+
+No additonal values are required other than those required by Tezos-K8s itself.
+
+Deploy with https://github.com/oxheadalpha/oxheadinfra
```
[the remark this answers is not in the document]
orcutt989: Overhauled Readme. Thanks for the feedback!! 🤠

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,441 @@
...
+aws ecr get-login-password --region YOUR_AWS_REGION | docker login --username AWS --password-stdin YOUR_ECR_URL
+
+# Build the image with Docker
+docker build -t snapshotEngine snapshotEngine/
+
+# Tag the image. Will be used in values.yaml
+docker tag snapshotEngine:latest YOUR_ECR_URL/snapshotEngine:latest
+
+# Push the image to ECR
+docker push YOUR_ECR_URL/snapshot-maker:latest
```
harryttd, line 144: `snapshotEngine`

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+This Helm Chart is configured to work only with an AWS EKS cluster. It requires AWS IAM Roles and Policies, an AWS OIDC provider, and the EKS CSI driver to perform necessary actions on AWS resources from Kubernetes Pods.
+
+Testing in minikube without AWS mocking tools is not possible. You should create your own EKS cluster to test.
+
+The Jobs and Pods triggered by this Helm chart are dependent on Kubernetes Service Accounts that depend on AWS IAM Roles, AWS OIDC Trust Policy, the AWS EKS CSI driver, and AWS IAM policy to perform actions on AWS resources from Kubernetes Pods.
+
+This Helm Chart is dependent on the AWS EKS CSI driver to create Kubernetes custom resources (VolumeSnapshots and VolumeSnapshotContents) and facilitate the creation of AWS EC2 EBS Volume Snapshots.
+
+Subsequent jobs and pods are dependent on AWS S3 buckets, AWS ACM Certificates, Route 53 DNS Records, and AWS Route 53 domains.
```
harryttd, line 278: I think this was from the old readme. Don't think we need all of this. Some of it is repeating what is already above and some of it seems to be warmer specific.

Also the resource creation list is a bit inaccurate. Warmer and scheduler deployments are always both created. also they will be created per node history type. So if you have rolling and archive, you will get 4 deployments.
Also a configmap is created.

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,441 @@
...
+This script runs indefinitely and performs the following steps -
+
+1. Removes unlabeled VolumeSnapshots. VolumeSnapshots need to be labeled with their respective `history_mode` in order for them to be used later with the snapshot engine system.
+2. Maintains 4 snapshots of a particular `history_mode` label. There isn't any particular reason for this, other than to keep the list of snapshots concise. "Maintains" meaning deletes the oldest snapshots once the count is over 4.
+3. If there are not any snapshots in progress, then a new one is triggered named with the current time and date.
+4. It waits until the snapshot is ready to use.
+
+We create only one snapshot at a time as having more than one in-progress slows down the snapshot process altogether.
+
+#### Snapshot Scheduler Deployment
```
harryttd, line 295: `####` -> `###` ?

harryttd (changes requested):
Great work on the readme! I left a few more closing comments.

Couple of things:
- Please rename the PR to something like `Snapshot Engine`
- This comment https://github.com/oxheadalpha/tezos-k8s/pull/364#discussion_r806483357 regarding publishing the engine docker image still needs to be addressed.
    - The credential issue is the first thing.
    - I think we will also need to update the github workflow file to properly name the docker images when they are published. We currently name the images `tezos-k8s<DOCKER FOLDER NAME>`. So all of our images consist of `-`.  With snapshot engine being named `snapshotEngine` I think it will end up making our image names inconsistent (`tezos-k8s-snapshotEngine` vs `tezos-k8s-snapshot-engine`. We might want to do some regex in the worflow file to always convert to `-`.

snapshotEngine/mainJob.yaml
```diff
@@ -0,0 +1,234 @@
...
+
+              # Run headless tezos node to validate storage on restored volume
+              tezos-node run --connections 0 --config-file /home/tezos/.tezos-node/config.json --rpc-addr=127.0.0.1:8732 &
+
+              # Time validation to 2 minutes. If this takes longer then there is a tezos error
+              # and this job is tossed.
+              # Tezos does not exit on error so we have to time the job.
+              sleep 2m && kill -s SIGINT 1 &
+
+              # Replace this with kubectl wait.  These loops wait on the RPC to come online and prevent log from printing same line
```
nicolasochem, line 48: Is this a FIXME? Are we actually planning to replace this with `kubectl wait`?

snapshotEngine/mainJob.yaml
```diff
@@ -0,0 +1,234 @@
...
+                exit 1
+
+              # Otherwise if RPC is online, then our storage is valid and we can check if the block 
+              # is finalized and get our metadata from the RPC endpoint.
+              else
+                printf "%s Node online! Storage is initialized.\n" "$(date "+%Y-%m-%d %H:%M:%S" "$@")"
+              if [ "${HISTORY_MODE}" = archive ]; then
+                TARGET="head"
+
+              # Tezos devs have advised us that it is safer to target HEAD~2 for rolling artifacts.
```
nicolasochem, line 71:
```suggestion
              # In tenderbake, any block at HEAD~2 is final. We are targeting HEAD~2 for rolling artifacts.
              # Otherwise, we may create a snapshot from an uncle block, which would not import succesfully.
```

snapshotEngine/zip-and-upload.sh
```diff
@@ -0,0 +1,564 @@
...
+
+# Store network name for liquid templating
+jq -n \
+--arg NETWORK "$NETWORK" \
+'{
+  "network": $NETWORK
+}' > _data/tezos_metadata.json
+
+# Grab liquid-templated chain website page
+curl -o index.md https://raw.githubusercontent.com/oxheadalpha/xtz-shots-website/master/snapshot.md
```
nicolasochem, line 551: We spent time separating out the website content from the snapshotter jobs, so perhaps this URL should be a helm value that we turn into an env var, instead of hardcoded?

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,441 @@
...
+        Effect: "Allow",
+        Resource: [
+          "ARN_OF_S3_BUCKET",
+          "ARN_OF_S3_BUCKET/*",
+        ],
+      },
+    }
+  ```
+
+3. Scope this new IAM role with a Trust Policy wit the following content:
```
nicolasochem, line 105:
```suggestion
3. Scope this new IAM role with a Trust Policy with the following content:
```

charts/snapshotEngine/scripts/snapshot-warmer.sh
```diff
@@ -0,0 +1,127 @@
...
+      exit 1
+    fi
+
+    sleep 5
+
+    # While no snapshots ready
+    while [ "$(getSnapshotNames readyToUse=false -l history_mode="${HISTORY_MODE}")" ]; do
+      printf "%s Snapshot is still creating...\n" "$(timestamp)"
+      sleep 10
+      # # Get EBS snapshot progress
```
nicolasochem, line 87: what's with this commented out section?

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,441 @@
...
+11. Rolling tarball filesize is generated.
+12. Metadata is added to `base.json` and uploaded.
+13. Build artifact-specific metadata json file and upload it to AWS S3 Bucket.
+14. Create and upload rolling tarball redirect file. This forwards to the latest archive artifact. (EX. `mainnet.xtz-shots.io/rolling-tarball >> mainnet.xtz-shots.io/tezos-mainnet-rolling-tarball-203942.lz4`)
+15. Upload `.rolling` file to S3 AWS bucket.
+16. Generate filesize, and SHA256 sum of `.rolling` file.
+17. Add metadata to `base.json` and upload.
+18. Add metadata to artifact-specific json and upload.
+19. Get metadata from artifact json files (curl) for web page.
+20. Build web page with Jekyll
```
nicolasochem, line 440: A section about jekyll would be good? Say that the engine expects an external jekyll repo, with a specific `snapshot.md` file in it.

snapshotEngine/snapshot-website-base/_config.yml
```diff
@@ -0,0 +1,60 @@
...
+description: > # this means to ignore newlines until "show_excerpts:"
+  Tezos Snapshots
+favicon: "./favicon.ico" # set the favicon of the site 
+
+show_excerpts: false # set to true to show excerpts on the homepage
+# url becomes localhost in development environments
+url: https://xtz-shots.io
+domain_name: xtz-shots.io
+
+remote_theme: oxheadalpha/xtz-shots-website
```
nicolasochem, line 12: We are still using remote theme but we relocated `snapshot.md` to the other repo? Was this the compromise?

charts/snapshot-scheduler/values.yaml
```diff
@@ -0,0 +1,4 @@
+tezos_k8s_images:
+  snapshot_maker: tezos-k8s-snapshot-maker:latest
+images:
+  octez: tezos/tezos:v11-release
```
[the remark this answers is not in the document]
nicolasochem: Is it possible to have the default match the default of tezos-k8s? so if I install snapshot-engine without values.yaml and the tezos chart without values.yaml, the snapshot engine will find the right image to snapshot? I think it would be `archive-node-0` in this case.

nicolasochem (approved):
Let's merge, all the remaining concerns can be addressed in subsequent PRs

Thank you! This was a huge chunk of work, as the length of the README can testify.

charts/snapshotEngine/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
```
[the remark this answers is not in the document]
harryttd: @orcutt989 The description was still missed. Please either make an issue or commit

charts/snapshotEngine/values.yaml
```diff
@@ -0,0 +1,16 @@
+tezos_k8s_images:
+  snapshotEngine: tezos-k8s-snapshot-maker:dev
```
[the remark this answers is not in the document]
harryttd: @orcutt989 Please see my previous comment. The image name was still not addressed and I'm not certain that the published chart will work because it will be looking for an image that doesn't exist. Please fix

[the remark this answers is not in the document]
nicolasochem: Looks like it worked in the published chart?

```
nochem@fedora /tmp $ wget https://github.com/oxheadalpha/tezos-helm-charts/raw/main/snapshotEngine-6.5.0.tgz
--2022-04-13 14:25:43--  https://github.com/oxheadalpha/tezos-helm-charts/raw/main/snapshotEngine-6.5.0.tgz
Resolving github.com (github.com)... 192.30.255.112
Connecting to github.com (github.com)|192.30.255.112|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://raw.githubusercontent.com/oxheadalpha/tezos-helm-charts/main/snapshotEngine-6.5.0.tgz [following]
--2022-04-13 14:25:44--  https://raw.githubusercontent.com/oxheadalpha/tezos-helm-charts/main/snapshotEngine-6.5.0.tgz
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.111.133, 185.199.108.133, 185.199.110.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.111.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 8448 (8.2K) [application/octet-stream]
Saving to: ‘snapshotEngine-6.5.0.tgz’

snapshotEngine-6.5.0.tgz                      100%[==============================================================================================>]   8.25K  --.-KB/s    in 0.001s  

2022-04-13 14:25:44 (16.1 MB/s) - ‘snapshotEngine-6.5.0.tgz’ saved [8448/8448]

nochem@fedora /tmp $ tar xvfz snapshotEngine-6.5.0.tgz 
snapshotEngine/Chart.yaml
snapshotEngine/values.yaml
snapshotEngine/templates/configmap.yaml
snapshotEngine/templates/rbac.yaml
snapshotEngine/templates/snapshot-scheduler.yaml
snapshotEngine/templates/snapshot-warmer.yaml
snapshotEngine/README.md
snapshotEngine/scripts/snapshot-warmer.sh
nochem@fedora /tmp $ cd snapshotEngine/
nochem@fedora /tmp/snapshotEngine $ cat values.yaml  
tezos_k8s_images:
  snapshotEngine: oxheadalpha/tezos-k8s-snapshot-engine:6.5.0
images:
  octez: tezos/tezos:v12.2
```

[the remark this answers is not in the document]
nicolasochem: ah, nevermind, it should be `tezos-k8s-snapshotengine:6.5.0`, this indeed needs fixing.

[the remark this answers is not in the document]
harryttd: JImmy did a commit directly to master after i pointed it out. I guess the release missed it?

I am making a quick PR for some other things. i will add this

charts/snapshotEngine/README.md
```diff
@@ -0,0 +1,414 @@
...
+    history_mode: archive
+    target_volume: var-volume
+  snapshot-rolling-node:
+    history_mode: rolling
+    target_volume: var-volume
+
+images:
+  octez: tezos/tezos:v12.2 # Version of Tezos that you will run
+
+chainWebsiteMarkdown: url_to_md_file # Url of markdown file that will host links to your
```
harryttd, line 217: This section needs to be updated based on comments we spoke about over slack

PR #368, opened 2022-02-17
add ability to pass node identity as values (fix #311)

### LINKED ISSUE #311: add ability to pass identity file as helm variable
It is useful especially for local storage: we will save 1-2 minutes each time the pod reboots.

nicolasochem:
with the advent of tarballs, spinning up tezos nodes with local storage
is very fast. But each time it starts, it generates a new identity, which
requires solving a proof-of-work puzzle which takes few seconds to 1-2
minutes.

You can now generate these identities out-of-band, and record them in
values.yaml at node level, and save this time at boot of your local
storage-backed node.

This is not supported at node_class or global level since each identity
must be unique.

~A controversial choice I made is to add these inside the tezos-config
configMap instead of a secret. Thinking about it, these identity keys
are not that secret, even though they use public key cryptography and
have a "secret" key. The impact of a leak is that you can spoof a peer
and potentially knock it off the network. It is not bad enough that we
would want to create a k8s secret. Due to helm limitations, we would have to
make a new top-level object in values.yaml especially for secrets, which
must match the structure of `nodes`. This would not be great for
usability.~

The identity contains a secret key, so it is stored in a kubernetes secret and pruned from the configmap containing all of the node's info.

~That being said, I add a `set +x` around the bottom of
config-generator.sh so we don't unnecessarily print this key in the
logs.~

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -45,6 +45,10 @@
     - "config-generator"
     - "--generate-config-json"
   envFrom:
+    {{- if len .node_identities }}
+    - secretRef:
+        name: {{ .node_class }}-indentities-secret
+    {{- end }}
     - secretRef:
         name: tezos-secret
     - configMapRef:
--- a/charts/tezos/templates/_helpers.tpl
+++ b/charts/tezos/templates/_helpers.tpl
@@ -86,11 +86,36 @@
   Returns a string "true" or empty string which is falsey.
 */}}
 {{- define "tezos.shouldConfigInit" }}
-{{- if not (.Values.node_config_network.genesis) }}
-{{- "true" }}
-{{- else }}
-{{- "" }}
+  {{- if not (.Values.node_config_network.genesis) }}
+    {{- "true" }}
+  {{- else }}
+    {{- "" }}
+  {{- end }}
 {{- end }}
+
+{{/*
+  If a Tezos node identity is defined for an instance, create a secret
+  for its node class. All identities for all instances of the node
+  class will be stored in it. Each instance will look up its identity
+  values by its hostname, e.g. archive-node-0.
+  Returns a string "true" or empty string which is falsey.
+*/}}
+{{- define "tezos.includeNodeIdentitySecret" }}
+  {{- range $index, $config := $.node_vals.instances }}
+    {{- if .identity }}
+      {{- $_ := set $.node_identities (print $.node_class "-" $index) .identity }}
+    {{- end }}
+  {{- end }}
+  {{- if len $.node_identities }}
+apiVersion: v1
+data:
+  NODE_IDENTITIES: {{ $.node_identities | toJson | b64enc }}
+kind: Secret
+metadata:
+  name: {{ $.node_class }}-indentities-secret
+  namespace: {{ $.Release.Namespace }}
+---
+  {{- end }}
 {{- end }}
 
 {{/*
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -17,8 +17,20 @@ data:
   ROLLING_TARBALL_URL: "{{ .Values.rolling_tarball_url }}"
   NODE_GLOBALS: |
 {{ .Values.node_globals | mustToPrettyJson | indent 4 }}
+
   NODES: |
-{{ .Values.nodes | mustToPrettyJson | indent 4 }}
+{{/*
+  Remove "identity" field from nodes. Creates a deep copy because
+  "unset" modifies the object.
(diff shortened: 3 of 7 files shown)


harryttd:
This is a great idea :)

I'm not so sure we should be putting the ids in plain text in a configmap. Controversial, lol, I know :). IMO if it is a vulnerability, it's a vulnerability and we should do everything to not introduce it. 

Can we expect users of tezos-k8s to be ok with this? Especially those who might use our setup for baking? What about us as a company with our upcoming production baking? We shouldn't be introducing vulnerabilities. Any data put into the wrong hands that can cause damage should be properly handled and secured.

What are everyone's thoughts?

Why is it so difficult to introduce a k8s secret, and why would this negatively effect usability? We can do this which I tested locally with `helm template`:

```yaml
# values.yaml
archive-node:
  config:
    shell:
      history_mode: archive
  instances:
    - identity:
        peer_id: "id"
        public_key: "id"
        private_key: "id"
        proof_of_work_stamp: "id"
rolling-node:
  config:
    shell:
      history_mode: rolling
  instances:
  - identity:
      peer_id: "id1"
      public_key: "id1"
      private_key: "id1"
      proof_of_work_stamp: "id1"
  - {}
  - identity:
      peer_id: "id3"
      public_key: "id3"
      private_key: "id3"
      proof_of_work_stamp: "id3"
```
```yaml
# nodes.yaml We do this in nodes.yaml so we can conditionally create the secret
# and use `envFrom` only if it will be created.
  {{- $_ := set $ "node_identities" dict }}
  {{- range $index, $_ := $.node_vals.instances }}
    {{- if .identity }}
      {{- $_ := set $.node_identities (toString $index) .identity }}
    {{- end }}
  {{- end }}
  {{- if len $.node_identities }}
apiVersion: v1
data:
  {{ $.node_identities | toJson }}
kind: Secret
metadata:
  name: {{ $.node_class }}-indentities-secret
  namespace: {{ $.Release.Namespace }}
---
  {{- end }}
```
```yaml
# _containers.tpl
...
    {{- if len .node_identities }}
    - secretRef:
        name: {{ .node_class }}-indentities-secret
    {{- end }}
```
```shell
❯  helm template test ./charts/tezos  -s templates/nodes.yaml
---
# Source: tezos-chain/templates/nodes.yaml
apiVersion: v1
data:
  # {"0":{"peer_id":"id","private_key":"id","proof_of_work_stamp":"id","public_key":"id"}}
  eyIwIjp7InBlZXJfaWQiOiJpZCIsInByaXZhdGVfa2V5IjoiaWQiLCJwcm9vZl9vZl93b3JrX3N0YW1wIjoiaWQiLCJwdWJsaWNfa2V5IjoiaWQifX0=
kind: Secret
metadata:
  name: archive-node-indentities-secret
  namespace: default
---
# Source: tezos-chain/templates/nodes.yaml
apiVersion: v1
data:
  # {"0":{"peer_id":"id1","private_key":"id1","proof_of_work_stamp":"id1","public_key":"id1"},"2":{"peer_id":"id3","private_key":"id3","proof_of_work_stamp":"id3","public_key":"id3"}}
  eyIwIjp7InBlZXJfaWQiOiJpZDEiLCJwcml2YXRlX2tleSI6ImlkMSIsInByb29mX29mX3dvcmtfc3RhbXAiOiJpZDEiLCJwdWJsaWNfa2V5IjoiaWQxIn0sIjIiOnsicGVlcl9pZCI6ImlkMyIsInByaXZhdGVfa2V5IjoiaWQzIiwicHJvb2Zfb2Zfd29ya19zdGFtcCI6ImlkMyIsInB1YmxpY19rZXkiOiJpZDMifX0=
kind: Secret
metadata:
  name: rolling-node-indentities-secret
  namespace: default
---
# Source: tezos-chain/templates/nodes.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: archive-node
...
        envFrom:
          - secretRef:
              name: archive-node-indentities-secret
---
# Source: tezos-chain/templates/nodes.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: rolling-node
...
        envFrom:
          - secretRef:
              name: rolling-node-indentities-secret
```
Then for config-generator:
```python
node_identity = NODE_IDENTITIES.get(MY_POD_CLASS).get(MY_POD_INDEX)
```

harryttd (changes requested):

harryttd:
Thank you

I can get it in a branch. This is under the `nodes` section. It isn't a new top level field and doesn't go in the configmap. I forgot to show that but it should be simple enough to loop through and `omit` the identity fields.

harryttd (approved):
Thank you!

PR #369, opened 2022-02-18
make tezos-client command work without passing client dir as param

nicolasochem:
This is very neat because I can open a shell into the node or baker and
type `tezos-client list known addresses` and it just works.

Q: why a symink instead of moving the actual dir to
/home/tezos/.tezos-client ?

A: the potential for breakage is high if we move the dir. This makes the
startup script a bit longer but there is virtually no risk of breakage.

### DIFF
--- a/charts/tezos/scripts/baker-endorser.sh
+++ b/charts/tezos/scripts/baker-endorser.sh
@@ -17,6 +17,9 @@ my_baker_account="$(cat /etc/tezos/baker-account )"
 CLIENT="$TEZ_BIN/tezos-client -d $CLIENT_DIR"
 CMD="$TEZ_BIN/tezos-$DAEMON-$proto_command -d $CLIENT_DIR"
 
+# ensure we can run tezos-client commands without specifying client dir
+ln -s /var/tezos/client /home/tezos/.tezos-client
+
 while ! $CLIENT rpc get chains/main/blocks/head; do
     sleep 5
 done
--- a/charts/tezos/scripts/tezos-node.sh
+++ b/charts/tezos/scripts/tezos-node.sh
@@ -2,6 +2,8 @@ set -x
 
 set
 
+# ensure we can run tezos-client commands without specifying client dir
+ln -s /var/tezos/client /home/tezos/.tezos-client
 #
 # Not every error is fatal on start.  In particular, with zerotier,
 # the listen-addr may not yet be bound causing tezos-node to fail.


charts/tezos/scripts/baker-endorser.sh
```diff
@@ -17,6 +17,9 @@ my_baker_account="$(cat /etc/tezos/baker-account )"
 CLIENT="$TEZ_BIN/tezos-client -d $CLIENT_DIR"
 CMD="$TEZ_BIN/tezos-$DAEMON-$proto_command -d $CLIENT_DIR"
 
+# ensure we can run tezos-client commands without specifying client dir
+ln -s /var/tezos/client /home/tezos/.tezos-client
+
```
harryttd, line 22: This is a great trick :)

A few things:
- I don't think the cmd just like this will work. We need to remove the directory first. Otherwise the symlink gets created _inside_ the target. Instead of becoming the target. https://askubuntu.com/questions/600714/creating-a-symlink-from-one-folder-to-another-with-different-names

  ```shell
  ~ $ ln -s /var/tezos/client /home/tezos/.tezos-client/
  ~ $ ls -l /home/tezos/.tezos-client/
  total 0
  lrwxrwxrwx    1 tezos    nogroup         17 Feb 18 05:20 client -> /var/tezos/client
  ~ $ tezos-client list known addresses
  ~ $ unlink /home/tezos/.tezos-client/client
  ~ $ ls -l /home/tezos/.tezos-client/
  total 0
  
  # Now remove the directory
  ~ $ rm -r .tezos-client/
  ~ $ ln -s /var/tezos/client/ .tezos-client
  ~ $ ls -la
  ...
  lrwxrwxrwx    1 tezos    nogroup         18 Feb 18 05:43 .tezos-client -> /var/tezos/client/
  ~ $ ls -l .tezos-client/
  total 32
  -rw-r--r--    1 tezos    nogroup      10297 Feb 18 05:44 NetXxkAx4woPL_baker_state
  -rw-r--r--    1 tezos    nogroup        408 Feb 18 05:44 NetXxkAx4woPL_highwatermarks
  -rw-r--r--    1 tezos    nogroup       4082 Feb 18 05:38 NetXxkAx4woPL_nonces
  -rw-r--r--    1 root     root           661 Feb 18 00:04 public_key_hashs
  -rw-r--r--    1 root     root          1655 Feb 18 00:04 public_keys
  -rw-r--r--    1 root     root           132 Feb 18 00:04 secret_keys
  -rw-r--r--    1 tezos    nogroup          0 Feb 18 00:08 wallet_lock
  ~ $ tezos-client list known addresses
  oxheadbaker: tz1....
  ```
- Why not move the `ln` above here so we don't need `-d`? https://github.com/oxheadalpha/tezos-k8s/blob/9a07742cbb341acc0b279129d828ea98448ebed2/charts/tezos/scripts/baker-endorser.sh#L17-L18
- You mentioned this makes the startup script take longer. How much longer?
- Do you think we would ever want in the future to stop using `/var/tezos` and use the default `.tezos-client` dir instead? Or there's not much of a reason to switch at this point.

harryttd: sounds great. thank you :)

harryttd (approved):

PR #370, opened 2022-02-18
move to hangzhou

nicolasochem:
Move default mkchain output to hangzhou private chain, and change some references from granada to hangzhou.

I tried doing mkchain with no parameters and launching the chain, hangzhou got activated and hangzhou baker & endorser were running.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -219,7 +219,7 @@ expected_proof_of_work: 26
 ## - If specifying the name of the network, the network's name must be
 ##   recognized by the docker image passed as .Values.images.octez above.
 ## - Additionally, you can pass a url that returns the config.json of the
-##   network. Example: "https://teztnets.xyz/granadanet". In general you
+##   network. Example: "https://teztnets.xyz/mondaynet". In general you
 ##   shouldn't need this and should just specifiy the network's name.
 node_config_network:
   chain_name: mainnet
@@ -231,7 +231,7 @@ protocols:
   # You may want to run two or more if you are expecting an on-chain upgrade.
   # Be mindful that running more protocols will start more containers in your
   # pods which may consume more memory.
-  - command: 010-PtGRANAD
+  - command: 011-PtHangz2
   #- command: alpha
 
 ## Activate a new chain from genesis. Uncommenting this will create an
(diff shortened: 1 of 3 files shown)


charts/tezos/values.yaml
```diff
@@ -240,38 +240,42 @@ protocols:
...
-#     minimal_block_delay: "2"
-#     liquidity_baking_subsidy: "2500000"
-#     liquidity_baking_sunset_level: 525600
-#     liquidity_baking_escape_ema_threshold: 100000
+#     initial_endorsers: 192
+#     delay_per_missing_endorsement: "2"
+#     liquidity_baking_escape_ema_threshold: 666667
+#     liquidity_baking_subsidy: '2500000'
+#     liquidity_baking_sunset_level: 10000000
+#     minimal_block_delay: '5'
```
harryttd, line 278: Just wanted to make sure that we don't intend for all of these values to be exactly the same as what are the default params for Hnet: https://gitlab.com/tezos/tezos/-/blob/master/src/proto_011_PtHangz2/lib_parameters/default_parameters.ml#L58-65

nicolasochem: it depends. the subsidy and threshold should match mainnet - we accidentally disabled LB on hangzhounet because the threshold was too low. I pushed a fix, thank you for your vigilance.

The sunset level is set to 10000000 to ensure it virtually never stops. After all, a testnet is for testing so we don't want the ability to test to stop.

Delay per missed endorsement is shorter, just like the delay between blocks is shorter.

And, stating the obvious, this code only matters if someone looks at values.yaml and uncomments it because they want to activate a chain. Which may happen :)

mkchain/tqchain/parameters.yaml
```diff
@@ -3,27 +3,32 @@ blocks_per_cycle: 8
...
-initial_endorsers: 1
-delay_per_missing_endorsement: "1"
-baking_reward_per_endorsement: ["200000"]
-minimal_block_delay: "2"
-liquidity_baking_subsidy: "2500000"
-liquidity_baking_sunset_level: 525600
-liquidity_baking_escape_ema_threshold: 100000
+initial_endorsers: 192
+delay_per_missing_endorsement: "2"
+liquidity_baking_escape_ema_threshold: 666667
```
harryttd, line 31: I think this also needs to be updated to `1000000`

nicolasochem: why? it is 666,667 on mainnet.

harryttd: It's not the same as in values.yaml

nicolasochem: ok, I switched values.yaml to 666,667 instead. Please approve

orcutt989 (approved):

PR #371, opened 2022-02-18
Update snapshot warmer

harryttd:
Addresses some comments from pr #364

### DIFF
--- a/charts/snapshot-warmer/templates/snapshotWarmer.yaml
+++ b/charts/snapshot-warmer/templates/snapshotWarmer.yaml
@@ -36,8 +36,10 @@ spec:
           env:
             - name: NAMESPACE
               value: {{ $.Release.Namespace }}
-            - name: HISTORY_MODE
-              value: {{ $.history_mode }}
+            - name: NODE_CLASS
+              value: {{ $node }}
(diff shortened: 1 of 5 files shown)


orcutt989:
image

orcutt989:
Testing now

charts/snapshot-warmer/templates/snapshotWarmer.yaml
```diff
@@ -1,41 +1,56 @@
...
+{{ $.Files.Get "scripts/snapshot-warmer.sh" | indent 14 }}
           env:
             - name: NAMESPACE
               value: {{ $.Release.Namespace }}
-            - name: HISTORY_MODE
-              value: {{ $history_mode }}
+            - name: NODE_CLASS
+              value: {{ $node }}
+            - name: NODE_CONFIG
+              value: {{ $config | toJson | quote }}
```
harryttd, line 42: @orcutt989 The reason for the issue we were having on Friday was due to me by mistake setting this as `values` instead of `value`.....  Helm locally will run the `template` cmd fine. But when installing the chart it will fail. All pulumi does is run template to render the yaml so that's why it wasn't failing for us.

PR #372, opened 2022-02-23
Use secrets for node identities

harryttd:


charts/tezos/values.yaml
```diff
@@ -92,6 +92,11 @@ accounts: {}
 ##   - "bake_using_account": Account name that should be used for baking.
 ##   - "bake_using_accounts": List of account names that should be used for baking.
 ##   - "is_bootstrap_node": Is this node a bootstrap peer.
+##   - "identity": An optional map containing a pre-generated Tezos node
+##                 identity. The identity file will be created at
+##                 /var/tezos/node/data/identity.json. Required fields are
+##                 "peer_id", "public_key", "secret_key", and
+##                 "proof_of_work_timestamp".
```
nicolasochem, line 99: isn't it redundant with the info that I added? also, identity can only be set at node level to avoid duplicates. Your explanation looks better than mine, so it's fine to remove mine, but there should be a "NOTE: you can only set this as instance level since each identity must be unique"

harryttd: Thanks, I shall remove

harryttd: FYI I also changed `private_key` to `secret_key`. I don't know if the former also works but the latter is what i see in identity.json files.

nicolasochem (approved):
I tried and it works. I looked at the configmap and the identity was not there. There was a secret for the identity, and the identity hash I saw on the node output matched the hash I provided in the values.yaml
Thank you

PR #392, opened 2022-03-11
handle new mandatory liquidity baking vote in alpha

nicolasochem:
The new protocol now turns the liquidity baking escape vote into an escape
toggle and make it mandatory, see [this MR](https://gitlab.com/tezos/tezos/-/merge_requests/4201)

This broke tezos-k8s, and consequently dailynet and mondaynet.

According to tzip, third-party baking software is encouraged to not
provide any default to the user, so they are forced to take a stand.

There are two ways to vote, either pass --liquidity-baking-toggle-vote
on/off/pass or pass a --votefile with your vote in json format.

Even when a votefile is passed, --liquidity-baking-toggle-vote is
mandatory, which I think is bad UX. I [opened a
ticket](https://gitlab.com/tezos/tezos/-/issues/2641)

My plan, when this ticket is fixed, is to allow tezos-k8s users to
simply pass structured vote data as a value to "runs" "baker".

For now I am leaving this PR as draft.

For example:

```
nodes:
  rolling-node:
    runs:
    - octez_node
    - baker:
        vote:
          liquidity_baking_toggle_vote: pass
```

This structured data is then passed to the baker daemon as a votefile.

This satisfies the requirement that tezos-k8s does not default to "pass"
and forces the user to take a stand.

This will allow us not to write special case satements in the baker
entrypoint script to pass this or that flag based on which protocol we
are running...

For now this draft PR is just a fix to make dailynet and mondaynet
deploy. I'll deploy it to teztnets as a submodule.


elric1:
I'd like to suggest a few things:
1. you should be able to specify your voting intention globally as well as per baker.
2. I think that we should provide a default for non-mainnet because it doesn't matter for test networks and likely doesn't make sense for private chains in the first place.
3. I'm not sure that it's a good idea to put it in the "runs" key section... Hmmm. I'll think about that, though.

harryttd:
Would it be normal for multiple bakers you are running to vote differently? Especially if they are all part of the same not class?

Perhaps if you are running not just your baker but someone else's. But I would think such a case a user would create multiple node classes. Maybe we shouldn't assume anything though. 

I'm just thinking like Roland about maybe making the value global and also on the node class level like we have for the config.json values.

elric1:
Another quick comment: can we make the helm templating throw an error if this isn't specified? I think that the UX would be much better if the chart failed to install rather than installed and failed 5 minutes later.

nicolasochem:
Good suggestions, thank you. I am waiting for feedback from nomadic about the CLI itself before moving forward.

The issue with validating helm template is that it is highly protocol dependent: ithaca has no mandatory flags, J has one, maybe K has two? How do we validate in this case?

Yes, it's realistic to bake for 2 bakers with different flags. Why would you do 2 node classes? You could attach 2 baker processes to one node (within the same pod).

It's possible to bake for 2 addresses with just one baker process. In this case, it should also be possible to pass flags per baker into this file. Today, it's not possible. [I opened a ticket, I hope they fix it soon](https://gitlab.com/tezos/tezos/-/issues/2384)

nicolasochem:
Let's just merge this? The octez fix I was asking for [didn't make it into 13.0rc1](https://gitlab.com/tezos/tezos/-/issues/2641) and we will inevitably bump more and more into this.

This is an acceptable workaround, meanwhile.

orcutt989 (approved):

charts/tezos/scripts/baker.sh
```diff
@@ -8,6 +8,14 @@ NODE_DATA_DIR="$TEZ_VAR/node/data"
 
 proto_command="{{ .command_in_tpl }}"
 
+if [ "${proto_command}" == "012-Psithaca" ]; then
+    extra_args=""
+else
+    echo '{"liquidity_baking_toggle_vote": "pass"}' > /etc/tezos/per_block_votes.json
```
harryttd, line 14: is this supposed to be templatized?

nicolasochem: you mean, pass the choice from values.yaml? I just want to hardcode pass and move on. Once 13.0rc2 is out we will handle baker's voting properly.

harryttd (approved):

PR #396, opened 2022-03-22
Install gems in Dockerfile instead of in zip and upload job

harryttd:
The zip and upload job is currently installing gems every time it is run. This takes a lot of time and isn't necessary. This PR installs the gems in the Dockerfile.

@orcutt989  I tested this over a week ago. It should work, but as part of your work on templating everything in the jekyll theme, please make sure it works fully.


snapshot_maker/zip-and-upload.sh
```diff
@@ -520,218 +520,35 @@ cd /srv/jekyll || exit
...
+chmod -R 777 snapshot.md
 chmod -R 777 _data
-chown jekyll:jekyll -R /usr/gem
-
-# convert to index.html with jekyll
-bundle install
 bundle exec jekyll build
 
+# mv _site/snapshots/index.html to _site/index.html
+mv _site/snapshot/*
```
harryttd, line 550: does this need a destination directory? only a source is specified

snapshot_maker/zip-and-upload.sh
```diff
@@ -520,218 +520,35 @@ cd /srv/jekyll || exit
...
-# archive tarball block timestamp
-ARCHIVE_TARBALL_BLOCK_TIMESTAMP=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/archive-tarball-metadata 2>/dev/null | jq -r '.block_timestamp')
-# archive tarball filesize
-ARCHIVE_TARBALL_FILESIZE=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/archive-tarball-metadata 2>/dev/null | jq -r '.filesize')
-# archive tarball sha256 sum
-ARCHIVE_TARBALL_SHA256SUM=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/archive-tarball-metadata 2>/dev/null | jq -r '.sha256')
-# archive tarball tezos version
-ARCHIVE_TARBALL_TEZOS_VERSION=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/archive-tarball-metadata 2>/dev/null | jq -r '.tezos_version')
+# clone https://github.com/oxheadalpha/xtz-shots-website
+git clone https://github.com/oxheadalpha/xtz-shots-website
```
harryttd, line 529: Just to mention:
- We discussed making the url templatizeable. I'm burned out from this tho so maybe we can push it to later when we use Helm in the scripts instead of yq....
- You can add `--depth 1` flag to the clone to only clone the latest commit instead of the whole history and all the branches
- Curious, unless unintentional, what is the point of line 528 which is literally the same as line 529 except for the word `git`? Same with line 531. They are the same command being written twice in a row.

snapshot_maker/zip-and-upload.sh
orcutt989: Great catch I forgot the `.` Fixed by [6acace3](https://github.com/oxheadalpha/tezos-k8s/pull/396/commits/6acace333f76509f5b777f6d5034b2f5cd190e57)

snapshot_maker/zip-and-upload.sh
orcutt989: Happy to templatize the repo, I just wanted to make sure I had the right idea here.

I will add the --depth 1, great suggestion thanks!

528 and 531 were just comments left over from me planning out what to do before I wrote the code. I will update them to convey more useful information.

snapshot_maker/zip-and-upload.sh
```diff
@@ -520,218 +520,35 @@ cd /srv/jekyll || exit
...
-\`\`\`
-### More details
-[About xtz-shots.io](https://xtz-shots.io/getting-started/).
-
-[Tezos documentation](https://tezos.gitlab.io/user/snapshots.html){:target="_blank"}.
-EOF
-
-chmod -R 777 index.md
+# TODO: Can this be done in the container?
+chmod -R 777 snapshot.md
```
harryttd, line 545: I'm pretty sure that we don't need this bec snapshot.md is coming from the theme repo now. Not certain

snapshot_maker/zip-and-upload.sh
```diff
@@ -533,13 +533,12 @@ jq -n \
...
   "network": $NETWORK
-}' > tezos-metadata.json
+}' > _data/tezos-metadata.json
 
 # Grab liquid-templated chain website page
 curl -o index.md https://raw.githubusercontent.com/oxheadalpha/xtz-shots-website/snapshots-md-updates/snapshot.md
 
-chmod -R 777 index.md
-chmod -R 777 _data
+chmod -R 777 ./*
```
harryttd, line 541: Is there another way to set the perms without giving full read/write/execute perms to all users? What if the owner of the dir is set to jekyll? Maybe it already is. I dont know. But 777 isn't the securest thing to do :)

harryttd: I think `chown -R jekyll:jekyll /snapshot-website-base` may work. i did this in the docker container

orcutt989: Great suggestion thanks! 🤠

nicolasochem (approved):

snapshot_maker/snapshot-website-base/_config.yml
```diff
@@ -6,7 +6,7 @@ favicon: "./favicon.ico" # set the favicon of the site
 
 show_excerpts: false # set to true to show excerpts on the homepage
 
-remote_theme: oxheadalpha/xtz-shots-website
+remote_theme: oxheadalpha/xtz-shots-website@snapshots-md-updates
```
harryttd, line 9: Branch needs to change before merging

snapshot_maker/zip-and-upload.sh
```diff
@@ -528,223 +528,35 @@ else
...
-# rolling snapshot filesize
-ROLLING_SNAPSHOT_FILESIZE=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/rolling-snapshot-metadata 2>/dev/null | jq -r '.filesize')
-# rolling snapshot sha256 sum
-ROLLING_SNAPSHOT_SHA256SUM=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/rolling-snapshot-metadata 2>/dev/null | jq -r '.sha256')
-# rolling snapshot tezos version
-ROLLING_SNAPSHOT_TEZOS_VERSION=$(curl -L http://"${S3_BUCKET}".s3-website.us-east-2.amazonaws.com/rolling-snapshot-metadata 2>/dev/null | jq -r '.tezos_version')
-
-CLOUDFRONT_URL="https://${S3_BUCKET}/"
+# Grab liquid-templated chain website page
+curl -o index.md https://raw.githubusercontent.com/oxheadalpha/xtz-shots-website/snapshots-md-updates/snapshot.md
```
harryttd, line 551: Branch needs to change before merging

harryttd (commented):
thank you. a couple of things need to be updated before merging

snapshot_maker/snapshot-website-base/_config.yml
orcutt989: all set

snapshot_maker/zip-and-upload.sh
orcutt989: all set

orcutt989:
@harryttd i think we are good now can you double check

harryttd:
Looks good thank you. I will merge now.

PR #397, opened 2022-03-24
make readiness probe optional

nicolasochem:
For Teztnets, it's useful to disable this probe: when I start a new
testnet, sometimes I am waiting for other participants to join for the
chain to kick off.

When recreating dailynet or mondaynet, I need to wait for other peers to
connect to me to download the most recent blocks: with a probe, that's
not possible.

Must be explicitly set to false, otherwise by default the probe will be
present. It can only be set at node type level. Since it's processed by
helm and not config-generator, I didn't see a way to make it
configurable at global or instance level.

I will need this to migrate teztnets to the new aws org without disruption.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,7 +203,7 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v1
         with:
-          version: v3.5.3
+          version: v3.8.1
 
       - name: Lint Helm Charts
         run: helm lint charts/*
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -38,7 +38,9 @@ spec:
         {{- include "tezos.container.logger"    $ | indent 8 }}
         {{- include "tezos.container.metrics"   $ | indent 8 }}
         {{- include "tezos.container.zerotier"  $ | indent 8 }}
+        {{- if not (eq  $.node_vals.readiness_probe false) }}
         {{- include "tezos.container.sidecar"   $ | indent 8 }}
+        {{- end }}
       initContainers:
         {{- include "tezos.init_container.config_init"         $ | indent 8 }}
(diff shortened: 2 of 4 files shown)


harryttd:
It is good to add this option. Just a few questions:

> when I start a new
testnet, sometimes I am waiting for other participants to join for the
chain to kick off.

If it doesn't have blocks and so is on 0, doesn't it return success?

> When recreating dailynet or mondaynet, I need to wait for other peers to
connect to me to download the most recent blocks: with a probe, that's
not possible.

Does this mean that the chain already exists and has blocks? So basically blocks stopped being produced and therefore the chain gets older and probe returns false? 

> Since it's processed by helm and not config-generator, I didn't see a way to make it
configurable at global or instance level.

Can set option in those places and then mount tezos-config configMap? Then probe looks at NODE_GLOBALS and it's own node options. Can return success msg saying why skipping validation.

harryttd (approved):
👍 👍

.github/workflows/ci.yml
```diff
@@ -203,7 +203,7 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v1
         with:
-          version: v3.5.3
+          version: v3.8.1
```
elric1, line 206: This is in one of my pull requests as well. I'll take it out, if we approve this one first.

charts/tezos/templates/nodes.yaml
```diff
@@ -38,7 +38,9 @@ spec:
         {{- include "tezos.container.logger"    $ | indent 8 }}
         {{- include "tezos.container.metrics"   $ | indent 8 }}
         {{- include "tezos.container.zerotier"  $ | indent 8 }}
+        {{- if not (eq  $.node_vals.readiness_probe false) }}
```
elric1: I think that this might be prettier if you put it in `_containers.tpl` just after:
```
```

PR #398, opened 2022-03-24
update flextesa in mkchain

nicolasochem:
The flextesa container used in mkchain no longer exists in the registry
(as I found out when changing laptop). Changing to a version that is
still there. I tested it and it works.


harryttd (approved):

smondet:
Just so you know, the ones we push there https://hub.docker.com/r/oxheadalpha/flextesa/tags are not supposed to be garbage collected AFAIU.

PR #399, opened 2022-03-24
update default octez to v12

nicolasochem:
Change version in test value files as well. Also fix a few outdated tests.

### DIFF
--- a/test/charts/mainnet.expect
+++ b/test/charts/mainnet.expect
@@ -28,11 +28,13 @@ data:
     }
   FULL_SNAPSHOT_URL: "https://mainnet.xtz-shots.io/full"
   ROLLING_SNAPSHOT_URL: "https://mainnet.xtz-shots.io/rolling"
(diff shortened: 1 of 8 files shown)


orcutt989 (approved):
Awesome!

test/charts/mainnet.expect
```diff
@@ -146,7 +151,7 @@ spec:
           readinessProbe:
             httpGet:
               path: /is_synced
-              port: 31732                                                
+              port: 31732
```
harryttd, line 154: theres some random white space being added in a bunch of places 🤷

nicolasochem: There is a helper script to re-generate the expected helm template. I ran it to update to v12 everywhere in the tests, then this happened. I believe whitespace was already there. Perhaps a tab vs space situation. This should be addressed separately.

elric1: I think that what's going on here is that a prior commit didn't update the expect files. Once we enforce this on every pull request we won't see this.

PR #400, opened 2022-03-25
Fix Helm error when a node class is set to null

harryttd:
If this was defined:
```yaml
nodes:
  tezos-node: null
```
the code removing node identities in the tezos-config configmap would fail. It assumes that a populated object is always set for a node class.

### DIFF
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -20,8 +20,9 @@ data:
 
   NODES: |
 {{/*
-  Remove "identity" field from nodes. Creates a deep copy because
-  "unset" modifies the object.
+  Don't add node classes without configuration to the configmap.
+  Also remove the secret "identity" field on node instances. Creates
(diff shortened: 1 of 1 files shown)


nicolasochem (approved):
Thank you for fixing this bug. I ran into it while trying to update mainnet-dev on oxheadinfra to 6.3.0.

mainnet-dev uses values.yaml merged from 2 yaml files: the mainnet yaml file, and a "diff" yaml file for everything that's different in dev.

mainnet was still on 6.0.0, but when I moved mainnet-dev to 6.3.0, I observed the bug fixed here. I was unwilling to move mainnet to 6.3.0.

A solution was to remove 

```
nodes:
    tezos-node: null
```

from mainnet values.yaml but then it would have affected mainnet as well.

Instead, I switched mainnet-dev to a submodule pointing to this PR, and `tezos-node: null` no longer gives me a pulumi error. Hence, this fix is solving my problem.

PR #403, opened 2022-04-06
switch to ithaca as default protocol for priv chains

nicolasochem:
* mkchain produces values.yaml for ithaca chain
* example in values.yaml comments is now ithaca
* remove any trace of the endorser in our code


harryttd:
Awesome :) Let's also update the Helm chart tests?

nicolasochem:
@harryttd  done

harryttd (approved):

PR #405, opened 2022-04-07
make hard-coded identity work when snapshot is being imported

nicolasochem:
Snapshot import wipes the data dir. But we may write an identity.json
file in it. This PR ensures that we are not deleting it after writing
it!

We also write it at every boot (because what if the user wants to change
it?)

We add the node id in the debug log for config-generator.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -425,12 +425,11 @@ def import_keys(all_accounts):
 def create_node_identity_json():
     identity_file_path = f"{DATA_DIR}/identity.json"
     path = Path(identity_file_path)
-    if path.exists() and path.stat().st_size > 0:
-        return
 
     # Manually create the data directory and identity.json, and give the
     # same dir/file permissions that tezos gives when it creates them.
     print("\nWriting identity.json file from the instance config")
+    print(f"Node id: {NODE_IDENTITIES.get(MY_POD_NAME)['peer_id']}")
 
     os.makedirs(DATA_DIR, 0o700, exist_ok=True)
     with open(
@@ -443,6 +442,7 @@ def create_node_identity_json():
     nogroup = getgrnam("nogroup").gr_gid
     chown(DATA_DIR, user=100, group=nogroup)
     chown(identity_file_path, user=100, group=nogroup)
+    print(f"Identity file written at {identity_file_path}")
 
 
 #
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -424,7 +424,6 @@ def import_keys(all_accounts):
 
 def create_node_identity_json():
     identity_file_path = f"{DATA_DIR}/identity.json"
-    path = Path(identity_file_path)
 
     # Manually create the data directory and identity.json, and give the
     # same dir/file permissions that tezos gives when it creates them.
--- a/utils/snapshot-downloader.sh
+++ b/utils/snapshot-downloader.sh
@@ -51,7 +51,8 @@ if [ -n "$snapshot_url" ] && [ -n "$tarball_url" ]; then
   echo "ERROR: Either only a snapshot or tarball url may be specified per Tezos node history mode."
 fi
 
-rm -rfv "$node_data_dir"
+# We are inporting a tarball or snapshot, first we remove everything execpt identity.json
+find $node_data_dir ! -name 'identity.json' -type f -exec rm -vf {} +
 mkdir -p "$node_data_dir"
 
 if [ -n "$snapshot_url" ]; then
--- a/utils/snapshot-downloader.sh
+++ b/utils/snapshot-downloader.sh
@@ -62,7 +62,6 @@ if [ -n "$snapshot_url" ]; then
 elif [ -n "$tarball_url" ]; then
   echo "Downloading and extracting tarball from $tarball_url"
   curl -LfsS "$tarball_url" | lz4 -d | tar -x -C "$data_dir"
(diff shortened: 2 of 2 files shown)


nicolasochem:
I verified it was working on oxheadinfra mainnet-dev by temporarily pointing a submod to this branch.

orcutt989 (approved):

utils/snapshot-downloader.sh
```diff
@@ -51,7 +51,8 @@ if [ -n "$snapshot_url" ] && [ -n "$tarball_url" ]; then
   echo "ERROR: Either only a snapshot or tarball url may be specified per Tezos node history mode."
 fi
 
-rm -rfv "$node_data_dir"
+# We are inporting a tarball or snapshot, first we remove everything execpt identity.json
+find $node_data_dir ! -name 'identity.json' -type f -exec rm -vf {} +
```
harryttd, line 55: I know this wasn't introduced in this PR, but we might want it to be addressed now....:

I recall asking at the time of the snapshot import/download functionality being built why the entire data dir should be wiped. I think a reason given was to make sure we really have a completely clean directory before importing a snapshot. And old state could mess with the import I guess.

It seems now though that giving the snapshot downloader script this kind of "ownership" over the tezos node data dir, that it can completely wipe its contents, could be excessive. And we've created a dependency on the behavior of the snapshot script to the data dir that is obscure and not clearly visible in other areas of the code. A bug was introduced because of this as we have just found.

What happens next time someone needs to write a file to the data dir and said person doesn't know that the dir gets wiped by the snapshot script if the script happens to run, and only if the script runs after the container within which the file write takes place? We didn't take into account initially that we would manually manipulate identities.json. Who knows what else in the future we might want to manipulate?

Perhaps the best thing to do is exit if data is found that shouldn't be there when we want to import a snapshot. What is the case of even having data that is around but not `context` (which is checked for at the beginning of the script)? How would that happen and is it reasonable to assume it really could happen? If it is a real edge case, shouldn't we treat it as an outlier and exit and make the user go in and handle it for themselves? This is a lot safer than wiping the entire dir automatically and keeping the obscure dependency of the dir to this script intact. 

At the bare minimum if it is not agreed to update this now (which i think we should, and exit and log a msg for the user), we leave a more elaborate comment explaining why we are doing a `rm -rf`. (btw the proposed change has  flag as only `-f` and not `-rf`)

Please think about and comment your thoughts!

harryttd (changes requested):

utils/config-generator.py
```diff
@@ -425,12 +425,11 @@ def import_keys(all_accounts):
 def create_node_identity_json():
     identity_file_path = f"{DATA_DIR}/identity.json"
     path = Path(identity_file_path)
```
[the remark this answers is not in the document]
nicolasochem: this variable `path` is used below, so we need it?

utils/snapshot-downloader.sh
nicolasochem: I felt it was ok to wipe the data dir because it's not supposed to have precious data in it. Your precious data is in the client folder. With xtz-shots having archive tarballs, it's never a concern to wipe this dir.

But you are making good points. What do we want to happen though? One way could be:

* if the context dir is present, assume snapshot has already been imported
* otherwise, check if there is any file at all. If there is, exit in error
* write identity.json afterwards

My concern is, what if there are files that we don't expect because of the filesystem? for example ext4 always has a `lost+found` file present. We may be introducing bugs in environments unknown to us.

harryttd: Once we check for `context`, why do we need to check for any other files? If we say that its existence is the determining factor that the node already has been syncing with the chain. If `context` isn't there, can we safely say there won't be any other blockchain data? Even if there are other folders, will the snapshot/tarball imports fail?

lost+found might be only in mounted dir `/var/tezos`.See [this](https://www.baeldung.com/linux/lost-found-directory#:~:text=The%20fsck%20utility%20creates%20the%20lost%2Bfound%20directory%20at%20the%20root%20of%20a%20volume.%20So%2C%20we%20can%20have%20multiple%20lost%2Bfound%20directories%20%E2%80%93%20one%20for%20each%20of%20the%20volumes). We could choose to rm it. Or not and just rely on `context` being there or not.

nicolasochem: > Once we check for context, why do we need to check for any other files?

This is what I thought you were requesting in your previous comment. I am fine with merging the PR a s is.

> If context isn't there, can we safely say there won't be any other blockchain data?

I am not sure.

>  Even if there are other folders, will the snapshot/tarball imports fail?

I am not sure, I don't think so.

harryttd: > Once we check for context, why do we need to check for any other files?
>> This is what I thought you were requesting in your previous comment. I am fine with merging the PR a s is.

I was suggesting that if we check for `context` and it isn't there, then we don't need to `find` and `rm -rf` as there shouldn't any other dirs. Maybe we should check for `store` as well, or rather check for the existence of any dir.
```shell
~ $ ls /var/tezos/node/data/
context        identity.json  lock           peers.json     store          version.json
```
I'm assuming `lock` is making sure only one process is running.

utils/config-generator.py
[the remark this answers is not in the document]
harryttd: It's not used after lines 428-429. A lambda function creates its own path arg

harryttd (approved):

PR #407, opened 2022-04-11
Snapshot downloader: support full history mode tarballs

picojulien:
Currently some history mode allows tarball import, other allow snapshot imports.
Furthermore it seems that tarball where not extracted at the right place. 

This MR 
- makes it more uniform by allowing tarball and snapshot imports for all history modes.
-  fix the data-dir where snapshot is downloaded
- remove peers.json file to prevent undesired connections to external network in case the tarball contains such file

### DIFF
--- a/utils/snapshot-downloader.sh
+++ b/utils/snapshot-downloader.sh
@@ -31,7 +31,8 @@ echo "My nodes history mode: '$my_nodes_history_mode'"
 snapshot_url=""
 tarball_url=""
 case "$my_nodes_history_mode" in
-  full)     snapshot_url="$FULL_SNAPSHOT_URL";;
+  full)     snapshot_url="$FULL_SNAPSHOT_URL"
+            tarball_url="$FULL_TARBALL_URL";;
 
   rolling)  snapshot_url="$ROLLING_SNAPSHOT_URL"
             tarball_url="$ROLLING_TARBALL_URL";;
--- a/charts/tezos/templates/_helpers.tpl
+++ b/charts/tezos/templates/_helpers.tpl
@@ -68,8 +68,12 @@
   Returns a string "true" or empty string which is falsey.
 */}}
 {{- define "tezos.shouldDownloadSnapshot" -}}
-  {{- if or (.Values.full_snapshot_url) (.Values.rolling_snapshot_url) (.Values.rolling_tarball_url) (.Values.archive_tarball_url) }}
-    {{- if and (.Values.rolling_tarball_url) (.Values.rolling_snapshot_url) }}
+  {{- if or (.Values.full_snapshot_url) (.Values.full_tarball_url)
+            (.Values.rolling_snapshot_url) (.Values.rolling_tarball_url)
+            (.Values.archive_tarball_url) }}
+    {{- if or (and (.Values.rolling_tarball_url) (.Values.rolling_snapshot_url))
+        (and (.Values.full_tarball_url) (.Values.full_snapshot_url))
+    }}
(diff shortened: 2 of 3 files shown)


harryttd:
Thank you for your PR @picojulien  :)

@elric1 has a PR modifying how the urls are structured. We'll need to converge yours and his. https://github.com/oxheadalpha/tezos-k8s/pull/394

Also, there isn't such a thing as an archive snapshot. Only full and rolling. In terms of tarballs, we don't currently offer full tars, but may do so in the future.

utils/snapshot-downloader.sh
```diff
@@ -60,8 +71,9 @@ if [ -n "$snapshot_url" ]; then
   curl -LfsS -o "$snapshot_file" "$snapshot_url"
 elif [ -n "$tarball_url" ]; then
   echo "Downloading and extracting tarball from $tarball_url"
-  curl -LfsS "$tarball_url" | lz4 -d | tar -x -C "$data_dir"
+  curl -LfsS "$tarball_url" | lz4 -d | tar -x -C "$node_data_dir"
```
nicolasochem, line 74: The current code definitely works for us. I suppose you are generating your own tarballs? @picojulien

utils/snapshot-downloader.sh
```diff
@@ -60,8 +71,9 @@ if [ -n "$snapshot_url" ]; then
...
   rm -fv "$node_data_dir/identity.json"
+  rm -fv "$node_data_dir/peers.json"
```
nicolasochem, line 76: tarballs from xtz-shots don't have this file. We actually changed this part yesterday. We are no longer deleting any file. We are assuming tarballs are sanitized and don't have empty files. (that's why this PR currently conflicts I guess)

utils/snapshot-downloader.sh
picojulien: Indeed for my experiment I am using my own tarball, and I realize, by reading you, that I should just have the `data` directory included in my tarball. I'll revert this.

utils/snapshot-downloader.sh
picojulien: Ok. I can amend my tarball building process.

utils/snapshot-downloader.sh
```diff
@@ -26,17 +26,28 @@ my_nodes_history_mode=$(< /etc/tezos/config.json jq -r "
 				 else .
 				 end")
 
+if [ "$my_nodes_history_mode" = "null" ]; then
```
nicolasochem, line 29: what is this for?

harryttd: it should already be handled via https://github.com/oxheadalpha/tezos-k8s/blob/d7f78e34adbe84087c16aa1cf0a6ed114b9a9e89/utils/config-generator.py#L570-L573

picojulien: I can't remember in which circumstances I needed that in my experiments, I removed it from this PR, as it might be orthogonal to it.

I'll open an other PR with better explanation  if I stumble on this case again. Apologies for the noise.

picojulien:
> Thank you for your PR @picojulien :)
> 
> Also, there isn't such a thing as an archive snapshot. Only full and rolling.

Ah, True ! I over-generalized this. 
 Archive mode could start with a full snapshot and reconstruct the contexts, but it's going much further than I intended to do in this PR.  
 I removed the snapshot part in archive mode.

harryttd (approved):
Thank you :) 

i changed the PR's title to reflect the new commits

PR #408, opened 2022-04-11
Add a "join_public_network" option

picojulien:
When spawning resilience testnet, we are in a setup where we don't define a private network, as we are using mainnet contexts, but we don't want to join public network.

This MR add an option in the value file to prevent connection to public network.


charts/tezos/values.yaml
```diff
@@ -243,6 +243,10 @@ expected_proof_of_work: 26
 node_config_network:
   chain_name: mainnet
 
+# If you dont want to join public network but need to set networks name to a
+# public one
```
nicolasochem, line 247:
```suggestion
# Setting node_config_network with a chain_name assumes the network is public.
# tezos-k8s will then configure the nodes to connect to boot-nodes for the selected chain.
# This may not always be desired. For example, you may want to run a mainnet replica
# in isolation with yes-nodes.
# Uncomment the line below to remove boot-node configuration from nodes on a public
# network. The nodes will not establish any connection outside of the cluster. 
```

nicolasochem (changes requested):
We do have a bit of "magic" behavior under the hood: when passing a network name, tezos-k8s assumes the network is public, runs `config init` to extract the hardcoded boot nodes, then adds them to the node configuration generated by config-generator.

This PR disables this behavior for the yes-nodes use case. I do not see a better way to do it.

I suggested a change for the values.yaml comment. Otherwise, LGTM.

utils/config-generator.py
```diff
@@ -52,8 +52,11 @@
 )
 
 # If there are no genesis params, this is a public chain.
-THIS_IS_A_PUBLIC_NET = True if not NETWORK_CONFIG.get("genesis") else False
-
+THIS_IS_A_PUBLIC_NET = True if ((not NETWORK_CONFIG.get("genesis"))) else False
```
harryttd, line 55: Line 55 should not contain any changes

utils/config-generator.py
```diff
@@ -52,8 +52,11 @@
...
+try:
+    JOIN_PUBLIC_NETWORK =  CHAIN_PARAMS.get("join_public_network")
+except:
+    JOIN_PUBLIC_NETWORK =  THIS_IS_A_PUBLIC_NET
```
harryttd, line 59: `.get` shouldn't ever throw an error. I don't think the try/except is necessary and it should be able to be simplified to:
```suggestion
JOIN_PUBLIC_NETWORK =  CHAIN_PARAMS.get("join_public_network", THIS_IS_A_PUBLIC_NET)
```

harryttd: I think we also need to do a more sophisticated check. Say `THIS_IS_A_PUBLIC_NET == False`, so we are spawning a new chain. But the user also set `JOIN_PUBLIC_NETWORK == True`. So it is a contradiction. We should probably throw an exception in this case.

charts/tezos/values.yaml
harryttd: @nicolasochem Perhaps the new flag should moved under `node_config_network` to keep the context of these fields together?

When defining a private chain, the format isn't exactly as a node expects a config.json to be because of the `activation_account_name`. I think it would be ok to put the new flag under `node_config_network` too. In addition to your comment, it should be made clear that `genesis` and  `join_public_network` are mutually exclusive, and there should be an error check in Helm to fail hard and fast if the conditions contradict. Also see comment in config-generator.py in regards to doing this validation in the backend.

I think an appropriate place for this validation is here, and it should fail if both `genesis` and `join_public_network` are true. In fact this check should return false if `!genesis and !join_public_network`. config-init container doesn't need to run if you want to run a public network node without actually joining the chain: https://github.com/oxheadalpha/tezos-k8s/blob/293abe0de4eaa3bf9c405b88e2ee95fecbf3643f/charts/tezos/templates/_helpers.tpl#L82-L94

harryttd (changes requested):
Thank you for this PR :) Please see my comments.

EDIT: I'm ok to take this over and to address my comments.

utils/config-generator.py
picojulien: Indeed. I amended my proposition.

picojulien: I'm not familiar with helm, so I don't know if we can push the check in the template, in order to fail earlier in the process ?

harryttd:
@picojulien It appears maintainers don't have access to your branch from your fork. So i can't add commits to this PR. Any way you can open up access on this PR so that i can push changes directly to your branch? Otherwise it makes it more complicated to add changes. I would likely need to make a new PR to get my changes in.

See: https://stackoverflow.com/a/57805943/6379084

charts/tezos/values.yaml
harryttd: Actually after more investigation, It seems we should still run config-init, as a url to a network configuration can be passed as the `chain_name` in values.yaml. We need config-init to fetch the json blob to produce the config.json. If the value passed was simply just the name of the network, then we wouldn't need config-init to run. We could just get the network name directly from `chain_name`.

nicolasochem (approved):
It works. I was able to launch jakartanet locally and restore from tarball. When passing the following values.yaml, I observe that the chain does not connect to anything after boot:

```
node_config_network:
  chain_name: "jakartanet"
  join_public_network: false

tezos_k8s_images:
  utils: localhost/tezos-k8s-utils:dev
rolling_snapshot_url: https://jakartanet.xtz-shots.io/rolling

images:
  octez: tezos/tezos:v13-release

nodes:
  rolling-node:
    runs:
      - octez_node
    instances:
      - is_bootstrap_node: true
        config:
          shell:
            history_mode: rolling
```

Thank you for your contribution.

PR #409, opened 2022-04-11
Separate deterministic generation of accounts and genesis definition

picojulien:
For resilience testnet, the structure of  the yaml file makes it necessary to activate the generation of deterministic data to declare bakers name and their dispatching amongst pod. 
We then  rewrite the wallet with a yes-wallet, who's keys are twisted for yes-crypto, that allows to bake "in the name of mainnet bakers" with patched nodes.
However we don't want to generate data for genesis as we are using a mainnet context.

This MR allows to split the `SHOULD_GENERATE_UNSAFE_DETERMINISTIC_DATA` into tow option that control which of accounts or genesis have to be generated.

### DIFF
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -6,7 +6,9 @@ data:
       "bootstrap_peers": {{ toJson .Values.bootstrap_peers }},
       "default_bootstrap_mutez": "{{ .Values.bootstrap_mutez }}",
       "expected-proof-of-work": {{ toJson .Values.expected_proof_of_work }},
-
+      "should_generate_unsafe_deterministic_data": {{ toJson .Values.should_generate_unsafe_deterministic_data }},
+      "should_generate_unsafe_deterministic_genesis_data": {{ toJson .Values.should_generate_unsafe_deterministic_genesis_data }},
+      "should_generate_unsafe_deterministic_account_data": {{ toJson .Values.should_generate_unsafe_deterministic_account_data }},
       "should_generate_unsafe_deterministic_data": {{ toJson .Values.should_generate_unsafe_deterministic_data }},
       "network": {{ mustToPrettyJson .Values.node_config_network | indent 8 | trim }},
       "protocol_activation": {{ .Values.activation | mustToPrettyJson | indent 8 | trim }}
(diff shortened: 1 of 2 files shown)


nicolasochem (commented):
I understand you are using the deterministic account generation to give good names to your accounts, but you then rewrite the wallets with yes wallets (how do you do that btw?)

I suppose that your mainnet replica yaml does not have a `genesis` block?

But rather mainnet config like:

```
node_config_network:
  chain_name: mainnet
```

In this case, it may be better to modify `fill_in_missing_genesis_block` to do nothing if `node_config_network` is like above?

picojulien:
> I understand you are using the deterministic account generation to give good names to your accounts, but you then rewrite the wallets with yes wallets (how do you do that btw?)

Yes. 
The yes-wallet is embeded in the docker image we are using, and we patched  `charts/tezos/scripts/tezos-node.sh` to place it at the proper  place/

> I suppose that your mainnet replica yaml does not have a `genesis` block?
> 
> But rather mainnet config like:
> 
> ```
> node_config_network:
>   chain_name: mainnet
> ```
> 
> In this case, it may be better to modify `fill_in_missing_genesis_block` to do nothing if `node_config_network` is like above?

Now that I understand a bit better this aprt of the code, I think that what  I want is to have a `do_not_verify_bakers_account`  option that would prevent the config_generator to check if aliases of baker accounts have been declared in the account list.
 
My yaml file looks like this
```
nodes:
 ...
    instances:
    - bake_using_accounts:
      - baker_0
      is_bootstrap_node: true
    - bake_using_accounts:
      - baker_1
      is_bootstrap_node: true
    - bake_using_accounts:
      - baker_2
      is_bootstrap_node: true
 ```
The file has no accounts section, and  ` baker_0`  is  an alias in the yes wallet. 
I'll try to do this.

**Comment by elric1** [comment]
It's probably easier to just elide the account names entirely by doing this:
```
nodes:
  eu:
    instances:
      - is_bootstrap_node: true
      - is_bootstrap_node: true
      - {}
      - {}
      - {}
```
The above will make 5 nodes the first two of which are bootstrap nodes, the next three of which aren't. The account names will be the node name: eu-N where N \elem {0,1,2,3,4}.

PR #411, opened 2022-04-13
Snapshot bugfix

### LINKED ISSUE #414: Service account variable is hardcoded in some places
In values.yaml we allow sa name to be set:
https://github.com/oxheadalpha/tezos-k8s/blob/d7f78e34adbe84087c16aa1cf0a6ed114b9a9e89/charts/snapshotEngine/values.yaml#L8

In both places here it is hardcoded:

https://github.com/oxheadalpha/tezos-k8s/blob/d7f78e34adbe84087c16aa1cf0a6ed114b9a9e89/snapshotEngine/mainJob.yaml#L12

https://github.com/oxheadalpha/tezos-k8s/blob/d7f78e34adbe84087c16aa1cf0a6ed114b9a9e89/snapshotEngine/snapshotMakerJob.yaml#L12

### LINKED ISSUE #417: Snapshot Engine: volumeSnapshotClassName is hardcoded
https://github.com/oxheadalpha/tezos-k8s/blob/d7f78e34adbe84087c16aa1cf0a6ed114b9a9e89/snapshotEngine/createVolumeSnapshot.yaml#L9

Right now the volume snapshot class is manually created in the cluster. Pulumi needs to be creating it, because it is cluster scoped and we only want to create it once. So we don’t want it in tezos-k8s which would want to create it for every Helm deployment we do to the cluster.
So Pulumi needs to create it and it would give it a name. That name needs to be passed into values.y which would then have to get it into the VolumeSnapshot resource defined in createVolumeSnapshot.yaml
See [https://aws.amazon.com/blogs/containers/using-ebs-snapshots-for-persistent-storage-with-your-eks-cluster/#:~:text=Step%204%3A[…]ass%20and%20Snapshot](https://aws.amazon.com/blogs/containers/using-ebs-snapshots-for-persistent-storage-with-your-eks-cluster/#:~:text=Step%204%3A%20create%20a%20Volume%20Snapshot%20Class%20and%20Snapshot)

orcutt989:
Currently deployed and working.

This fixes some bugs found with the snapshot-engine.

1. Fix wrong image name in values.yaml on dockerhub
2. Pass in markdown for chain site as `snapshotMarkdownTemplateUrl` value.
3. Fix #414 
4. Fix #417


nicolasochem (approved):

nicolasochem (approved):
I didn't try, but it looks like everything is here. Thanks

charts/snapshotEngine/templates/snapshot-scheduler.yaml
```diff
@@ -38,6 +38,11 @@ spec:
                 configMapKeyRef:
                   name: snapshot-configmap
                   key: IMAGE_NAME
+            - name: SERVICE_ACCOUNT
+              valueFrom:
+                configMapKeyRef:
+                  name: snapshot-configmap
+                  key: SERVICE_ACCOUNT
```
harryttd, line 45: Why in order to retrieve the value for service_account and image_name env vars does it need to go to the configmap? The configmap itself gets the values from values.yaml. Why don't you just get the values directly from there?

snapshotEngine/mainJob.yaml
```diff
@@ -220,6 +220,11 @@ spec:
                 configMapKeyRef:
                   name: snapshot-configmap
                   key: NAMESPACE
+            - name: SNAPSHOT_MARKDOWN_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: snapshot-configmap
+                  key: SNAPSHOT_MARKDOWN_URL
```
harryttd, line 227: SNAPSHOT_MARKDOWN_TEMPLATE_URL

snapshotEngine/snapshotMakerJob.yaml
```diff
@@ -34,4 +34,9 @@ spec:
                 configMapKeyRef:
                   name: snapshot-configmap
                   key: TEZOS_IMAGE
+            - name: SERVICE_ACCOUNT
+              valueFrom:
+                configMapKeyRef:
+                  name: snapshot-configmap
+                  key: SERVICE_ACCOUNT
```
harryttd, line 41: why is this needed?

harryttd (changes requested):

charts/snapshotEngine/templates/snapshot-scheduler.yaml
orcutt989: The service_account is used by yq for other templates.  We cannot pass these values into the yaml templates otherwise until the yaml files are turned into part of the helm chart.

snapshotEngine/snapshotMakerJob.yaml
orcutt989: Usd by YQ to pass into yaml manifests that are not in the helm template.

snapshotEngine/mainJob.yaml
harryttd: you only fixed it in one spot.....

orcutt989: Thanks got it now!

charts/snapshotEngine/templates/snapshot-scheduler.yaml
harryttd: not sure i follow. We are in the snapshot-scheduler which gets deployed directly by helm. Hence it can reference values just like it is currently doing so for namespace and history mode.

orcutt989: Yes you're right thanks.

snapshotEngine/snapshotMakerJob.yaml
harryttd: I don't see it being used in the snapshot-maker.sh script

orcutt989: https://github.com/oxheadalpha/tezos-k8s/blob/snapshot-bugfix/snapshotEngine/snapshot-maker.sh#L175-L176

harryttd: Great, thanks for pointing it out.

just fyi, instead of picking each individual key off of a configmap, you can pass the whole configmap to set all the env vars within at once

harryttd (approved):
thank you. please squash and delete branch

PR #415, opened 2022-04-13
add explanatory comments to snapshotEngine's values.yaml

nicolasochem:
also replace `v12.2` with `v12-release` to follow tezos-k8s convention


charts/snapshotEngine/values.yaml
```diff
@@ -15,4 +39,11 @@ nodes:
...
     target_volume: var-volume
 
+# snapshotEngine generates static webpages to accompany the generated
+# artifacts. They are rendered by jekyll.
+# Below specify the URL for the source markdown file template to render for
+# every individual artifact.
+# It will be rendered based on metadata generated by the snapshot engine.
+# See an example at:
+#   https://github.com/oxheadalpha/xtz-shots-website/blob/master/snapshot.md
 chainWebsiteMarkdown: ""
```
harryttd, line 49: Thank you for the comment. 
Do we agree from slack convo to name this something like `jekyllSnapshotTemplateUrl`?

charts/snapshotEngine/values.yaml
```diff
@@ -1,12 +1,36 @@
...
   snapshotEngine: tezos-k8s-snapshot-engine:dev
 
+# the tezos version used to run `tezos-node snapshot import/export`
 images:
-  octez: tezos/tezos:v12.2
+  octez: tezos/tezos:v12-release
 
+# snapshotEngine containers interact with the kubernetes control
+# plane to create volume snapshots. This requires a special role.
+# Specify this role below.
```
harryttd, line 10:
```suggestion
# snapshotEngine containers interact with the kubernetes control
# plane to create volume snapshots. This requires a special IAM role
# so k8s can create the snapshots and volumes in the cloud.
# Specify this role below.
```
good? I believe that's how it works.

charts/snapshotEngine/values.yaml
```diff
@@ -1,12 +1,36 @@
...
 iam_role_arn: ""
 service_account: snapshot-engine-sa
 
+# snapshotEngine chart does not run tezos nodes.
+# It searches for separately defined tezos nodes and targets them
+# by taking CSI volumeSnapshots of their volume, then restoring them
+# into new pvcs and running jobs against them.
+# Note that these tezos nodes should be in the same namespace.
+# snapshotEngine expects one archive node and one rolling node in order
+# to generate all artifacts.
```
harryttd, line 20: Thank you for the great comment.

This sentence is true, although i think it might imply it is necessary to deploy both rolling and archive nodes. But a user can choose to deploy one or the other. They are mutually exclusive.

charts/snapshotEngine/values.yaml
```diff
@@ -1,12 +1,36 @@
...
+#   <NODE_CLASS>:
+#     history_mode: <HISTORY_MODE>
+#     target_volume: <TARGET_VOLUME>
+# where:
+# NODE_CLASS is the node class as defined in tezos-k8s "tezos" chart.
+# HISTORY_MODE is archive or rolling
+# TARGET_VOLUME is the name of the pod's volume containing tezos data dir.
+#
+# The default configuration below will be looking for tezos-nodes with
+# class `snapshot-archive-node` and `snapshot-rolling-node`.
```
harryttd, line 33: good?
```suggestion
# The default configuration below will be looking for tezos nodes defined in the 
# tezos chart with class names `snapshot-archive-node` and `snapshot-rolling-node`. 
# If you'd like to name them differently, make sure to match the names in both the 
# tezos chart and here in the snapshot engine chart. 
# Also, the tezos-k8s chart names the node data volume 'var-volume' by default.
```

harryttd (changes requested):
Thank you for the great comments

charts/snapshotEngine/values.yaml
nicolasochem: I just wrote a ticket for this after thinking about it https://github.com/oxheadalpha/tezos-k8s/issues/416

charts/snapshotEngine/values.yaml
nicolasochem: You mean they are not mutually exclusive? Right now in oxheadinfra we are deploying both.

We have not really tested deploying one without the other, so I think we are ok?

harryttd: Ah sorry , yes meaning you don't need both. It fine with deploying only one. If a user wants a website, they can write it to just handle one history mode.

Logically it would work. For each class defined in yaml, we create 2 new deployments, the warmer and the scheduler. The diff node classes don't relate at all to one another. They get separate deployments that target only their respective node with matching history mode.

harryttd (approved):

orcutt989 (approved):
Thank you

PR #421, opened 2022-04-13
remove hardcoded volume class name, fix #417

### LINKED ISSUE #417: Snapshot Engine: volumeSnapshotClassName is hardcoded

nicolasochem:
from kube doc

A volume snapshot can request a particular class by specifying the name
of a VolumeSnapshotClass using the attribute volumeSnapshotClassName. If
nothing is set, then the default class is used if available.

https://kubernetes.io/docs/concepts/storage/volume-snapshots/

### DIFF
--- a/snapshotEngine/createVolumeSnapshot.yaml
+++ b/snapshotEngine/createVolumeSnapshot.yaml
@@ -6,6 +6,5 @@ metadata:
   labels:
     history_mode: rolling
 spec:
-  volumeSnapshotClassName: test-snapclass
   source:
     persistentVolumeClaimName: ""


nicolasochem:
It won't work, we will have to create the snapclass outside of helm and pass its name as a value.

nicolasochem:
It won't work, we will have to create the snapclass outside of helm and pass its name as a value.

Issue #395, opened 2022-03-22
docusaurus pipeline


nicolasochem:
Set up a `docs` folder in this repo and a docusaurus build pipeline that deploys the contents of the docs folder into a static webpage hosted by github.

The goal is to break out the README file into more detailed documentation split in sections (private chain, private node etc)

orcutt989:
https://github.com/oxheadalpha/tezos-k8s/pull/426

PR #426, opened 2022-04-18
Docusaurus documentation site

orcutt989:
This PR creates the initial build pipeline and site structure for a docusaurus-based documentation site for tezos-k8s.

Fix https://github.com/oxheadalpha/tezos-k8s/issues/395

http://www.oxheadalpha.com/tezos-k8s/


nicolasochem (changes requested):
Lets' not put the default blog post and images there? Let's just have one blank page named "tezos-k8s documentation?"

Also do we want to deploy on release or on push? I think deploying on release makes more sense, but we should deploy on push for now just to see that the pipeline works.

.github/workflows/docusaurus.yml
```diff
@@ -0,0 +1,47 @@
+name: docusaurus
+
+on:
+  push:
```
nicolasochem, line 4: Can we only deploy documentation during releases? We don't want unreleased features to be there.

docusaurus/src/pages/index.md
```diff
@@ -0,0 +1,7 @@
+# Tezos K8s
```
nicolasochem, line 1: Can this page move to the top level of the repo under `docs/` folder? So it's easier for contributors to find.

nicolasochem (dismissed):

.github/workflows/docusaurus.yml
orcutt989: Done

docusaurus/src/pages/index.md
orcutt989: Done

nicolasochem (dismissed):
thank you

nicolasochem (commented):
there are 2 nojekyll files, normal?

nicolasochem (approved):
thanks!

PR #427, opened 2022-04-21
Ability to adjust or disable the snapshot storage init time

orcutt989:
# Overview
This PR provides the ability for a user to set the time limit for the storage initialization container. 

# Background
This container starts a headless Tezos node and initializes the storage restored from an EBS snapshot. This is necessary to account for missing head blocks that frequently occur.  This is done for both rolling and archive tarballs, and Tezos snapshot workflows.

# Details
We've added the `restoredStorageInitTime`, `disableStorageInitTime` of which default to `2m` and `false` respectively.

2m has been plenty of time for testnets and mainnet in the past, but now it seems that mainnet might need more time to init storage as it has been a few days since a successful job.

Upon investigation it seems that all the jobs are killed during the head reconstruction, so it must mean that we need more time to let the head rebuild on mainnet.

In the future it may be necessary to let the job run without being killed to determine a proper init time, so the ability to disable the time limit has also bee implemented.

### DIFF
--- a/charts/snapshotEngine/values.yaml
+++ b/charts/snapshotEngine/values.yaml
@@ -70,4 +70,4 @@ volumeSnapClass: ""
 restoredStorageInitTime: 2m
 
 # We also provide the ability to disable the time limit for debugging purposes.
-disableStorageInitTime: false
+disableStorageInitTime: "false"
--- a/charts/snapshotEngine/values.yaml
+++ b/charts/snapshotEngine/values.yaml
@@ -68,3 +68,6 @@ volumeSnapClass: ""
 # grows in size over time.
 # This time format is in unix sleep time format ex. 1s, 2m, 3h, 4d.
 restoredStorageInitTime: 2m
+
+# We also provide the ability to disable the time limit for debugging purposes.
+disableStorageInitTime: false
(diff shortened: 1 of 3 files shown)


charts/snapshotEngine/values.yaml
```diff
@@ -56,4 +56,18 @@ snapshotMarkdownTemplateUrl: ""
...
+# of the KILL -9 that occurs during the EBS snapshot process.  
+# Tezos also does not exit on error, but rather just hangs stdout on a random error. 
+# Therefore we limit this initialization process and pitch the job in a reasonable amount of time to account for these
+# random errors. It may be necessary to increase this time for different changes, or as a particular chain
+# grows in size over time.
+# This time format is in unix sleep time format ex. 1s, 2m, 3h, 4d.
+restoredStorageInitTime: 2m
+
+# We also provide the ability to disable the time limit for debugging purposes.
+disableStorageInitTime: "false"
```
nicolasochem, line 73: But if you set it to `true` it will fail? Isn't there a problem where helm interprets `true` as a boolean and the value never gets to the bash script?

orcutt989: Yes, but that's why I added `| quote` to the value now so that it works. https://github.com/oxheadalpha/tezos-k8s/pull/427/files#diff-cb1335f37197d6090d8312cef92528eb49f9c17ee997cc7b9cfecf1a991f2f17R9

nicolasochem (approved):
thank you. I didn't try but I know it's deployed already and works.

PR #428, opened 2022-04-22
Add pull policy for tezos-signer

novalis:


nicolasochem (approved):
This was caught by using mkchain + private docker tezos build combination, where the signer would fail with a pull error.
Thank you for your contribution.

PR #430, opened 2022-04-22
Document how to use a custom tezos build

novalis:


nicolasochem (approved):
thank you

README.md
```diff
@@ -423,6 +424,21 @@ You can optionally spin up an RPC authentication backend allowing trusted users
...
+
+Within the minikube environment, run
+`make docker-image` from inside a clone of the
+`[tezos](https://gitlab.com/tezos/tezos)` repository.  This will
+create a docker image called `tezos_build:latest`.  Inside
+`$CHAIN_NAME_values.yaml`, change the `images` section to:
+
+```yaml
+images:
+  octez: tezos:latest
```
harryttd, line 437: If the image created is called tezos_build, shouldn't the image be named the same thing here? @novalis

elric1: I have found that it's much faster to make the image outside the minikube environment and copy it in for some reason. Might be worth a try to see if it is still the case.

harryttd: @novalis I think what @elric1 is pointing out makes sense. I don't think minikube is the place to be cloning repositories to then build images. And i do believe doing it all internally is slower especially with the tezos gitlab repo being about 3.3 GB. It makes sense that the build should be happening externally. Minikube can then use, say docker desktop, to contain the images and pull from there. Or you can import the image into minikube.

elric1: Strangely, I even found _that_ to be slow. I didn't dig into why it was, though, so it might just be my setup.
By _that_, I mean building externally with:
```
:; eval $(minikube -p minikube docker-env)
```
What I found to be faster was building completely externally and then doing something like:
```
:; docker image save ... | ( eval $(minikube -p minikube docker-env); docker image load )
```

novalis: What I did while testing was exactly @harryttd's suggestion: I just used my normal clone of tezos/tezos, and ran `make docker-image` after running `eval $(minikube docker-env)`.  It didn't seem particularly slow, although I'm not sure what to compare it against.

I don't understand @elric1's suggestion: what args to `docker image save` would be equivalent to the ones that the Makefile uses?  And why should that be faster?   Also, I think we should not suggest anything other `make`, since the Makefile is kept up-to-date by the tezos/tezos folks, and we don't want to be suggesting things that are not supported by upstream.

harryttd: I think my confusion with the current comment in the readme is:
```
Within the minikube environment, run
`make docker-image` from inside a clone of the
`[tezos](https://gitlab.com/tezos/tezos)` repository.
```
This makes it sound like one should clone the repo into minikube docker land and then build it.

elric1: There are also missing intructions. You should include the URL:
https://tezos.gitlab.io/introduction/howtoget.html#setting-up-the-development-environment-from-scratch
to help install the prerequisites.

novalis: How's this version?

nicolasochem (approved):

README.md
elric1: @novalis: sorry, I wasn't precise. What I meant to say is that I have found in the past that:
```
:; make docker-image
```
is much faster than:
```
:; ( eval $(minikube -p minikube docker-env); make docker-image )
```
In the first case, you are using the docker environment directly on your host and in the second case, you are using the docker environment that is running via minikube within the docker environment on your host. Because `make docker-image` just runs `docker build ...` and that builds things within a docker image, when you do it after selecting the innner docker environment, all of the I/O will be much slower as it is sent to minikube which redirects it to its containing docker which then sends it to the OS. And thus, I found that building it without using `eval $(minikube docker-env)` and then copying the image in was much faster for me.

PR #429, opened 2022-04-22
Draft: Configure initial delegation

novalis:
Should wait for https://gitlab.com/tezos/tezos/-/merge_requests/5071 to merge.

PR #435, opened 2022-04-27
Support for tezos-reward-distributor

novalis:


nicolasochem (approved):
A useful chart to deploy a public baker. Thanks.

This deploys TRD as a cronjob.

By default, it is configured to connect to an archive node in the same namespace deployed with tezos-k8s with default values.yaml from mkchain.

I was able to test and check that it loads properly. Not quite that it distributes rewards yet, because this is dependant on an octez change to configure delegations at genesis (see #429).

It's ok though. It's a standalone chart that can be modified later based on needs. Meanwhile, it does not affect other parts of the repo. Thank you for your contribution.

PR #436, opened 2022-05-02
Make remote signer secure

nicolasochem:
Right now, when configuring tezos-k8s with a remote signer, you can
still sign transfers from the node shell, or by connecting to the signer
endpoint. This PR makes the signer only allow bakes, endorses and
preendorses, by restricting it to the magic bytes 11-12-13. It's not
much (you can still shell in to the signer and transfer whatever you
like) but it's a start.

I am also enforcing high watermark while we are at it.

I tested in a private chain and was able to verify that it still bakes
and endorses, but transfers from the node shell fail with:

```
magic byte 0x03 not allowed
Fatal error:
transfer simulation failed
~ $
```


orcutt989 (approved):

harryttd (approved):

nicolasochem:
thanks for review, we just deployed on testnets and it's working. merging.

PR #437, opened 2022-05-03
templates with secrets as volumemounts instead of env vars

orcutt989:


nicolasochem (approved):
I didn't try locally, but this is currently deployed on teztnets, and it works for activation, signer and node (3 types of pods that use config-gen.py) so I think we are ok.

Thank you for bringing the expect yamls up to date, I should have done that in my previous PR. Soon the Ci will enforce this.

nicolasochem:
Merging, so @elric1 can work on the env var PR. @elric1 you can review this as you merge the code into your branch :)

PR #433, opened 2022-04-26
Secrets as VolumeMounts instead of EnvVars

nicolasochem:
Sorry about the extra long branch name, I have been using the "create branch from issue" feature, and unfortunately this issue had a long title. We will try to keep issue titles shorter in the future.

This PR switches all EnvVars to VolumeMounts as a security precaution.

Secrets in Kubernetes as EnvVars is a CIS Benchmark violation and would not pass an audit. Other reasons include protection from logging leaks, and also possibly being enumerated in `/proc/` and seen with a `ps`.

https://stackoverflow.com/questions/51365355/kubernetes-secrets-volumes-vs-environment-variables/53367504#53367504

https://www.tenable.com/audits/items/CIS_Kubernetes_v1.20_v1.0.0_Level_2_Master.audit:98de3da69271994afb6211cf86ae4c6b

Tested with Minikube.

PR #438, opened 2022-05-04
snapshot engine: fix #416 (remove the hardcoding) and #433 (add url in metadata)

### LINKED ISSUE #416: remove `snapshot-website-base` in snapshot engine; download it instead
the _config.yml in xtz-shots-website and snapshot-website-base is identical except for one line.

Instead, let's have these 3 config in values.yaml:

```
jekyllRemoteThemeRepository: oxheadalpha/xtz-shots-website
jekyllConfig: https://raw.githubusercontent.com/oxheadalpha/xtz-shots-website/master/_config.yml
snapshotMarkdownTemplate: https://raw.githubusercontent.com/oxheadalpha/xtz-shots-website/master/snapshot.md
```

Then the website builder job:
* downloads `jekyllConfig`
* adds `remote_theme: oxheadalpha/xtz-shots-website` to the downloaded config based on the `jekyllRemoteThemeRepository` param
* uses this config to build the remote theme

nicolasochem:
This is addressing the little bit of leftover work from the snapshotEngine merge:

snapshotEngine should be totally separate from xtz-shots since anyone can deploy their own. So, all the xtz-shots specific stuff should be hosted on the xtz-shots-website repo, and the snapshotEngine should grab whatever it needs to build the static page.

This is done in this PR. We no longer have a _config.yml in the tezos-k8s repository: we now grab it from xtz-shots-website and append the remote theme config to it with a heredoc.... not super slick but it does the job.

The advantage is that we no longer have to modify config.yml here whenever there is some change in the website metadata (for example, the list of networks on the top right of the page used to say "hangzhounet" despite us having removed it from xtz-shots-website. no longer).

The only leftover references to xtz-shots website are the commented out values in values.yaml (because of course, anyone who would want to deploy a snapshot website should still look into ours to understand how it works)

The only leftover files are Gemfile and Gemfile.lock which I think is ok. 

I am also adding full URL to the metadata, this way operators can simply 1. query the metadata 2. parse the url and and 3. download the snapshot


snapshotEngine/mainJob.yaml
```diff
@@ -231,16 +231,31 @@ spec:
...
+              valueFrom:
+                configMapKeyRef:
+                  name: snapshot-configmap
+                  key: JEKYLL_CONFIG
+            - name: JEKYLL_REMOTE_THEME_REPOSITORY
               valueFrom:
                 configMapKeyRef:
                   name: snapshot-configmap
-                  key: SNAPSHOT_MARKDOWN_TEMPLATE_URL
+                  key: JEKYLL_REMOTE_THEME_REPOSITORY
```
harryttd, line 258: As a suggestion: The values from the configMap can be picked and ranged over instead of having to repeat these blocks of env vars 5 times. Additionally if we don't care about polluting the env vars of the container, the entire config map can be mounted at once with `envFrom`.

snapshotEngine/zip-and-upload.sh
```diff
@@ -158,6 +159,7 @@ if [ "${HISTORY_MODE}" = archive ]; then
             "block_timestamp": $BLOCK_TIMESTAMP,
             "filename": $ARCHIVE_TARBALL_FILENAME,
             "sha256": $SHA256,
+            "url": "https://${NETWORK}.${SNAPSHOT_WEBSITE_DOMAIN_NAME}/${ARCHIVE_TARBALL_FILENAME}",
```
harryttd, line 162: The url can be defined as a variable by prefixing `https` with `S3_BUCKET` and used across this file, appending the filename each time.

snapshotEngine/zip-and-upload.sh
```diff
@@ -548,11 +554,21 @@ jq -n \
...
+# Grab Jekyll config
+curl -o _config.yml "${JEKYLL_CONFIG}"
+
+# Add remote theme to config
+cat <<EOF >> _config.yml
+remote_theme: ${JEKYLL_REMOTE_THEME_REPOSITORY}
+plugins:
+- jekyll-remote-theme
+EOF
+
```
harryttd, line 571: I can't leave a comment on line [541](https://github.com/oxheadalpha/tezos-k8s/pull/438/files#diff-54485a3da425f8ab97a77059aa9b26d885e1f0c46e95c0f1ad8938625f9d9ffeR541) as it isn't part of the diff, but `config` should be removed from the comment.

harryttd (changes requested):
Great PR :)

- I'm assuming this is tested works?
- Do we want to make deploying a website optional? That might be too much work right now. I'm not sure.
- A user may not want a custom domain for their site and might just want to use a raw url to the bucket. Do we want to handle this now?

snapshotEngine/mainJob.yaml
harryttd: Never mind picking and ranging bec this file is not helm templated. But the configmap is able to be mounted with envFrom.

nicolasochem: what do you mean by "picking and ranging"?

harryttd: something along these lines:
```yaml
env:
{{ rang k, v := pick .Values "a" "b" }}
 - name: k
   value: v
 {{ end }}
```

snapshotEngine/zip-and-upload.sh
harryttd: sorry i missed line 575 also. has xtz-shots.io.

harryttd (approved):
i left one more comment but skip if you want

nicolasochem:
> * I'm assuming this is tested works?

Now yes. It's deployed on ithaca-shots. Here is the new metadata:

```
nochem@fedora ~/workspace/pulumi-polkadot/charts/plenty-bridge () $ curl -Ls https://ithacanet.xtz-shots.io/rolling-tarball-metadata
{
  "block_hash": "BLiT1qiY15DVM3p23KKNDGrQH3aMS5VdDVipY9dDXjk3QUkUz8Z",
  "block_height": "492980",
  "block_timestamp": "2022-05-06T02:29:30Z",
  "filename": "tezos-ithacanet-rolling-tarball-492980.lz4",
  "url": "https://ithacanet.xtz-shots.io/tezos-ithacanet-rolling-tarball-492980.lz4",
  "sha256": "911222c63243baadca77900df1af650c9bde7a25ffdbb86f518de51fdc67ea53",
  "filesize_bytes": "92478089",
  "filesize": "88M",
  "tezos_version": "6e2037c9 (2022-04-07 12:44:15 +0200) (12.3)",
  "chain_name": "ithacanet",
  "history_mode": "rolling",
  "artifact_type": "tarball"
}
```

> * Do we want to make deploying a website optional? That might be too much work right now. I'm not sure.
> * A user may not want a custom domain for their site and might just want to use a raw url to the bucket. Do we want to handle this now?

No, let's defer until we are actually trying to deploy without a website. Thanks for approval.

PR #440, opened 2022-05-07
Fixes for new networks

orcutt989:


nicolasochem (approved):
This fixes a problem deploying new networks on xtz-shots when no data is present yet. It also solves an issue introduced last-minute in #438

It's currently deployed on all 3 xtz-shots namespaces and appears working.

thank you

PR #441, opened 2022-05-09
repair zerotier

nicolasochem:
I found zerotier not working when testing #393

After debugging, it looks like it was not waiting for the ipv4 address
to be present before attempting to parse it.

There is logic to wait for the ipv4 addres to be present, by conuting
the number of inet strings in the output of ip a. It was meant to match
a 2, but it matches a 1, for some reason. Setting the value to 2 fixes
it.

Also bump zerotier version to 1.6 (latest is 1.8 but it requires rust
dependencies, so 1.6 sounds good enough).

### DIFF
--- a/zerotier/Dockerfile
+++ b/zerotier/Dockerfile
@@ -1,37 +1,37 @@
 FROM alpine
 
-ENV ZEROTIER_VERSION=1.4.6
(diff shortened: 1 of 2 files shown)


harryttd (approved):
THank you for fixing. Weird that it used to work

PR #444, opened 2022-05-11
docusaurus initial

nicolasochem:


orcutt989 (approved):
Lgtm thanks. Let's see how it looks?

docs/README.md
```diff
@@ -0,0 +1,19 @@
...
+```
+cd docs/
+./values_to_doc.sh
+```
+
+To render locally:
+
+```
+npm install
+./node_modules/.bin/docusaurus serve --build
```
harryttd, line 18: this can just be `npm run build && npm start` instead of needing to reference the binary directly. You can see all of the npm scripts in package.json.

docs/Private-Node.md
```diff
@@ -0,0 +1,61 @@
...
+Running this results in:
+
+- Creating a Helm [release](https://helm.sh/docs/intro/using_helm/#three-big-concepts) named tezos-mainnet for your k8s cluster.
+- k8s will spin up one regular (i.e. non-baking node) which will download and import a mainnet snapshot. This will take a few minutes.
+- Once the snapshot step is done, your node will be bootstrapped and syncing with mainnet!
+
+You can find your node in the oxheadalpha namespace with some status information using `kubectl`.
+
+```shell
+kubectl -n oxheadalpha get pods -l appType=tezos-node
```
harryttd, line 22: Once we are copying from README, we should probably change `appType=octez-node` everywhere.

docs/Prerequisites.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+This allows you to run Docker commands inside of minikube. For example: `docker images` to view the images that minikube has.
+
+If you want to unset your shell from using minikube's docker daemon:
+
+```shell
+eval $(minikube docker-env -u)
+```
+
```
harryttd, line 86: I believe this is missing this section https://github.com/oxheadalpha/tezos-k8s#tezos-k8s-helm-chart

docs/04-other-helm-charts.md
```diff
@@ -0,0 +1,9 @@
+# Other Helm charts
+
+The Tezos-k8s project also maintains several other helm charts:
+
+* **pyrometer**: a chart to deploy the [Pyrometer Tezos monitoring tool](https://gitlab.com/tezos-kiln/pyrometer)
+* **rpc-auth**: a vending-machine to dispense Tezos RPC access
+* **tezos-faucet**: a faucet to dispense pre-activated accounts (backend only)
+* **snapshotEngine**: an utility to generate Tezos snapshots from running nodes, using Kubernetes CSI driver
+* **tezos-reward-distributor**: a chart to deploy [TRD](https://tezos-reward-distributor-organization.github.io/tezos-reward-distributor/)
```
harryttd, line 9: Should we link to them?

docs/values_to_doc.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
harryttd, line 1: The file is not committed with file owner execute perms. I could not run this script without giving execute perms first. I believe the CI would fail.

docs/sidebars.js
```diff
@@ -1,31 +1,34 @@
...
+  tezosK8sSidebar: [
+    'index',
+    {
+      type: 'category',
+      label: 'Common use cases',
+      collapsed: false,
+      items: [
+        'Private-Node',
+        'Private_Chain',
+        'Tezos-RPC-Service']
```
harryttd, line 17: Shouldn't the prereqs appear first? The user is directed from the index page to the use cases. But the use cases don't just describe what they are but also how to implement them. A user can't do that without seeing the prereqs first which only appear later.

On the index page itself, and example is given how to just do a simple install of the helm chart to get a mainnet node. Should we mention prereqs somewhere there first?

README.md
```diff
@@ -37,92 +37,6 @@ Follow the prerequisites step first. Then you can jump to either [joining mainne
 
 NOTE: You do not need to clone this repository! All necessary components will be installed.
 
-## Prerequisites
```
harryttd, line 40: Why are we removing the prereqs?

charts/tezos/values.yaml
```diff
@@ -30,14 +30,18 @@ chain_initiator_job:
...
-# local testnets.
-should_generate_unsafe_deterministic_data: false
+# The `accounts` object of values.yaml defines Tezos accounts used in the chart.
+# By default no account is configured:
+accounts: {}
+#
+# `accounts` is a map where keys are account aliases and values are maps of
+# fields `key`, `is_bootstrap_baker_account`, `bootstrap_balance` and `signer_url`.
+#
+# The key field can be set to a public or private key. For a bootstrap baker,
```
harryttd, line 42: The `key`

charts/tezos/values.yaml
```diff
@@ -48,104 +52,117 @@ should_generate_unsafe_deterministic_data: false
...
+#
+# Use `nodes` to configure the Tezos nodes running in your chart.
+#
+# `nodes` is a dictionary where each key/value pair defines a statefulset and a
+# number of instances thereof. The name (key) defines the name of the
+# statefulset and will be the base of the pod names. The instances are defined
+# as a list because their names are simply -N appended to the statefulsetname.
+# Said names are typically kebab case.
 #
+# https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/
```
harryttd, line 104: Why this link here?

charts/tezos/values.yaml
```diff
@@ -30,14 +30,18 @@ chain_initiator_job:
 # account is not explicitly set below.
 bootstrap_mutez: "4000000000000"
 
-# # ACCOUNTS
+# # Accounts
```
harryttd, line 33: Maybe keep section headers as upper case? Easier to see separation between sections where they start and end. Especially when using magic shell script to delineate and parse pieces out.

harryttd:
Lol i was in the middle of a review

harryttd (changes requested):

charts/tezos/values.yaml
nicolasochem: The issue is then they don't look nice in the doc (unless the shell script makes them lowercase, but I think we are ok)

README.md
nicolasochem: They are in docusaurus now, README now should be doc on how to develop on tezos-k8s, not use it.

docs/sidebars.js
```diff
@@ -1,31 +1,34 @@
...
-  tutorialSidebar: [
+  tezosK8sSidebar: [
+    'index',
+    {
+      type: 'category',
+      label: 'Common use cases',
+      collapsed: false,
+      items: [
+        'Private-Node',
+        'Private_Chain',
```
harryttd, line 16: Page url names are inconsistent with `-` vs `_`

README.md
harryttd: Shouldn't we be removing the entire readme then? It is all about how to use it. Or do you mean setting it up to be used?

(I don't see why not still we can't have the prereqs. Plenty of projects describe what the prereqs are before actually saying how to use the project itself.)

Shouldn't at the minimum there be a  link to docs now?

nicolasochem: Yes, I think we should move most of the README to the docs and just leave the development section in the README (and a link to the documentation of course)

docs/04-other-helm-charts.md
harryttd: I meant linking to each chart in our charts directory

nicolasochem:
@harryttd thanks for review. I fixed in subsequent PR https://github.com/oxheadalpha/tezos-k8s/pull/462

charts/tezos/values.yaml
nicolasochem: It was there as an annotation to a different comment. Now removed.

docs/index.md
```diff
@@ -1,7 +1,35 @@
...
+* a Tezos node
+* a Tezos RPC service
+* a Tezos baker (coming soon)
+* a Tezos private chain
+
+on your laptop or in a cloud platform such as AWS, Digitalocean or GCP.
+
+Tezos-k8s is actively maintained by the team at [Oxhead Alpha](https://oxheadalpha.com), a Tezos core development company.
+
+## Quick start
```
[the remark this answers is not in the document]
nicolasochem: Can we leave it? The audience for such section is a person who already has access to helm and a cluster, and just wants to install tezos.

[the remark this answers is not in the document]
harryttd: Can we say we assume you have followed the prereqs (link here)? And to get started etc.... There is no direction from the index.md page how to actually get started from scratch. The first thing a user sees is how to deploy a mainnet node as well as that being the next page as well.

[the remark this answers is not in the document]
nicolasochem: done

PR #452, opened 2022-05-21
make unsafe disclaimer removal work with mkchain

nicolasochem:
by default helm changes Y into true. But we want the env var to be
actually Y, so we have to overload the PyYAML class to allow quoted
strings.


harryttd (approved):
https://yaml.org/type/bool.html

PR #453, opened 2022-05-21
Deprecate 'type' field in accounts

nicolasochem:
This field was previously used by config_generator, but @elric1 changed
the code to infer the key type from the string.

We are deprecating this field and throwing a helm message when it's
found. It is pruned by helm and no longer passed to config_generator.

In a later release, we can fail the chart install when it is found.

I tried:

* mkchain, no params
* mkchain, 2 bakers
* mkchain with old "type" field: gives me a warning, ignores content of
  field
* mkchain then delete private keys; observe they are reconstructed
  deterministically when unsafe setting in values.yaml is set to true
* pass an actual public key as non bootstrap baker


harryttd (approved):
This is great. Thank you. In the future i think we can get rid of 'type' altogether and just use pytezos props and methods to determine the type of key.

PR #455, opened 2022-05-24
Revert "chain-initiator.sh: piping to head loses error status"

elric1:
This reverts commit 9e9517c02a2c0c05cec80dabfdb4f0d3f2701653.

Turns out that because pipefail is set, that this commit is incorrect.
I did issues w.r.t. this in testing, but must have been on a branch or
perhaps an older image with a different /bin/sh as I can't reproduce
them any more.


nicolasochem (approved):
thank you

PR #456, opened 2022-05-24
v14 prep: set per-block-votes from values.yaml

nicolasochem:
This gives a tezos-k8s baker the ability to set per-block votes in the `protocols` section of values.yaml. It is also made mandatory.

This fits the tzip recommendation of not giving a default choice to the user.

I added some logic in the baker script to support v13 (v13 requires the cli flag even when the vote json is present)

~Helm will give a warning when the vote is not set.~

~After kathmandu goes live, v13 of octez will no longer be supported so we can turn this warning into an error, and remove v13 backward compatibility.~

Helm gives an error when the vote is not set. This breaks compatibility with Ithaca (which is ok)

Verified on 13.0 and 14.0. Now that 14.0 is out, it is time to merge this and do a release. **I also set the default octez version to v14 everywhere.**

For now, voting is global. Custom votes per baker will be supported later, by allowing an override of `protocols` at statefulset instance level.


nicolasochem:
Update: it works on octez 14.0 and kathmandunet is using this PR as submodule, [here is the config](https://github.com/oxheadalpha/teztnets/blob/main/kathmandunet/values.yaml#L243-L249)

We can merge once 14.0 is final (this will break compat with 13.0)

orcutt989 (approved):
Cool! Thanks! 🔥

charts/tezos/values.yaml
```diff
@@ -314,8 +314,17 @@ protocols:
   ## You may want to run two or more if you are expecting an on-chain upgrade.
   ## Be mindful that running more protocols will start more containers in your
   ## pods which may consume more memory.
+  ## Syntax:
+  ## * command: the suffix of the tezos-baker binary
```
harryttd, line 318: prefix?

charts/tezos/values.yaml
```diff
@@ -314,8 +314,17 @@ protocols:
...
+  ## * vote: a map of per-block votes. These are protocol specific.
+  ##   "on" and "off" must be between quotes.
```
harryttd, line 320: what about pass?

in the alpha proto below it is set as pass.

charts/tezos/values.yaml
```diff
@@ -314,8 +314,17 @@ protocols:
...
+  ## Note that we are not providing default votes since every baker needs
+  ## to make an explicit educated choice on every toggle.
```
harryttd, line 322: Does this mean that a one command deployment of the chart won't work anymore?

charts/tezos/scripts/baker.sh
```diff
@@ -9,12 +9,23 @@ BAKER_EXTRA_ARGS_FROM_ENV=${BAKER_EXTRA_ARGS}
...
+- command: ${proto_command}
+  vote:
+    liquidity_baking_toggle_vote: "on"
+EOF
+  exit 1
+fi
+extra_args="--votefile ${per_block_vote_file}"
+
+tezos_version=$(tezos-client --version | sed -e 's/ //g')
+if [[ "$tezos_version" == *"13"* ]]; then
```
harryttd, line 26: the client command also outputs a hash and date. Isn't it possible for them to contain "13"? Or being that they don't at the current time, we will eventually remove this code and it doesn't matter. What about v14 tho? I

harryttd (changes requested):

charts/tezos/values.yaml
nicolasochem: Yes it will, because the command deploys a node, not a baker, which works out of the box.

test/charts/private-chain.expect.yaml
```diff
@@ -194,6 +194,15 @@ metadata:
   name: tezos-config
   namespace: testing
 ---
+# Source: tezos-chain/templates/configs.yaml
```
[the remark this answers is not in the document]
nicolasochem: done

charts/tezos/values.yaml
nicolasochem: suffix like `tezos-baking-013-PtJakart`

charts/tezos/values.yaml
nicolasochem: pass needs not be between quotes. the problem with "on" and "off" is that yaml (or helm) interprets it as true or false.

charts/tezos/scripts/baker.sh
nicolasochem: fixed by matching to 13.0. I don't expect any new v13 release at this point, so it works.

charts/tezos/values.yaml
harryttd: ah my bad

charts/tezos/templates/nodes.yaml
```diff
@@ -59,6 +59,9 @@ spec:
         - name: tezos-accounts
           secret:
             secretName: tezos-secret
+        - name: per-block-votes
```
harryttd, line 62: Shouldn't this also need the hasPerBlockVotes check? If the configmap doesn't exist it will fail to mount.

nicolasochem: My bad. Fixed and I tested both baker and regular node on minikube.

charts/tezos/templates/NOTES.txt
```diff
@@ -1,5 +1,10 @@
...
+{{- range $k, $v := .Values.accounts -}}
+  {{- if $v.type -}}
 DEPRECATION WARNING: You specified a type for account {{ $k }}. This field is now ignored, the type of key is set automatically.
-  {{- end }}
-{{- end }}
+  {{- end -}}
+{{- end -}}
+{{- range $p := .Values.protocols -}}
+  {{- if (not $p.vote) }}
+WARNING: you did not specify the liquidity baking toggle vote in "protocols" for protocol "{{ $p.command }}", your baker will not start.
```
harryttd, line 8: Sorry I didn't notice earlier. Should we fail instead of warning being the baker will fail anyways? Just exit early in helm?

nicolasochem: OK. Another issue is that when you are not a baker, the warning/error should not be thrown. So I need to rework this a bit.

harryttd: Perhaps the check for bakers could be with Roland's code for creating the baker container. 

'''yaml

{{- define "tezos.container.bakers" }}
  {{- if has "baker" $.node_vals.runs }}
    {{- $node_vals_images := $.node_vals.images | default dict }}
    {{- range .Values.protocols }}
      {{- $_ := set $ "command_in_tpl" .command }}
      {{- include "tezos.generic_container" (dict "root" $
                                                  "name" (print "baker-"
                                                          (lower .command))
                                                  "type"        "baker"
                                                  "image"       "octez"
      ) | nindent 0 }}
    {{- end }}
  {{- end }}
{{- end }}
'''

nicolasochem: Done. As a consequence, Ithaca w/o flag no longer works, but I think that's fine, it's an old protocol.

harryttd (approved):
Awesome :) I'm assuming works as a submod in OI. Did not test the latest commits myself

PR #468, opened 2022-07-06
preserve ordering of bootstrap parameters in mkchain

nicolasochem:
Bootstrap parameters for mkchain are saved in a [file](https://github.com/oxheadalpha/tezos-k8s/blob/master/mkchain/tqchain/parameters.yaml), but PyYAML reorders them so they appear alphabetically in the values.yaml file outputted by mkchain. I like to have my chain bootstrap parameters sorted always the same way.

We pass a parameter to the pyyaml dumper to tell it to preserve ordering.


harryttd:
This is a good idea :) Something i've wanted

We could keep the official lib and just add in the `sort_keys=False` arg to the dump function: https://stackoverflow.com/a/55171433/6379084

nicolasochem:
> We could keep the official lib and just add in the `sort_keys=False` arg to the dump function: https://stackoverflow.com/a/55171433/6379084

Ooh thanks. How did I miss this?

nicolasochem:
@harryttd your suggestion has been applied, please review

PR #470, opened 2022-07-06
Generate a genesis block when omitted from config

nicolasochem:
The chain id is deterministically generated from the genesis block hash,
and it is important that every new public chain has a different one.
On teztnets, we accidentally gave the same block hash to half-baked
kathmandunet and jakartanet.

To prevent this in the future, we add a tezos-k8s feature that generates
a genesis block hash when absent. It is derived from chain_name.

@elric1 had already written some code for that but it was gated by a
flag `should_generate_unsafe_deterministic_data`. The purpose of that
flag is to autogenerate deterministic well-known accounts for private
simulations. It also generates the genesis block hash, but that's not
unsafe per se. So I moved this code out of the flag. When the chain_name
is provided, it's using it as the seed, otherwise it uses "foo" as
before.

Mkchain now outputs a values.yaml without genesis block. It will be
generated with the method introduced here.

### DIFF
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -66,7 +66,9 @@ if not THIS_IS_A_PUBLIC_NET and JOIN_PUBLIC_NETWORK:
 def main():
     all_accounts = ACCOUNTS
 
-    fill_in_missing_genesis_block()
+    if not THIS_IS_A_PUBLIC_NET:
+        fill_in_missing_genesis_block()
+
     if SHOULD_GENERATE_UNSAFE_DETERMINISTIC_DATA:
         all_accounts = fill_in_missing_accounts()
         fill_in_missing_keys(all_accounts)
@@ -148,14 +150,12 @@ def main():
 # If NETWORK_CONFIG["genesis"]["block"] hasn't been specified, we generate a
 # deterministic one.
 def fill_in_missing_genesis_block():
-    genesis_config = NETWORK_CONFIG.get("genesis")
-    if genesis_config and not genesis_config.get("block"):
+    genesis_config = NETWORK_CONFIG["genesis"]
+    if not genesis_config.get("block"):
         print("Deterministically generating missing genesis_block")
-        if NETWORK_CONFIG.get("chain_name"):
-            seed = NETWORK_CONFIG["chain_name"]
-        else:
-            seed = "foo"
-
+        if not NETWORK_CONFIG.get("chain_name"):
+            raise Exception("Genesis config is missing 'chain_name'.")
+        seed = NETWORK_CONFIG["chain_name"]
         gbk = blake2b(seed.encode(), digest_size=32).digest()
         gbk_b58 = b58encode_check(b"\x01\x34" + gbk).decode("utf-8")
         genesis_config["block"] = gbk_b58


mkchain/tqchain/mkchain.py
```diff
@@ -24,7 +24,7 @@ def quoted_scalar(dumper, data):  # a representer to force quotations on scalars
 MyDumper.add_representer(QuotedString, quoted_scalar)
 # end https://stackoverflow.com/a/52424865/207209
 
-from tqchain.keys import gen_key, get_genesis_vanity_chain_id, set_use_docker
```
harryttd, line 27: we can remove the function from the file now that we aren't using it anymore

utils/config-generator.py
```diff
@@ -66,8 +66,8 @@
 def main():
     all_accounts = ACCOUNTS
 
+    fill_in_missing_genesis_block()
```
harryttd, line 69: As a general thing, I don't think this should be run unless we are dealing with a private chain. As it isn't relevant otherwise.

Also, i do believe bec it is being run even on a public network, it will break. As the function tries to get `genesis` off of the network config, and it won't exist on a public net. (we shouldn't fix this by defaulting it to an empty object)

harryttd (changes requested):

utils/config-generator.py
nicolasochem: Running it on a public chain is the point here. I want to eliminate one step of teztnets creation, where I have to manually generate a genesis block. Instead, it will be autogenerated. This is relevant at activation.

If I give you my values with the Genesis block absent, you will regenerate it.

I need to convert this code into typescript so the output of teztnets also contains this genesis block. Otherwise people won't be able to run `--network https://teztnets.xyz/networkname`

If the network is well known and hardcoded into Octez, then it does not matter. The Genesis hash will be hardcoded too.

nicolasochem: But, you are right that the code fails when the network is hardcoded i.e. mainnet. Let me fix this.

nicolasochem: I tried with the following values and it worked, considering it resolved:

```
node_config_network:
  chain_name: "https://teztnets.xyz/ghostnet"

tezos_k8s_images:
  utils: localhost/tezos-k8s-utils:dev

rolling_tarball_url: "https://ghostnet.xtz-shots.io/rolling-tarball"
rolling_snapshot_url: null
```

harryttd (changes requested):

utils/config-generator.py
nicolasochem: With the last iteration of my code, `fill_in_genesis_block_hash` is only run if `THIS_IS_A_PUBLIC_NET` is false. (I don't like this denomination anymore after our discussion, I think it should be `THIS_IS_A_CUSTOM_NET` but anyway)

This function fail with an explanatory error msg when the chain name is not found instead of defaulting to `foo`.

I don't know why this default is here to begin with, perhaps octez fills in a chain name when none is provided. Anyway, to the best of my knowledge this default is not useful to anyone now).

harryttd (approved):

Issue #446, opened 2022-05-13
Snapshot Engine: Serve all snapshots' metadata on a single and fixed endpoint


pasqu4le:
The snapshot engine currently produces [JSON metadata](https://github.com/oxheadalpha/tezos-k8s/blob/912041f605af47c812b07197218ccf5c345c37a6/charts/snapshotEngine/README.md#json) inlcuding:
> a `base.json` containing a list of all artifacts created

On `xtz-shots.io` however it seems that these are served on different URLs, depending on the network, e.g.:
- https://mainnet.xtz-shots.io/base.json
- https://ithacanet.xtz-shots.io/base.json
- etc.

but not https://xtz-shots.io/base.json

Especially in the hope of standardizing this metadata IMO it would be great if:
1. metadata for all snapshots (regardless of network) was served on a single endpoint
2. the endpoint used a fixed name, more memorable and descriptive than `base.json`

P.s. big fan of seeing this served on xtz-shots.io, we are currently trying to adopt it in [`tezos-packaging`](https://github.com/serokell/tezos-packaging).

harryttd:
@itkach Wanted to push the prio on this

PR #475, opened 2022-07-20
Snapshot Engine: One metadata for all artifacts

orcutt989:
Fixes https://github.com/oxheadalpha/tezos-k8s/issues/446

This PR provides all subdomains on xtz-shots to the zip-and-upload-job where they are fed into a python script that constructs a list of all current artifacts across all networks and provides it at `xtz-shots.io/snapshots.json`.

TODO:

- [ ] Right now `snapshots.json` is served at `subdomain.xtz-shots.io/snapshots.json` so every subdomain has the same `snapshots.json`, this needs to be moved up to `xtz-shots.io/snapshots.json`.

### DIFF
--- a/snapshotEngine/updateAvailableSnapshotMetadata.py
+++ b/snapshotEngine/updateAvailableSnapshotMetadata.py
@@ -5,34 +5,26 @@ import urllib.request
 import boto3
 import sys
 
-allSubDomains = os.environ['ALL_SUBDOMAINS']
+allSubDomains = os.environ['ALL_SUBDOMAINS'].split(",")
 
-# Iterate over each subDomain/base.json and append to snapshot.json
-filename = "snapshots.json"
-with open (filename, "w") as json_file:
-  json.dump("[]",json_file)
+filename="snapshots.json"
 
-if not exists(filename):
-  sys.exit('##### ERROR '+filename+' does not exist! #####')
+# Write empty top-level array to initialize json
+json_object = []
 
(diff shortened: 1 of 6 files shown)


snapshotEngine/updateAvailableSnapshotMetadata.py
```diff
@@ -0,0 +1,30 @@
...
+    for entry in data:
+      json_object.append(entry)
+
+# Write to file
+with open (filename, 'w') as json_file:
+  json_string = json.dumps(json_object, indent=4)
+  json_file.write(json_string)
+
+# Upload to S3 bucket
+s3 = boto3.resource('s3')
```
nicolasochem, line 28: It's probably better to use the aws cli in the parent bash script to upload the files created by this script.

This python script will be extended to do jinja rendering of the snapshot jekyll page. Then we will have multiple jekyll pages: one per network, then one per artifact. This means, some pages will sometimes be deleted from the jekyll bucket.

Knowing that, it's probably better for this script to write all files to an output dir (which it is already doing for the metadata file), and use `aws s3 sync` on the parent bash script, with remote deletion enabled. It  seems that [you can't do that in python](https://www.reddit.com/r/awspotentially/comments/ppwoch/synchronise_files_to_s3_with_boto3/)

nicolasochem (changes requested):

orcutt989:
Moved to https://github.com/oxheadalpha/tezos-k8s/pull/483

PR #477, opened 2022-08-04
podman compat in development mode

nicolasochem:
Podman does not accept docker image tags in the format image:tag as
local, it requires `localhost/` when the image was built locally.

This small change saves me time because a `mkchain` default yaml works
on my machine again.

I believe this will not break the development workflow for docker users,
please confirm.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -12,8 +12,8 @@ images:
   tezedge: tezedge/tezedge:v1.6.8
 # Images that are part of the tezos-k8s repo go here with 'dev' tag
 tezos_k8s_images:
-  utils: tezos-k8s-utils:dev
-  zerotier: tezos-k8s-zerotier:dev
+  utils: localhost/tezos-k8s-utils:dev
+  zerotier: localhost/tezos-k8s-zerotier:dev
 
 ## Properties that are templated for some k8s resources. There are container
 ## scripts that will look up some of these values. They should not be modified.
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -9,15 +9,15 @@ deployments:
 
 images:
   zerotier:
-    image: tezos-k8s-zerotier
+    image: localhost/tezos-k8s-zerotier
     dockerfile: ./zerotier/Dockerfile
     context: ./zerotier
   utils:
-    image: tezos-k8s-utils
+    image: localhost/tezos-k8s-utils
     dockerfile: ./utils/Dockerfile
     context: ./utils
   faucet:
-    image: tezos-k8s-faucet
+    image: localhost/tezos-k8s-faucet
     dockerfile: ./faucet/Dockerfile
     context: ./faucet
 
--- a/test/charts/mainnet.expect.yaml
+++ b/test/charts/mainnet.expect.yaml
@@ -158,7 +158,7 @@ spec:
               path: /is_synced
               port: 31732                                                        
         - name: sidecar
-          image: "tezos-k8s-utils:dev"
+          image: "localhost/tezos-k8s-utils:dev"
           imagePullPolicy: IfNotPresent
           args:
             - sidecar
@@ -233,7 +233,7 @@ spec:
             - mountPath: /var/tezos
               name: var-volume                
         - name: config-generator
-          image: "tezos-k8s-utils:dev"
+          image: "localhost/tezos-k8s-utils:dev"
           imagePullPolicy: IfNotPresent
           args:
             - config-generator
@@ -263,7 +263,7 @@ spec:
             - mountPath: /etc/secret-volume
               name: tezos-accounts        
         - name: snapshot-downloader
-          image: "tezos-k8s-utils:dev"
+          image: "localhost/tezos-k8s-utils:dev"
           imagePullPolicy: IfNotPresent
           args:
             - snapshot-downloader
--- a/test/charts/mainnet2.expect.yaml
+++ b/test/charts/mainnet2.expect.yaml
@@ -189,7 +189,7 @@ spec:
               path: /is_synced
               port: 31732                        
         - name: logger
-          image: "tezos-k8s-utils:dev"
+          image: "localhost/tezos-k8s-utils:dev"
           imagePullPolicy: IfNotPresent
           args:
             - logger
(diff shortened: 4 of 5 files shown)


harryttd:
By any chance have you seen this thread? It looks like there are some workarounds for building images that work both with Pman and docker. See a the following bunch of comments. It seems that podman is not a drop in replacement for docker. https://github.com/containers/buildah/issues/1034#issuecomment-805809794
Also, Perhaps in mkchain you can check for the existence of podman as opposed to docker. IMO this seems like a more viable solution to me.

nicolasochem:
> By any chance have you seen this thread? It looks like there are some workarounds for building images that work both with Pman and docker. See a the following bunch of comments. It seems that podman is not a drop in replacement for docker. [containers/buildah#1034 (comment)](https://github.com/containers/buildah/issues/1034#issuecomment-805809794) Also, Perhaps in mkchain you can check for the existence of podman as opposed to docker. IMO this seems like a more viable solution to me.

From the link you posted:

> Yes Docker hard codes the assumption of all images come from docker.io if a registry is not specified. Where as Podman hard codes localhost to indicate no registry is involved.

I could indeed tag `docker.io` or detect whether docker/podman in mkchain but you have not indicated why this PR is not viable. The point is, we should hardcode to localhost to indicate that no registry is involved (image is deposited to minikube's local storage). Localhost is compatible with both docker and podman.

harryttd:
```shell
│ Events:                                                                                                                 │
│   Type     Reason     Age                From               Message                                                     │
│   ----     ------     ----               ----               -------                                                     │
│   Normal   Scheduled  25s                default-scheduler  Successfully assigned ithacanet/activate-job-42nfz to minik │
│ ube                                                                                                                     │
│   Normal   BackOff    21s (x2 over 22s)  kubelet            Back-off pulling image "localhost/tezos-k8s-utils:dev"      │
│   Warning  Failed     21s (x2 over 22s)  kubelet            Error: ImagePullBackOff                                     │
│   Normal   Pulling    8s (x2 over 23s)   kubelet            Pulling image "localhost/tezos-k8s-utils:dev"               │
│   Warning  Failed     8s (x2 over 23s)   kubelet            Failed to pull image "localhost/tezos-k8s-utils:dev": rpc e │
│ rror: code = Unknown desc = Error response from daemon: Get "http://localhost/v2/": dial tcp 127.0.0.1:80: connect: con │
│ nection refused                                                                                                         │
│   Warning  Failed     8s (x2 over 23s)   kubelet            Error: ErrImagePull                                         │
│
```

nicolasochem:
Do you have the container locally on your minikube with the correct tag?

harryttd:
Some of the comments in that thread seem to indicate that you can modify some podman config that can tell it where to look for images. https://github.com/containers/buildah/issues/1034#issuecomment-1128361292

harryttd:
>  Do you have the container locally on your minikube with the correct tag?

Tag or image name? 

My images look like this:
```
oxheadalpha/tezos-k8s-utils         latest                 f49290981ffb   8 weeks ago     207MB
tezos-k8s-utils                    dev
```

nicolasochem:
> > Do you have the container locally on your minikube with the correct tag?
> 
> Tag or image name?
> 
> My images look like this:
> 
> ```
> oxheadalpha/tezos-k8s-utils         latest                 f49290981ffb   8 weeks ago     207MB
> tezos-k8s-utils                    dev
> ```

Yes, so as part of this PR, you are expected to always tag your images `localhost` whenever you build them. `cd utils && docker build -t localhost:tezos-k8s-utils/dev .`

The aliasing link you provided does not work for localhost.

harryttd:
I'm probably failing to fully understand your difficulty and solution so thinking out loud here.

I would assume most users run docker. How would this effect remote chart users? The default images in the chart will now have localhost prepended. That is what will be looked up from dockerhub. The ci deploys images as `oxheadalpha/tezos-k8s-${{ matrix.container }}`. 
Cloning the repo would require explicit instructions on how to build images and explain why for the users who use docker.

Due to my assumption of most users running docker, why not check for podman existence? Or how about adding a new flag to mkchain? I don't see why this is so bad `mkchain --podman` or similar as your default command? All you need is one flag.

nicolasochem:
> I'm probably failing to fully understand your difficulty and solution so thinking out loud here.
> 
> I would assume most users run docker. How would this effect remote chart users? The default images in the chart will now have localhost prepended. That is what will be looked up from dockerhub. The ci deploys images as `oxheadalpha/tezos-k8s-${{ matrix.container }}`.

Remote chart users use released charts with images as `oxheadalpha/tezos-k8s-${{ matrix.container }}`. They are not affected.

The default image in the **cloned** chart now has localhost prepended.

> Cloning the repo would require explicit instructions on how to build images and explain why for the users who use docker.

It's already the case? When you clone the repo, according to DEVELOPMENT.md you are presumed to use devspace (although that's not necessary, you can just go and build the images by hand). I just changed the devspace config to add localhost to the images it generates.

How are you building images locally? Are you using devspace?


> 
> Due to my assumption of most users running docker, why not check for podman existence? Or how about adding a new flag to mkchain? I don't see why this is so bad `mkchain --podman` or similar as your default command? All you need is one flag.

But why not merge this instead? It requires you to build the image with name `localhost/tezos-k8s-whatever` instead of just `tezos-k8s-whatever`. If you skip this step, you will have an error as you pasted above. If you skip this step on master branch, docker will try to pull an image named `tezos-k8s-whatever` without org name from docker hub, which will fail for a confusing reason. It's minor, and I would not care if I was not using podman, but my fix looks better to me than a flag.

harryttd:
The reason i disagree with this PR is bec it seems to me like it is trying to solve a problem for a minority of users who are using podman. This is based on my assumption that most people will be using docker and won't understand why localhost should be included in the image name. I'm not using devspace at the current time.

If devspace is used you can use a profile to specify podman vs docker. Otherwise just build manually with prepended localhost. And mkchain can check for the existence of podman or any other container client without needing a flag automatically.

If you insist on implementing this way i will accept :) 
The ci would need to be updated. I tested the current shell script and it includes localhost in the image name.
 In addition i don't think it is immediately obvious to users that they need to include localhost in the image name. They may not look in the devspace file. I would think leaving a comment in the development readme would be beneficial.

nicolasochem:
Thanks for considering. I'll continue editing my charts by hand for now. I think the world is not ready to move away from docker :wink: so people will indeed be confused by this `localhost`.

harryttd:
you can always update mkchain code if you want :)

PR #478, opened 2022-08-16
Pyrometer: put ingress in chart

nicolasochem:
Since the beginning of tezos-k8s we have always deployed ingresses
outside of our charts.

Is it the right approach? Maybe not: the default helm chart created with
`helm init` has a pre-configured ingress template file.

This ingress template file is sufficient for our use cases: we can pass
ingress annotations and TLS configuration with values.yaml.

Therefore, we add an option to define the ingress with helm directly.
This allows to remove an ingress definition from the teztnets and
oxheadinfra code bases, simplifying it.

We should consider doing that for Tezos RPC as well.

### DIFF
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -2,6 +2,9 @@
 config: {}
 images:
   pyrometer: registry.gitlab.com/tezos-kiln/pyrometer:latest
+service:
+  type: ClusterIP
+  port: 80
 ingress:
   enabled: false
   className: ""
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -11,14 +11,7 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  hosts:
-    - host: chart-example.local
-      paths:
-        - path: /
-          pathType: Prefix
-          backend:
-            serviceName: "pyrometer"
-            portName: 'http'
+  host: chart-example.local
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -2,9 +2,6 @@
 config: {}
 images:
   pyrometer: registry.gitlab.com/tezos-kiln/pyrometer:latest
-service:
-  type: ClusterIP
-  port: 80
 ingress:
   enabled: false
   className: ""
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -15,7 +15,10 @@ ingress:
     - host: chart-example.local
       paths:
         - path: /
-          pathType: ImplementationSpecific
+          pathType: Prefix
+          backend:
+            serviceName: "pyrometer"
+            portName: 'http'
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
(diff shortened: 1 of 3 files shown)


charts/pyrometer/values.yaml
```diff
@@ -2,3 +2,21 @@
...
+  enabled: false
+  className: ""
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: chart-example.local
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
```
harryttd, line 18: should we remove?

nicolasochem: `helm init` gives me similar content in the default values.yaml so I think it's fine.

nicolasochem: Ingress is [not enabled in the default values.yaml](https://github.com/oxheadalpha/tezos-k8s/pull/478/files#diff-e650dc3182d15c3ecd5ef92d4ab8c2a708be3c54263da825d5f6cb6f3dba1fccR9). If you enable it, you certainly want to mess around with it, or you already know what you are doing, in which case you will switch to the path type of your choice.

Are you just saying the line should be removed?

harryttd: Actually it should be configured to map to the pyrometer application. The ingress exposes the rule which will hit pyrometer backend. `host` is optional. https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-rules But it can be populated with a domain name that you create and `tls` can be populated with certs.
Right now it is mapping to places that don't exist. The ingress if it can be tested in minikube should be tested there. Or just deploy it now to OI and test it there if it hasn't been tested already. 
```ts
  const rpcDomain = `rpc.${namespace}.aryeh.${oxheadHostedZoneName}`
  const rpcCert = new aws.acm.Certificate(rpcDomain, {
    validationMethod: "DNS",
    domainName: rpcDomain,
  })

  const rpcCertValidationRecord = new aws.route53.Record(
    `${rpcDomain}-certValidation`,
    {
      name: rpcCert.domainValidationOptions[0].resourceRecordName,
      records: [rpcCert.domainValidationOptions[0].resourceRecordValue],
      ttl: 300,
      type: rpcCert.domainValidationOptions[0].resourceRecordType,
      zoneId: aryehOxheadZone.id,
    }
  )

  const certValidation = new aws.acm.CertificateValidation(rpcDomain, {
    certificateArn: rpcCert.arn,
    validationRecordFqdns: [rpcCertValidationRecord.fqdn],
  })

  // Wait for all resources of both charts to be ready
  const ingressDependencies = pulumi
    .all([albController.chart.ready, externalDns.chart.ready])
    .apply(([a, b]) => [...a, ...b])

  const rpcIngress = new tezos.aws.RpcIngress(
    rpcDomain,
    {
      metadata: {
        name: rpcDomain,
        namespace,
      },
      host: rpcDomain,
      tls: [
        {
          hosts: rpcCert.domainValidationOptions.apply((dvos) =>
            dvos.map((dvo) => dvo.domainName)
          ),
        },
      ],
      skipAwait: false,
    },
    { provider: cluster.provider, dependsOn: ingressDependencies }
  )
```

harryttd: this may be helpful: https://github.com/oxheadalpha/tezos-pulumi/blob/main/aws/components/ingress/helpers.ts

nicolasochem: This branch is deployed into oxheadinfra:

https://github.com/oxheadalpha/oxheadinfra/blob/main/mainnet-baker-monitor/pyrometer-values.yaml#L1-L20

Please, let me know how to proceed.

nicolasochem: I just changed some part of the default config to make it more like what's in oxheadinfra. Let me know if it looks good now.

harryttd (approved):

PR #479, opened 2022-08-17
New faucet

nicolasochem:
Make faucet work with the new codebase:

https://github.com/avysel/tezos-faucet
https://github.com/avysel/tezos-faucet-backend

Introduce more chart yaml files inspired by the default chart from "helm new".

Delete the old faucet preaccount creation logic (no longer needed).

This has been deployed on teztnets for ~one month~ two months and works.


orcutt989:
Some minor things to consider from kube-score.

```bash
 helm template -n oxheadalpha -f faucet_values.yaml charts/tezos-faucet | kube-score score -
apps/v1/Deployment tezos-faucet in oxheadalpha                                💥
    [CRITICAL] Container Security Context User Group ID
        · faucet-backend -> Container has no configured security context
            Set securityContext to run the container in a more secure context.
        · faucet -> Container has no configured security context
            Set securityContext to run the container in a more secure context.
    [CRITICAL] Container Security Context ReadOnlyRootFilesystem
        · faucet-backend -> Container has no configured security context
            Set securityContext to run the container in a more secure context.
        · faucet -> Container has no configured security context
            Set securityContext to run the container in a more secure context.
    [CRITICAL] Container Resources
        · faucet-backend -> CPU limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.cpu
        · faucet-backend -> Memory limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.memory
        · faucet-backend -> CPU request is not set
            Resource requests are recommended to make sure that the application
            can start and run without crashing. Set resources.requests.cpu
        · faucet-backend -> Memory request is not set
            Resource requests are recommended to make sure that the application
            can start and run without crashing. Set resources.requests.memory
        · faucet -> CPU limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.cpu
        · faucet -> Memory limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.memory
        · faucet -> CPU request is not set
            Resource requests are recommended to make sure that the application
            can start and run without crashing. Set resources.requests.cpu
        · faucet -> Memory request is not set
            Resource requests are recommended to make sure that the application
            can start and run without crashing. Set resources.requests.memory
    [CRITICAL] Container Image Pull Policy
        · faucet-backend -> ImagePullPolicy is not set to Always
            It's recommended to always set the ImagePullPolicy to Always, to
            make sure that the imagePullSecrets are always correct, and to
            always get the image you want.
        · faucet -> ImagePullPolicy is not set to Always
            It's recommended to always set the ImagePullPolicy to Always, to
            make sure that the imagePullSecrets are always correct, and to
            always get the image you want.
    [CRITICAL] Container Ephemeral Storage Request and Limit
        · faucet-backend -> Ephemeral Storage limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.ephemeral-storage
        · faucet -> Ephemeral Storage limit is not set
            Resource limits are recommended to avoid resource DDOS. Set
            resources.limits.ephemeral-storage
    [CRITICAL] Pod NetworkPolicy
        · The pod does not have a matching NetworkPolicy
            Create a NetworkPolicy that targets this pod to control who/what
            can communicate with this pod. Note, this feature needs to be
            supported by the CNI implementation used in the Kubernetes cluster
            to have an effect.
    [CRITICAL] Pod Probes
        · Container is missing a readinessProbe
            A readinessProbe should be used to indicate when the service is
            ready to receive traffic. Without it, the Pod is risking to receive
            traffic before it has booted. It's also used during rollouts, and
            can prevent downtime if a new version of the application is failing.
            More information: https://github.com/zegl/kube-score/blob/master/README_PROBES.md
networking.k8s.io/v1/Ingress tezos-faucet                                     💥
    [CRITICAL] Ingress targets Service
        · /info -> No service match was found
            No service with name tezos-faucet and port number 3000 was found
        · /send -> No service match was found
            No service with name tezos-faucet and port number 3000 was found
        · / -> No service match was found
            No service with name tezos-faucet and port number 80 was found
v1/Service tezos-faucet in oxheadalpha                                        🤔
    [WARNING] Service Type
        · The service is of type NodePort
            NodePort services should be avoided as they are insecure, and can
            not be used together with NetworkPolicies. LoadBalancers or use of
            an Ingress is recommended over NodePorts.
```

orcutt989 (approved):
LGTM

nicolasochem:
I'm not quite sure about kube-score. Most of these things are real issues but not fixable easily, and they are all accross tezos-k8s and we should fix it as a stretch goal across the codebase:

* requests, limits not set: ok
* nodeport service not recommended: ok
* network policy should be present: ok

I disagree with some other recommendations:

* no liveness probe: for such a simple service, does it matter?
* set pull policy to always: why? if we enforce unique container tags, it's not necessary

And some look wrong:
* no service tezos-faucet exposing port 3000: that's not true, it is there
* ephemeral storage limit & request not set: these pods are not using ephemeral storage

So I'll ignore but let's revisit this cross-codebase.

nicolasochem:
I just redeployed to teztnets and faucets still work after the most recent configmap+ secret change, merging this.

orcutt989:
> I'm not quite sure about kube-score. Most of these things are real issues but not fixable easily, and they are all accross tezos-k8s and we should fix it as a stretch goal across the codebase:
> 
> * requests, limits not set: ok
> * nodeport service not recommended: ok
> * network policy should be present: ok
> 
> I disagree with some other recommendations:
> 
> * no liveness probe: for such a simple service, does it matter?
> * set pull policy to always: why? if we enforce unique container tags, it's not necessary
> 
> And some look wrong:
> 
> * no service tezos-faucet exposing port 3000: that's not true, it is there
> * ephemeral storage limit & request not set: these pods are not using ephemeral storage
> 
> So I'll ignore but let's revisit this cross-codebase.

https://www.sobyte.net/post/2022-07/k8s-limit-local-temporary-storage/

```
Temporary storage, such as emptyDir volumes, container logs, image layers and container writable layers, use /var/lib/kubelet
by default, which protects the root partition of the node by limiting the capacity of temporary storage. By limiting the 
temporary storage capacity, you can also protect the root partition of the node.
```

I think the implication is that ephemeral storage is enabled by default and emptyDirs arent the only thing that use ephemeral storage. By not setting a limit I believe you are putting the node storage at risk potentially.

PR #480, opened 2022-08-23
Prometheus Exporter for Pyrometer

nicolasochem:
Quoting pyrometer [architecture doc](https://gitlab.com/tezos-kiln/pyrometer/-/blob/main/doc/monitoring.md):

> Primary installation target for initial monitoring implementation is a
personal computer. Consequently, implementation should prioritize
simplicity when it comes to number of individual, isolated components,
processes, their runtime dependencies,
administration/configuration.

Pyrometer is a self-sustaining tool that manages its own alerts and alerting channels. But what if I want to use it as a pure monitoring tool and and off the alert management to my prometheus stack in my kubernetes cluster?

I am introducing a Prometheus exporter for pyrometer. It consumes pyrometer events with the webhook and monitors only one of them: baker health status. It then aggregates the number of unhealthy bakers and exposes this as a prometheus metric.

The exporter is a sidecar so no modifications to pyrometer are required. The sidecar is included in the chart and relies on the utils container (just like the tezos sidecar). The ServiceMonitor and PrometheusRule are also included in the chart.

This gives you:

* the concept of an active alert that can be fed into an incident management system such as pagerduty.
* the ability to monitor a baker baking for several addresses, where it is not desirable to alert for an individual unhealthy address, but only when all the configured bakers are unhealtly. The threshold is configured in the chart.

I also add a default config in the values.yaml. Launching the tezos chart with no parameters, and the pyrometer chart with no parameters, will result in pyrometer acting as node monitoring for the node.

### DIFF
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -28,6 +28,12 @@ ingress:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - chart-example.local
+
+# Prometheus Operator is required in your cluster in order to enable
+# serviceMonitor and prometheusRule below.
 serviceMonitor:
   enabled: false
+prometheusRule:
+  enabled: false
   numberOfUnhealthyBakersAlertThreshold: 1
+  labels: {}
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -2,6 +2,9 @@
 config: {}
 images:
   pyrometer: registry.gitlab.com/tezos-kiln/pyrometer:latest
+service:
+  type: ClusterIP
+  port: 80
 ingress:
   enabled: false
   className: ""
--- a/charts/pyrometer/values.yaml
+++ b/charts/pyrometer/values.yaml
@@ -32,7 +32,7 @@ ingress:
 # Prometheus Operator is required in your cluster in order to enable
 # serviceMonitor and prometheusRule below.
 serviceMonitor:
-  enabled: true
+  enabled: false
 prometheusRule:
   enabled: false
   numberOfUnhealthyBakersAlertThreshold: 1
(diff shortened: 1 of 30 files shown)


charts/pyrometer/values.yaml
```diff
@@ -1,7 +1,18 @@
...
+config:
+  node_monitor:
+    nodes:
+    - http://tezos-node-rpc:8732
+  webhook:
+    enabled: true
+    url: http://127.0.0.1:31732/pyrometer_webhook
+service:
+  type: ClusterIP
+  port: 80
```
harryttd, line 15: I think current file doesn't have this bec it is hardcoded in the svc so i don't think it needs to be added back in. The only reason we might want it is to templatize these values for both the svc and ing instead of hardcoding the values in both

charts/pyrometer/values.yaml
```diff
@@ -13,3 +24,12 @@ ingress:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - chart-example.local
+
+# Prometheus Operator is required in your cluster in order to enable
+# serviceMonitor and prometheusRule below.
+serviceMonitor:
+  enabled: false
```
harryttd, line 31: do we want to have labels too?

harryttd (commented):
Looks good just a couple of questions. Also i didn't test and am assuming it works

charts/pyrometer/values.yaml
nicolasochem: thanks, you made this remark in the other PR and you are right. It was re-added in a rebase error, fixed now.

charts/pyrometer/values.yaml
nicolasochem: Not needed in my case: the prometheusrule needs a label to attach itself to the right prometheus, while all service monitors are scraped by the operator by default.

harryttd (approved):

PR #488, opened 2022-09-13
Octez: Add storage upgrade container (for master branch and release)

orcutt989:
This adds a storage init container to the master branch so we can release ahead of ghostnet/kathmandu activation.

See https://github.com/oxheadalpha/tezos-k8s/pull/486#issuecomment-1244943539 for why it is not parametrizable for now.

Includes efficiency suggested here by @harryttd  https://github.com/oxheadalpha/tezos-k8s/pull/486/files#r968888271

This is already implemented in the [monosite](https://github.com/oxheadalpha/tezos-k8s/commit/9b00e7df7d209ffddbdbafe103e59eb2b7582540#diff-551782c22fef270d6ceef4d2786b684f8e5585179b4f115ccac22653df0267bc) branch, but @nicolasochem said he wanted it for release on master this week.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -229,6 +229,13 @@
   {{- end }}
 {{- end }}
 
+{{- define "tezos.init_container.upgrade_storage" }}
+    {{- include "tezos.generic_container" (dict "root"   $
+                                           "type"        "upgrade-storage"
+                                           "image"       "octez"
+    )  | nindent 0 }}
+{{- end }}
+
 {{- define "tezos.container.sidecar" }}
   {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
     {{- include "tezos.generic_container" (dict "root"  $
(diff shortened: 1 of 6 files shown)


nicolasochem (approved):
thanks, let's merge and do a release. It's ok that it's not configurable. If that turns out to be a problem, we can make it configurable later, but my bias is to help the user, and this always helps the user, even though it adds an init container that's useless most of the time.

charts/tezos/templates/_containers.tpl
```diff
@@ -229,6 +229,13 @@
   {{- end }}
 {{- end }}
 
+{{- define "tezos.init_container.upgrade_storage" }}
+    {{- include "tezos.generic_container" (dict "root"   $
+                                           "type"        "upgrade-storage"
+                                           "image"       "octez"
```
harryttd, line 235: Only as another FYI on my previous [comment]( https://github.com/oxheadalpha/tezos-k8s/pull/486/files#r968888271), just to understand how this code works:

The configmap will still be mounted bec the generic_container function adds it by default to utils docker containers. So you would do this instead to not mount it.
```yaml
 "with_config" 0
```

harryttd (approved):

PR #490, opened 2022-10-04
kathmandu private chain by default

nicolasochem:
Make kathmandu the default protocol for new private chains.

Also repair private chain: don't update storage when there is none.


orcutt989 (approved):
Great catch! Thanks!!

PR #496, opened 2022-10-11
fix pre-configured identities

nicolasochem:
There were two issues:

1. the logic that avoid performing storage upgrades when there is no storage was merely checking the existence of /var/tezos/node/data. But in case of [pre-generated identity passed in values.yaml](https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/values.yaml#L148-L153), we write a identity.json in this dir, so the node still tries to perform the upgrade and fails, failing the pod. So now, we check for the existence of the context dir, which will only exist in subsequent starts, or after snapshot or tarball import
2. The identity owner was set to the old uid of 100, setting it now to the new uid of 1000. (not sure if it is an actual problem, but I initially thought it was, so I fixed it either way)

### DIFF
--- a/charts/tezos/scripts/upgrade-storage.sh
+++ b/charts/tezos/scripts/upgrade-storage.sh
@@ -1,8 +1,8 @@
 set -ex
 
-if [ ! -e /var/tezos/node/data ]
+if [ ! -e /var/tezos/node/data/context ]
 then
-  printf "No data dir found, probably initial start, doing nothing."
+  printf "No context in data dir found, probably initial start, doing nothing."
   exit 0
 fi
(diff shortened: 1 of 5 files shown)


orcutt989:
Great find! Should we be using tezos user instead of uuid in case it changes again?

nicolasochem:
> Great find! Should we be using tezos user instead of uuid in case it changes again?

This runs in the utils container, not the octez container, so there is no tezos user, and if we add one, and the one in octez changes uid, we will have to change the uid in utils anyway.

orcutt989:
Lgtm

orcutt989 (approved):

PR #497, opened 2022-10-11
remove VDF conditional on protocol

nicolasochem:
After jakarta deprecation, every protocol now supports VDF daemon, so we remove the conditional. This conditional was temporary to support jakarta (which has no VDF daemon)


orcutt989 (approved):
Lgtm

PR #499, opened 2022-10-19
Move containers from docker hub to github registry

nicolasochem:
There are many advantages:
* it's better integrated with github
* when we add a new container, the package gets added automatically without any hack (we had to use a hack in docker hub because there is no API for adding a container)
* we can generate containers for every branch and PR

Besides it's totally transparent to the user: most users use our released helm charts wihch have the container path hardcoded (the `publish_helm` action takes care of this): here, I am just hardcoding the ghcr.io URL instead of the default docker hub location for all our packages.

So, there is no user-facing impact on this.

I want to go further and suggest that we change our submodule workflow.

Today, when we want to test a branch of tezos-k8s, we typically check out a submodule inside the infra repo and we have some logic to build containers within the submodule and push them to AWS ECR.

Instead, I am suggesting a new workflow: whenever we do a change to tezos-k8s, we have to set the tag for the PR build for every image inside of values.yaml. Submodules are still needed in case we change the helm chart, but we no longer build the containers in the submodule.

* **PRO**: we solve the problem of "leaky" submods: we had cases where different charts were on different submodules within a single repo, but it was all being pushed to a signle ECR registry and for some reason the builds were getting mixed up between namespaces
* **PRO**: faster deployments. The ECR workflow requires to build and push containers, which is fast locally in theory because of docker caching, but still heavy and slower than release deploys in my experience. Also, when using actions, there was no caching so a submod deploy was taking much longer
* **PRO**: we remove the dependency on AWS: this works on any k8s cluster now
* **CON**: we can't deploy with local changes from laptop. Any change has to be pushed to a branch (it can be on a fork if the code is not yet appropriate for pushing upstream)
* **CON**: we have to remember to set the tag to the PR build, then revert to release after release (although this could be fixed with a bit of extra logic in the code where we decide whether submod/release to build the helm params)

What do yall think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
       id: meta
       uses: docker/metadata-action@v3
       with:
-        images: ghcr.io/${{ github.repository_owner }}/${{ matrix.container }}
+        images: ghcr.io/${{ github.repository_owner }}/tezos-k8s-${{ matrix.container }}
         tags: |
           type=ref,event=branch
           type=ref,event=pr
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,60 +140,37 @@ jobs:
           echo $container_list
           echo "::set-output name=matrix::$container_list"
 
-  publish_containers:
-    # based on
-    # https://github.com/docker/build-push-action#usage
+  publish-to-ghcr:
     runs-on: ubuntu-latest
     needs: list_containers_to_publish
-    if: github.event_name == 'release' && github.event.action == 'created'
     strategy:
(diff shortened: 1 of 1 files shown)


orcutt989:
Awesome! Could we set it up so master defaults to release so we don't have to reset? I worked in an environment where we used branch names as the container tag for testing and master was latest. Perhaps we could do something like that ?

orcutt989 (approved):

nicolasochem:
> Awesome! Could we set it up so master defaults to release so we don't have to reset? I worked in an environment where we used branch names as the container tag for testing and master was latest. Perhaps we could do something like that ?

No objection on hard-coding "latest" in the chart in source code instead of today's "`dev`".

PR #500, opened 2022-10-26
Set node config for pods without dedicated pod configs 

puppetninja:


nicolasochem (commented):
This looks fine. Is it ready? If so, please add description to the PR. We would also consider moving the default config in values.yaml to use this instead of the per-instance config.

utils/config-generator.py
```diff
@@ -33,12 +48,17 @@
 
 for cl, val in NODES.items():
     if val != None:
+        # get general config
+        if "config" in val:
+            MY_POD_CONFIG["config"] = val["config"]
```
harryttd, line 53: Why is this needed exactly? The code currently merges the config here https://github.com/puppetninja/tezos-k8s/blob/0b4e3250057f75db9517506db64bb622b905dd6f/utils/config-generator.py#L561-L564

nicolasochem: But for that you need to set `node_globals`?

The goal of this is to set a common config for all nodes in the statefulset, not all nodes globally. I don't believe we have this capability now.

puppetninja: yes, basically we are trying to to do auto scaling for nodes instances, for which individual configs can not be defined in advances. We expect the config defined under `nodes` section would be inherited by instances that are scaled out

nicolasochem (approved):
I did a few tests and it is already possible to set config at statefulset level:

```
nodes:                                                                                                                         
  archive-baking-node:                                                                                                         
    runs:                                                                                                                      
      - octez_node                                                                                                             
      - baker                                                                                                                  
    storage_size: 15Gi                                                                                                         
    config:                                                                                                                    
      shell:                                                                                                                   
        history_mode: archive                                                                                                  
      metrics_addr:                                                                                                            
        - :9932                                                                                                                
    instances:                                                                                                                 
      - is_bootstrap_node: true 
        bake_using_accounts:
          - archive-baking-node-0
```

The change proposed (in the most recent version of the code) is an one-liner that ensures that `MY_POD_CLASS` is set regardless of whether the pod is listed in values.yaml `nodes:` section.

For all supported purposes, the pod should be listed. So this code change does not change functionality. However, in combination with a horizontal pod autoscaler (not supported in tezos-k8s yet), this will ensure that any additional replicas spun up by the autoscaler will also be configured according to the statefulset-level config in values.yaml (and node_globals as well).

I tested mkchain with no parameters and everything else looks like working, so I'm approving.

But before merging, please change the PR name and description to something that explains clearly why you are doing this. (feel free to copy and paste some of my approval above).

Also moving forward please push branches to the oxheadalpha fork so the container build tests will work.

puppetninja:
I have updated this PR description and added description to commit, I think I had permission issue to push to this repo's branches but let me try again...

puppetninja:
> I have updated this PR description and added description to commit, I think I had permission issue to push to this repo's branches but let me try again...

No, I still don't have permission...

harryttd:
Do we want to allow non org users to push to our repo? That would mean all those users can run actions against our repository. Not sure that is a good idea.

nicolasochem:
> Do we want to allow non org users to push to our repo? That would mean all those users can run actions against our repository. Not sure that is a good idea.

No, instead I added Chao to the group (I vouch for him) and protected the master branch.

puppetninja:
opened #507

PR #501, opened 2022-10-28
Service monitor

nicolasochem:
Add a service monitor to target the new metrics endpoint of v14.

Also remove the old "metrics" container since it is deprecated and replaced by the native endpoint.

Also add a serviceMonitor for convenience. We already have a service monitor in the pyrometer chart, I'm adding a similar one here. This helps the user not writing their own monitor, and saves a step in the upcoming guide.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -343,34 +343,6 @@
   {{- end }}
 {{- end }}
 
-{{- define "tezos.container.metrics" }}
-{{- if has "metrics" $.node_vals.runs }}
-- image: "registry.gitlab.com/nomadic-labs/tezos-metrics"
-  args:
-    - "--listen-prometheus=6666"
-    - "--data-dir=/var/tezos/node/data"
-  imagePullPolicy: IfNotPresent
-  name: metrics
-  ports:
-    - containerPort: 6666
-      name: tezos-metrics
-  volumeMounts:
-    - mountPath: /etc/tezos
-      name: config-volume
-    - mountPath: /var/tezos
-      name: var-volume
-    - mountPath: /etc/secret-volume
-      name: tezos-accounts
-  envFrom:
-    - configMapRef:
-        name: tezos-config
-  env:
-{{- include "tezos.localvars.pod_envvars" . | indent 4 }}
-    - name: DAEMON
-      value: tezos-metrics
-{{- end }}
-{{- end }}
-
 {{/*
 // * The zerotier containers:
 */}}
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -156,10 +156,10 @@
   ports:
     - containerPort: 8732
       name: tezos-rpc
-    - containerPort: 9001
-      name: metrics
     - containerPort: 9732
       name: tezos-net
+    - containerPort: 9932
+      name: metrics
     {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
   readinessProbe:
     httpGet:
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -156,6 +156,8 @@
   ports:
     - containerPort: 8732
       name: tezos-rpc
+    - containerPort: 9001
+      name: metrics
     - containerPort: 9732
       name: tezos-net
     {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
--- a/charts/tezos/templates/static.yaml
+++ b/charts/tezos/templates/static.yaml
@@ -18,11 +18,13 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ $key }}
+  labels:
+    appType: tezos-node
 spec:
   ports:
     - port: 9732
       name: p2p
-    - port: 9001
+    - port: 9932
       name: metrics
   publishNotReadyAddresses: true
   clusterIP: None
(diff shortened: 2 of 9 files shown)


orcutt989 (approved):
Looks great! Declutter FTW.

charts/tezos/templates/_containers.tpl
```diff
@@ -158,6 +158,8 @@
       name: tezos-rpc
     - containerPort: 9732
       name: tezos-net
+    - containerPort: 9932
+      name: metrics
```
harryttd, line 162: `{{- if .Values.serviceMonitor.enabled }}` ?

charts/tezos/templates/static.yaml
```diff
@@ -18,7 +18,14 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ $key }}
+  labels:
+    appType: tezos-node
 spec:
+  ports:
+    - port: 9732
+      name: p2p
```
harryttd, line 26: I'm curious, why wasn't this needed before?

charts/tezos/templates/static.yaml
```diff
@@ -18,7 +18,14 @@ apiVersion: v1
...
+    - port: 9932
+      name: metrics
```
harryttd, line 28: `{{- if .Values.serviceMonitor.enabled }}` ?

charts/tezos/values.yaml
```diff
@@ -230,6 +230,7 @@ nodes:
         config:
           shell:
             history_mode: rolling
+          metrics_addr: [ "0.0.0.0:9932" ]
```
harryttd, line 233: Comment out? bec monitor is default not enabled

charts/tezos/templates/_containers.tpl
harryttd, line 162: `{{- if .Values.serviceMonitor.enabled }}` ?

harryttd (changes requested):

charts/tezos/templates/static.yaml
nicolasochem: The metrics endpoint needs to be here so the servicemonitor can pick up the metrics. While debugging this, I noticed there was no endpoint associated with the p2p port either, so I added it. It never mattered, because:
* internally we target pods directly
* when we want external p2p access we deploy a different service anyway, of type `LoadBalancer`

I guess this service was pretty useless?

nicolasochem:
@harryttd answering all of your remaining comments at once:

I have enabled metrics collection by default on tezos-k8s, by choice.

I have also ensured that octez default config in tezos-k8s is exposing its prometheus port.

It's better this way, than making it dependent of `ServiceMonitor` being `enabled` in the values: you could want to scrape this ports with something else than the prometheus operator (for example a manually provisioned prometheus).

So then, we would have `metricsExposed: true` separate from `serviceMonitor: enabled: true` but I feel that it's simpler to have it enabled all the time. It's not a security concern (it stays within the cluster and you can't kill the node with such metrics requests, in theory)

A future `NetworkPolicy` in our chart will tighten security appropriately, I think.

charts/tezos/templates/static.yaml
harryttd: The service is a headless service. K8s then allows you to target each pod individually instead of creating a clusterIP that would load balance across the pods.

I wonder because ports are now defined, rpc port isn't, so would that break? i'm not sure i get why we need to set ports here. Being we can anyways target the pod and use the port number.

nicolasochem: I just tried mkchain with 2 bakers and 1 node, and nodes are peering with each other and everything is working. So, it didn't break. I'm not sure why p2p is here though. Maybe safe to remove.

RPC is served from a different service.

It makes sense to have a p2p service targeting several pods, because it's fine to load balance the p2p connections. But, I did forget that the purpose of this (headless) service was to be able to target pods individually.

Knowing that, maybe it would have been better to leave it alone, and set the metrics port on the rpc service.

I'm pretty sure I did this for a reason. Probably the service monitor would not work if the port was not named on the service.

harryttd: an example where we target a specific  pod rpc endpoints is indexers: https://github.com/oxheadalpha/tezos-k8s/blob/cb8eed6aa0d0627cc46ffc9986c6a5eca61be323/charts/tezos/values.yaml#L449

So we need to know that we can still target any port we want. If we can, i don't know why we need to specify any ports now. 9732, and the metrics port. As long as the sts pod containers are exposing the port.

It appears to me the tezos-service-monitor is load balancing across all nodes. Does this allow us to monitor each node individually instead of conflating each nodes' metric data together?

nicolasochem: Screenshot from 2022-10-28 17-38-58

it's labeling per pods

harryttd: so every 15s it sends a request to /metrics and each pod will eventuallly be hit. What if there are many many pods? They won't have their metrics updated for a while.

harryttd: >So we need to know that we can still target any port we want. If we can, i don't know why we need to specify any ports now. 9732, and the metrics port. As long as the sts pod containers are exposing the port.

once we deploy this release, we can prob test rpc of a specific node just by sshing into another one. That is prob sufficient. And then we can also remove the ports from the headless svc later just so it isn't confusing why we are exposing some ports on the svc and not others. not a big deal

nicolasochem: I don't think that's what is going on. I am aware of "pod monitor" but I've never used it, ServiceMonitor is what I have been using.

See these heavily thumbed-up comments: https://github.com/prometheus-operator/prometheus-operator/issues/3119#issuecomment-610445145

PR #502, opened 2022-10-28
Add reources management for nodes

puppetninja:


charts/tezos/templates/_containers.tpl
```diff
@@ -43,6 +43,7 @@
      *    with_config    bring in the configMap defaults true only on utils.
      *    with_secret    bring in the secrets map including the identities.
      *    localvars      set env vars MY_* Defaults to true only on utils.
+     *    resources      set resources management to be empty.
```
nicolasochem, line 46: unclear what this means

charts/tezos/templates/_containers.tpl
```diff
@@ -238,9 +246,11 @@
 
 {{- define "tezos.container.sidecar" }}
   {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
-    {{- include "tezos.generic_container" (dict "root"  $
-                                                "type"  "sidecar"
-                                                "image" "utils"
+    {{- $sidecarResources := dict "requests" (dict "memory" "150Mi") "limits" (dict "memory" "200Mi") -}}
```
nicolasochem, line 249: sidecar resources and limits are hardcoded? why are they necessary?

charts/tezos/templates/_containers.tpl
```diff
@@ -260,24 +270,45 @@
 
 {{- define "tezos.container.node" }}
 {{- if eq (include "tezos.getNodeImplementation" $) "octez" }}
-    {{- include "tezos.generic_container" (dict "root"        $
-                                                "type"        "octez-node"
-                                                "image"       "octez"
-                                                "with_config" 0
-    ) | nindent 0 }}
+    {{- if hasKey $.node_vals "resources" }}
```
nicolasochem, line 273: there may be a more compact way to ahcieve this? if resources are undefined, then omit these settings in the final object returned by the function

nicolasochem (commented):
Left some inline questions. Also, it would be good to "document" this in the comments of `charts/tezos/values.yaml`.

charts/tezos/templates/_containers.tpl
puppetninja: yes, I will work on this

charts/tezos/templates/_containers.tpl
puppetninja: I think user has no overview of resource needs for sidecar so we should hard code it here as when user doesn't specify the readiness_probe we inject this side automatically to replace it

puppetninja:
opened #508

PR #504, opened 2022-11-04
Update default tezos-k8s images to be GHCR one

puppetninja:


nicolasochem (approved):
Sounds good, @orcutt989 was asking for the same thing here

https://github.com/oxheadalpha/tezos-k8s/pull/499#issuecomment-1284074350

The only issue I have is that a new developer may make the assumption that their change was taken into account, but if the change is in the image, not the chart, running it on minikube would not propagate them.

For that to happen you need a development environment such as devspace. We have not really been using devspace, and all developers should be now aware that any change to the image requires pointing to ghcr's branch in the values file.

Let's wait for another approval.

orcutt989:
Are we currently building and pushing to GHCR @nicolasochem ? I know we merged your PR, but I haven't thought about how this works enough. Don't we want people who use the helm chart for it to come from wherever we are pushing? I don't think snapshot engine container is in GHCR yet, but I may be wrong- I'm just thinking out loud.

orcutt989:
> > Are we currently building and pushing to GHCR @nicolasochem ?
> 
> We sure are.
> 
> > Don't we want people who use the helm chart for it to come from wherever we are pushing?
> 
> What do you mean?
> 
> > I don't think snapshot engine container is in GHCR yet, but I may be wrong- I'm just thinking out loud.
> 
> Yes it is: https://github.com/oxheadalpha/tezos-k8s/pkgs/container/tezos-k8s-snapshotengine

Ah I see! Well then sounds good to me. Thanks for clarifying!

orcutt989:
@puppetninja Could you update the description with a “what”, “why”, and “how” sort of explanation of what this does?

orcutt989:
@nicolasochem CI was failing im not sure id that was a concern…

nicolasochem:
> @nicolasochem CI was failing- I'm not sure if that was a concern…

It's because the PR was opened from a fork, so docker images won't build, but this PR is not touching the containers so it looked fine.

harryttd:
Shouldn't the tag be master and not latest? The current latest image in the registry is 6.11.1. The image could become out of date when master is updated but doesn't become latest tag. Like it is right now.

PR #506, opened 2022-11-10
Default image tags should be master, not latest

harryttd:
https://github.com/oxheadalpha/tezos-k8s/pull/504#issuecomment-1309538319
> Shouldn't the tag be master and not latest? The current latest image in the registry is 6.11.1. The image could become out of date when master is updated but doesn't become latest tag. Like it is right now.
> 
> So if i pull master, the image could be wrong and not reflect what the code actually is on master.


nicolasochem (approved):
ty

PR #508, opened 2022-11-12
Add reources management for nodes

puppetninja:
This is a duplicated one of #502, PR generated from official repo branch to run all the CI jobs

This PR would set the resource manage for octez nodes and sidecar in the same pod. For bakers, there are no resources management for it however

Users can manage the node resources for octez node placement. Additionally, users can use k8s autoscalers to scale out node pods based on certain criteria
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale
This scaling mechanism is for pod, so both octez node and sidecar resources should be specified for the pod resource usage to be calculated. For now, pods with bakers running inside is not supported.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -43,6 +43,8 @@
      *    with_config    bring in the configMap defaults true only on utils.
      *    with_secret    bring in the secrets map including the identities.
      *    localvars      set env vars MY_* Defaults to true only on utils.
+     *    resources      set container resources management, i.e. request
+     *                   and limit, default value is an empty dict.
      */ -}}
 
 {{- define "tezos.generic_container" }}
@@ -74,6 +76,9 @@
   {{- if eq .image "utils" }}
     {{- $_ := set . "args" (list .type) }}
   {{- end }}
+{{- end }}
+{{- if not (hasKey . "resources") }}
+    {{- $_ := set . "resources" dict }}
 {{- end }}
 
 {{- /*
@@ -164,6 +169,10 @@
       port: 31732
     {{- end }}
   {{- end }}
+{{- if .resources }}
+  resources:
+{{ toYaml .resources | indent 4 }}
+{{- end }}
 {{- end }}
 
 
@@ -237,9 +246,11 @@
 
 {{- define "tezos.container.sidecar" }}
   {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
-    {{- include "tezos.generic_container" (dict "root"  $
-                                                "type"  "sidecar"
-                                                "image" "utils"
+    {{- $sidecarResources := dict "requests" (dict "memory" "80Mi") "limits" (dict "memory" "100Mi") -}}
+    {{- include "tezos.generic_container" (dict "root"      $
+                                                "type"      "sidecar"
+                                                "image"     "utils"
+                                                "resources" $sidecarResources
     ) | nindent 0 }}
   {{- end }}
 {{- end }}
@@ -249,6 +260,7 @@
                                                 "type"        "octez-node"
                                                 "image"       "octez"
                                                 "with_config" 0
+                                                "resources"   $.node_vals.resources
     ) | nindent 0 }}
 {{- end }}
 
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -114,6 +114,8 @@ should_generate_unsafe_deterministic_data: false
 #          using just "baker".
 # - "storage_size": the size of the PV
 # - "resources": resources specifications for the node.
+#                Optionally set resources and limits for octez node
+#                See: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
 # - "images": Optional specification of images to use for the tezos node and
 #           baker. Options are "octez" with a tezos/tezos image.
 #           If no images are provided, the containers will default to the images
@@ -188,8 +190,6 @@ should_generate_unsafe_deterministic_data: false
 #     runs:
 #       - octez_node
 #       - baker
-#     # You can optionally set resources and limits for your octez node
-#     # See: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
 #     resources:
 #       requests:
 #         memory: 16192Mi
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -113,6 +113,7 @@ should_generate_unsafe_deterministic_data: false
 #          bakers and accusers, so "baker-011-pthangz2" is configured
 #          using just "baker".
 # - "storage_size": the size of the PV
+# - "resources": resources specifications for the node.
 # - "images": Optional specification of images to use for the tezos node and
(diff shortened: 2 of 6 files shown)


charts/tezos/templates/_containers.tpl
```diff
@@ -240,9 +249,11 @@
 
 {{- define "tezos.container.sidecar" }}
   {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
-    {{- include "tezos.generic_container" (dict "root"  $
-                                                "type"  "sidecar"
-                                                "image" "utils"
+    {{- $sidecarResources := dict "requests" (dict "memory" "80Mi") "limits" (dict "memory" "100Mi") -}}
```
puppetninja, line 252: we hardcode a sidecar when readniess probe is not specified, follow the same way to set the resources part

nicolasochem (changes requested):
This still requires "documentation" in the chart's default values.yaml 

What we typically do is provide a commented-out example in the values file, with some additional explanatory commentary.

charts/tezos/templates/_containers.tpl
```diff
@@ -261,25 +272,36 @@
...
+  {{- if eq (include "tezos.getNodeImplementation" $) "octez" }}
+    {{- $octezNodeParams := dict "root"        $
+                                 "type"        "octez-node"
+                                 "image"       "octez"
+                                 "with_config" 0
+    -}}
+
+    {{- if hasKey $.node_vals "resources" }}
+      {{- $_ := set $octezNodeParams "resources" $.node_vals.resources -}}
+    {{- end }}
```
harryttd, line 284: I don't think this is needed. `node_vals` is already set on `$`, so it can be directly accessed in the generic container function. The function itself checks for the existence of a `resources` property on the `node_vals`.

This comment applies to the other similar changes in this file.

charts/tezos/values.yaml
```diff
@@ -192,6 +192,11 @@ should_generate_unsafe_deterministic_data: false
 #       - octez_node
 #     # - tezedge_node
 #       - baker
+#    resources:
+#      requests:
+#        memory: 8096Mi
+#      limits:
+#        memory: 9120Mi
```
harryttd, line 199: - The indentations is off a little bit
- A comment should be left stating that currently resources are only set for node/baker.
- a comment for `resources` should be put under the  `# Params at the statefulset level:` section above. Perhaps put it under `storage_size`.

harryttd (changes requested):

charts/tezos/templates/_containers.tpl
```diff
@@ -43,6 +43,8 @@
      *    with_config    bring in the configMap defaults true only on utils.
      *    with_secret    bring in the secrets map including the identities.
      *    localvars      set env vars MY_* Defaults to true only on utils.
+     *    resources      set container resources management, i.e. request
+     *                   and limit, default value is an empty dict.
```
harryttd, line 47: This shouldn't be needed. Just check for it on `$.node_vals`.

charts/tezos/values.yaml
```diff
@@ -187,6 +187,11 @@ should_generate_unsafe_deterministic_data: false
 #     runs:
 #       - octez_node
 #       - baker
+#    resources:
+#      requests:
+#        memory: 8096Mi
+#      limits:
```
nicolasochem, line 193: Maybe the example limit should be set a 16Gi? Double of the request? We have not used limits so I don't want to give too low a value as an example...

charts/tezos/values.yaml
```diff
@@ -187,6 +187,11 @@ should_generate_unsafe_deterministic_data: false
 #     runs:
 #       - octez_node
 #       - baker
```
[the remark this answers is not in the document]
puppetninja: done

charts/tezos/templates/_containers.tpl
puppetninja: I added this as we use this to hardcode the resource managements for side also, so the resource management is set for the whole pod and can be applied for autoscaler policies

puppetninja: so not just checking `$.node_vals` for the node per se is what I meant

charts/tezos/templates/_containers.tpl
```diff
@@ -260,12 +277,18 @@
...
-                                                  "image"       "octez"
-      ) | nindent 0 }}
+      {{- $bakerParams := dict "root" $
+                               "name" (print "baker-"
+                                       (lower .command))
+                               "type"        "baker"
+                               "image"       "octez"
+      -}}
+
+      {{- if hasKey $.node_vals "resources" }}
```
nicolasochem, line 287: it looks like you can simply pass `"resources" $.node.vals.resources` and if the key is absent it would pass null, and still work? So you can avoid setting $bakerParams at all and just modify the dict passed to `include "tezos.generic_container"`.

nicolasochem (commented):
Why assign requests to the baker as well as the node? Then if you set requests of 3300Mi on a baking node, the pod actually requires 6600Mi. and more if you have more bakers. Can we set a request for the node only, so the value configured ends up being the total pod request?

charts/tezos/templates/_containers.tpl
puppetninja: this is always part of the pod, only by adding the autoscaler can calculate the total resource consumption made by the pod

charts/tezos/values.yaml
puppetninja: done

nicolasochem:
@harryttd @orcutt989 all issues resolved, please review again

orcutt989:
@nicolasochem @harryttd There are still requested changes on this PR. Please dismiss them if they have been addressed.

nicolasochem (approved):

harryttd:
I'm reviewing right now

charts/tezos/templates/_containers.tpl
nicolasochem: @harryttd the sidecar is also instantiated using generic_container, so resource have to be a parameter to the function. otherwise the sidecar would have the same resources than the actual node.

charts/tezos/values.yaml
```diff
@@ -187,6 +188,13 @@ should_generate_unsafe_deterministic_data: false
...
+#     # You can optionally set resources and limits for your octez node/baker
+#     # See: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```
harryttd, line 192:
```suggestion
#     # You can optionally set resources and limits for your octez node
#     # See: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```

Also shouldn't this be on line 116 instead?

harryttd (changes requested):
Almost ready. I think i understand the PR more fully now. I didn't get that the scaler wouldn't scale if it was targeting specific containers as opposed to the entire pod resources. So if it couldn't find any resources defined on any container it wouldn't scale.

charts/tezos/values.yaml
puppetninja: Indeed

harryttd (approved):

PR #510, opened 2022-11-21
snapshotEngine: Skip snapshot if metadata missing when building website

orcutt989:
We had an issue where the website was not building because one artifact's metadata did not have a tezos version. This was due to a fluke and we are investigating, but in the mean-time it would be best to build the site despite missing metadata.

### DIFF
--- a/snapshotEngine/getLatestSnapshotMetadata.py
+++ b/snapshotEngine/getLatestSnapshotMetadata.py
@@ -46,7 +46,10 @@ for network, snapshots in snapshots_per_network.items():
         # Parses date from tezos build of each artifact, compares to last date, updates if older, otherwise its newer
         for snapshot in snapshots:
             matches=datefinder.find_dates(snapshot['tezos_version'])
-            tezos_build_datetime=list(matches)[0]
+            try:
+                tezos_build_datetime=list(matches)[0]
+            except IndexError:
+                continue
             if tezos_build_datetime < last_tezos_build_datetime:
                 latest_tezos_build_version=[src for time, src in datefinder.find_dates(snapshot['tezos_version'], source=True)][1]
                 last_tezos_build_datetime=tezos_build_datetime


nicolasochem (approved):
agreed, thanks

snapshotEngine/getLatestSnapshotMetadata.py
```diff
@@ -46,7 +46,10 @@
         # Parses date from tezos build of each artifact, compares to last date, updates if older, otherwise its newer
         for snapshot in snapshots:
             matches=datefinder.find_dates(snapshot['tezos_version'])
-            tezos_build_datetime=list(matches)[0]
+            try:
+                tezos_build_datetime=list(matches)[0]
+            except IndexError:
+                continue
             if tezos_build_datetime < last_tezos_build_datetime:
```
harryttd, line 53: I have not tested this code but just from looking at it I speculate if it will work. If there is an index error the code will continue. But it immediately assumes that `tezos_build_datetime` exists even if there was an index error, and compares it to `last_tezos_build_datetime`. I assume either it won't fine the former variable or there will be a type error.

orcutt989: Thanks- are you suggesting I move a test for tezos build time into the try as well?

harryttd, line 53: I Believe either of these should work. But I would test.

```suggestion
            tezos_build_datetime=list(matches)[0] if len(matches) else None
            if tezos_build_datetime and tezos_build_datetime < last_tezos_build_datetime:
```
-----
```suggestion
            try:
                tezos_build_datetime=list(matches)[0]
            except IndexError:
                tezos_build_datetime = None
            if tezos_build_datetime and tezos_build_datetime < last_tezos_build_datetime:
```

PR #511, opened 2022-11-30
Multiple baker processes

nicolasochem:
When baking for several addresses, it's recommended to spin up several baking processes.

Otherwise, baking is sequential and can be too slow, especially in combination with remote signers.

This PR switches to one baking process per baker.

There are problems: how many processes do we start? There has to be a constant number per statefulset.

The most sensible thing seems to be to lock the replica count of any baking statefulset to one?

Or enforce a constant number of bakers for all members of the statefulset within helm? I'm open to suggestions here.

There is a mechanism from @elric1 to guess the baking account when it's not specified. Since this change likely breaks it, I removed it.


nicolasochem:
replaced by #517

PR #512, opened 2022-11-30
remove traces of tezedge

nicolasochem:
Tezedge has been deprecated, removing this option from tezos-k8s.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -83,7 +83,7 @@
 {{- $node_vals_images := $.node_vals.images | default dict }}
 {{- if eq .image "octez" }}
   image: "{{ or $node_vals_images.octez $.Values.images.octez }}"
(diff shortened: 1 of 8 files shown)


charts/tezos/templates/_containers.tpl
```diff
@@ -83,9 +83,7 @@
 {{- $node_vals_images := $.node_vals.images | default dict }}
 {{- if eq .image "octez" }}
   image: "{{ or $node_vals_images.octez $.Values.images.octez }}"
-{{- else if eq .image "tezedge" }}
-  image: "{{ or $node_vals_images.tezedge $.Values.images.tezedge }}"
-{{- else }}
+{{- else if eq .image "utils" }}
```
harryttd, line 86: This should prob just be an `else` bec if no image is provided then no image will be set

harryttd (changes requested):

harryttd (approved):

PR #514, opened 2022-12-07
TRD mainnet adaptations

nicolasochem:
I deployed TRD against a mainnet baker with tezos-k8s and found a few missing parameters from the original work. Adding them.

Also, I am commenting out by default the TRD params in values.yaml, otherwise they get deep-merged by helm, which yields odd results.

I am also adding the option to upload the resulting data from trd to a bucket. Since TRD runs in a cronjob, it's hard to shell into it to look at the files. Uploading the entire PVC to a bucket is a nice substitute.

### DIFF
--- a/charts/tezos-reward-distributor/templates/cronjob.yaml
+++ b/charts/tezos-reward-distributor/templates/cronjob.yaml
@@ -27,10 +27,10 @@ spec:
           volumes:
             - name: storage
               persistentVolumeClaim:
-                claimName: tezos-reward-distributor-volume
+                claimName: {{ include "tezos-reward-distributor.fullname" . }}-volume
             - name: config-volume
               configMap:
-                 name: tezos-reward-distributor-config
+                name: {{ include "tezos-reward-distributor.fullname" . }}-config
           initContainers:
             # Work around a bug where fsGroup is ignored
             - name: change-ownership-container
--- a/charts/tezos-reward-distributor/templates/cronjob.yaml
+++ b/charts/tezos-reward-distributor/templates/cronjob.yaml
@@ -31,6 +31,9 @@ spec:
             - name: config-volume
               configMap:
                 name: {{ include "tezos-reward-distributor.fullname" . }}-config
+            - name: secret-volume
+              secret:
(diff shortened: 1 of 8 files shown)


charts/tezos-reward-distributor/templates/cronjob.yaml
```diff
@@ -44,18 +45,61 @@ spec:
...
+              command:
+              - /bin/sh
+              args:
+              - "-c"
+              - |
+{{ tpl ($.Files.Get (print "scripts/bucket_upload.sh")) $ | indent 16 }}
+              env:
+              - name: AWS_ACCESS_KEY_ID
+                value: "{{ .Values.bucket_upload.access_key_id }}"
+              - name: AWS_SECRET_ACCESS_KEY
```
harryttd, line 95: should be in secret preferably mounted as file

harryttd (changes requested):
When do the values get deep merged?

nicolasochem:
> When do the values get deep merged?

When you pass custom values.yaml with TRD config, it gets merged with this dummy default config that was in the yaml. It's best to put nothing and let the user set the entire config in custom values.ayml

harryttd:
ah understood. How about leaving the properties as actual values but leaving them empty or with empty strings for example. Especially for properties that must be set to indicate to the user they must set them

charts/tezos-reward-distributor/values.yaml
```diff
@@ -117,7 +117,8 @@ trd_config:
 # when the cronjob is not running.
 bucket_upload:
   access_key_id:
```
harryttd, line 119: access_key_id should really also be a secret. Then you can write the secrets to a file that is stored in the Secret itself. Instead of using properties in the Secret.
```
export AWS_SECRET...
export AWS_ACCESS...
```
and then just `source` it in the upload script without needing to explicitly write `export ...`

harryttd: https://kubernetes.io/docs/concepts/configuration/secret/#use-case-dotfiles-in-a-secret-volume

harryttd (approved):
awesome. looks good. i'm assuming it works

PR #517, opened 2022-12-08
one process per baker

nicolasochem:
When baking for several addresses, today we instantiate only one baker process. This is unsupported and bad, because it does things sequentially. The supported way to bake for several addresses is to run one baking process per address. This PR changes tezos-k8s to do that.

This is a second, less intrusive attempt after feedback from Aryeh.

* helm computes the maximum number of baking processes that any pod in the statefulset would need
* instantiates the max number of bakers for each pod (no choice here..)
* the baker entrypoint figures which baker it must bake for thanks to a new BAKER_INDEX env var. If it must idle, it enters an infinite sleeping loop.

This works when using `bake_using_accounts`. I verified that I did not break `bake_using_account`.

I verified by:
* creating a private chain with 3 bakers using mkchain
* assigning two bakers to the first node and one baker to the second

I also ran mkchain with no parameters to ensure it still works.

Various cleanups:

* remove a octez v13 specific catch (must pass toggle even when vote file absent) in the baker entrypoint

Note that in the future, the baker will be its own pod to this will become irrelevant.

### DIFF
--- a/charts/snapshotEngine/README.md
+++ b/charts/snapshotEngine/README.md
@@ -320,12 +320,12 @@ Overview of functionality of containers in Kubernetes Job Pods.
 
 ##### init-tezos-filesystem Container
 
-In order for the storage to be imported successfully to a new node, the storage needs to be initialized by the `tezos-node` application.
+In order for the storage to be imported successfully to a new node, the storage needs to be initialized by the `octez-node` application.
 
 This container performs the following steps -
 
 1. Chowns the history-mode-snapshot-cache-volume to 100 so subsequent containers can access files created in them.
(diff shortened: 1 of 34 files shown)


nicolasochem:
> I find the DAEMON env var useful because I set `PS1` to include it:
> 
> ```
> node_globals:
>   env:
>     all:
>       PS1: ": $HOSTNAME $DAEMON; "
>       TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER: "Y"
> ```

I reinstated it. Thanks for looking, a review would me much appreciated.

harryttd (commented):
Looks great to me. I have not thoroughly tested but the code looks good.

Do you mind:
- removing the comment `remove the DAEMON env var...` from the PR description
- leaving a comment that this PR's changes should be temporary. Based on changes that should be taking place in the future to how the node and baker will be functioning differently.

harryttd (approved):

PR #518, opened 2022-12-09
515 snapshot engine implement restriction on artifact creating frequency

orcutt989:
Snapshot Engine creates artifacts back-to-back by default. As soon as one job is done another one is started.  This can create A LOT of artifacts especially for non-mainnet networks.  This can be a headache to manage so many files and also could be very expensive to host. We've decided to implement a simple configurable parameter in the helm values for the snapshotEngine chart that will allow the job to delay for a user-determined amount of time.

**Usage:**
in the snapshot-values.yaml file one would put their desired sleep time (a valid unix sleep time value) as such -
```yaml
artifactDelay:
  rolling: 1m
  archive: 0m
```

This will delay the artifact generation by a user determined amount of time for a given chain namespace and a chain history mode.

**Notes:**
It is a rudimentary  feature for the time being.  The time is not factored in to artifact generation time. For example rolling snapshots on Mainnet take about 8 hours, so any value less than the artifact generation time for a given history mode and network will be ignored. Think of this value as more of a "floor" value.  **No artifact will take less time than the configured time value.**

Eventually we would like this time value to reflect the total amount of time (including artifact generation time), but this will require some complex calculations and record keeping of generation time.  Another caveat would be a fresh deployment "guessing" what the first amount of time will be.  For now we have this simple delay implemented.

### DIFF
--- a/charts/snapshotEngine/templates/configmap.yaml
+++ b/charts/snapshotEngine/templates/configmap.yaml
@@ -10,6 +10,8 @@ data:
   DISABLE_STORAGE_INIT_TIME: {{ $.Values.disableStorageInitTime | quote }}
   NETWORK_OVERRIDE: {{ $.Values.networkOverride | default "" | quote }}
   ALL_SUBDOMAINS: {{ $.Values.allSubdomains }}
+  ARCHIVE_SLEEP_DELAY: {{ $.Values.artifactDelay.archive }}
+  ROLLING_SLEEP_DELAY: {{ $.Values.artifactDelay.rolling }}
 kind: ConfigMap
 metadata:
   name: snapshot-configmap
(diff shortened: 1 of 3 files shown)


nicolasochem (approved):
thanks, code looks good, if it works as expected on xtz-shots feel free to merge.

I didn't know that "sleep 1m" or "sleep 1h" was a thing...

orcutt989:
> thanks, code looks good, if it works as expected on xtz-shots feel free to merge.
> 
> I didn't know that "sleep 1m" or "sleep 1h" was a thing...

I know me neither. I wrote a whole function to process the integer and the unit, and then I saw a dump of the `sleep` usage and I was like...oh, I didnt even need to do this. 😂

PR #523, opened 2022-12-13
SnapshotEngine: move sleep to end of job

orcutt989:
This moves the configurable sleep that delays tezos artifacts (snapshots & tarballs) to the end of the job instead of the beginning. This allows us to see development changes without having to reconfigure the timeout.


nicolasochem (approved):
thanks, the change looks straightforward, approved

PR #524, opened 2022-12-15
MY_NODE_CLASS is not always in the environment.

elric1:
This fixes tezos-signer and activate-jobs run their config-generator without errors.


nicolasochem (approved):
thank you, I just stumbled upon this myself.

PR #525, opened 2022-12-17
Handle Metadata from xtz-shots

nicolasochem:
This introduces under the hood "magic" to download the right snapshot for you, when using default values.

There was a previous attempt #394, however since then we have made some progress in formalizing the metadata at xtz-shots.io, to the point where it makes sense to introduce these changes instead.

The metadata file https://xtz-shots.io/tezos-snapshots.json is intended to become some sort of standard across ecosystem snapshot providers.

It provides a flat list of snapshots across tezos networks, containing size in bytes and sha256sum.

We introduce a new value `snapshot_source` which is defaulting to the xtz-shots metadata URL. I hope marigold will provide one such file soon as well.

When this file is set, config-generator will download it, and find the right snapshot, based on the network name and the octez version. It will then pass it to the node pods for download.

We also introduce `prefer_tarballs` which will make the nodes download a tarball instead of a snapshot. This defaults to False to preserve today's behavior.

The variables `full_snapshot_url` `archive_tarball_url` etc... still exist. They used to have default set to mainnet URLs from xtz-shots, but they default to null now.

If you set these variables, they will override the metadata behavior that we are introducing. As before, you can't set a value for snapshot and tarball of same history mode at the same time, or helm will error.

When using the new method, two new behaviors are introduced:

1. snapshot-downloader verifies the free space on the filesystem, and does not attempt to download if it is insufficient
1. after downloading, it will check the sha256sum against the one in metadata.

Now, our "quick start" guide will become more compelling: with just one line of helm, it will be possible to start a mainnet node, a ghostnet node, or any network on xtz-shots.

Test:

These 4 values yaml are provided as reference.

I tested them all on limanet (because artifacst are small and fast).

When it's merged and released, the `tezos_k8s_images` value can go and only `node_config_network:chain_name` needs to be set.

```
/# starts a limanet rolling node, downloads a snapshot, verifies checksum and unpacks it
tezos_k8s_images:
  utils: ghcr.io/oxheadalpha/tezos-k8s-utils:snapshot_metadata
node_config_network:
  chain_name: limanet
```

```
/# starts a limanet rolling node, downloads a tarball, verifies checksum and unpacks it
tezos_k8s_images:
  utils: ghcr.io/oxheadalpha/tezos-k8s-utils:snapshot_metadata
node_config_network:
  chain_name: limanet
prefer_tarballs: true
```

```
/# verify that old method of downloading tarball still works
tezos_k8s_images:
  utils: ghcr.io/oxheadalpha/tezos-k8s-utils:snapshot_metadata
node_config_network:
  chain_name: limanet
rolling_tarball_url: https://limanet.xtz-shots.io/rolling-tarball
rolling_snapshot_url: null
```

```
/# verify that old method of downloading snapshot still works
tezos_k8s_images:
  utils: ghcr.io/oxheadalpha/tezos-k8s-utils:snapshot_metadata
node_config_network:
  chain_name: limanet
rolling_tarball_url: null
rolling_snapshot_url: https://limanet.xtz-shots.io/rolling
```

I also tried a private chain with mkchain with no parameters. it works

I am updating the python version to 3.10 in utils because I wanted to use python's new `case` statement. I tried 3.11 but pytezos is broken, I opened a ticket and referenced it in the source code.

### DIFF
--- a/charts/tezos/scripts/snapshot-importer.sh
+++ b/charts/tezos/scripts/snapshot-importer.sh
@@ -1,4 +1,4 @@
-set -ex
+set -e
 
 bin_dir="/usr/local/bin"
 data_dir="/var/tezos"
@@ -6,6 +6,14 @@ node_dir="$data_dir/node"
 node_data_dir="$node_dir/data"
 node="$bin_dir/tezos-node"
 snapshot_file=${node_dir}/chain.snapshot
+if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
+    block_hash=$(cat ${node_dir}/chain.snapshot.block_hash)
+fi
+
+if [ ! -f ${snapshot_file} ]; then
+    echo "No snapshot to import."
+    exit 0
+fi
 
 if [ -d ${node_data_dir}/context ]; then
     echo "Blockchain has already been imported. If a tarball"
@@ -16,8 +24,12 @@ fi
 
 cp -v /etc/tezos/config.json ${node_data_dir}
 
+if [ "${block_hash}" != "" ]; then
+  block_hash_arg="--block ${block_hash}"
+fi
+
 ${node} snapshot import ${snapshot_file} --data-dir ${node_data_dir} \
-    --network $CHAIN_NAME
+    --network $CHAIN_NAME ${block_hash_arg}
 find ${node_dir}
 
 rm -rvf ${snapshot_file}
--- a/charts/tezos/scripts/snapshot-importer.sh
+++ b/charts/tezos/scripts/snapshot-importer.sh
@@ -12,10 +12,6 @@ if [ ! -f ${snapshot_file} ]; then
     exit 0
 fi
 
-if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
-    block_hash=$(cat ${node_dir}/chain.snapshot.block_hash)
-fi
-
 if [ -d ${node_data_dir}/context ]; then
     echo "Blockchain has already been imported. If a tarball"
     echo "instead of a regular tezos snapshot was used, it was"
@@ -25,8 +21,8 @@ fi
 
 cp -v /etc/tezos/config.json ${node_data_dir}
 
-if [ "${block_hash}" != "" ]; then
-  block_hash_arg="--block ${block_hash}"
+if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
+    block_hash_arg="--block $(cat ${node_dir}/chain.snapshot.block_hash)"
 fi
 
 ${node} snapshot import ${snapshot_file} --data-dir ${node_data_dir} \
--- a/charts/tezos/scripts/snapshot-importer.sh
+++ b/charts/tezos/scripts/snapshot-importer.sh
@@ -6,13 +6,14 @@ node_dir="$data_dir/node"
 node_data_dir="$node_dir/data"
 node="$bin_dir/octez-node"
 snapshot_file=${node_dir}/chain.snapshot
-if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
-    block_hash=$(cat ${node_dir}/chain.snapshot.block_hash)
-fi
 
 if [ ! -f ${snapshot_file} ]; then
     echo "No snapshot to import."
     exit 0
+fi
+
+if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
+    block_hash=$(cat ${node_dir}/chain.snapshot.block_hash)
 fi
 
 if [ -d ${node_data_dir}/context ]; then
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -16,6 +16,9 @@ data:
   ROLLING_SNAPSHOT_URL: "{{ .Values.rolling_snapshot_url }}"
   ROLLING_TARBALL_URL: "{{ .Values.rolling_tarball_url }}"
   ARCHIVE_TARBALL_URL: "{{ .Values.archive_tarball_url }}"
+  PREFER_TARBALLS: "{{ .Values.prefer_tarballs }}"
+  SNAPSHOT_SOURCE: "{{ .Values.snapshot_source }}"
+  OCTEZ_VERSION: "{{ .Values.images.octez }}"
   NODE_GLOBALS: |
 {{ .Values.node_globals | mustToPrettyJson | indent 4 }}
 
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -287,16 +287,15 @@ snapshot_source: https://xtz-shots.io/tezos-snapshots.json
 prefer_tarballs: false
 
 # By default, tezos-k8s will attempt to download the right artifact from
-# `snapshot_source` set above.
-# You can override and hard-code a snapshot URL source below.
-# When any of the below variables is set, `snapshot_source` above will be
-# ignored for all artifact types.
+# `snapshot_source` set above. You can override and hard-code a snapshot URL
+# source below. When any of the below variables are set, `snapshot_source` above
+# will be ignored for all artifact types.
 ## NOTE: `*_tarball_url` and `*_snapshot_url` are mutually exclusive
 ## and cannot both be specified at the same time.
+archive_tarball_url: null # e.g. https://mainnet.xtz-shots.io/archive-tarball
 full_snapshot_url: null
 full_tarball_url: null
 rolling_snapshot_url: null
-archive_tarball_url: null
 rolling_tarball_url: null
 
 # List of peers for nodes to connect to. Gets set under config.json `p2p` field
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -273,10 +273,10 @@ signers: {}
 # ```
 # End Signers
 
-# When spinning up nodes, tezos-k8s will attempt to download a snapshot
-# from a known source.
-# If you want to sync from scratch, set to 'null'
-# For a private chain, set to 'null'
+# When spinning up nodes, tezos-k8s will attempt to download a snapshot from a
+# known source. This should be a url to a json metadata file in the format
+# xtz-shots uses. If you want to sync from scratch or for a private chain, set
+# to `null`.
 snapshot_source: https://xtz-shots.io/tezos-snapshots.json
 
 # By default, tezos-k8s will download and unpack snapshots.
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -265,17 +265,31 @@ signers: {}
 # ```
 # End Signers
 
-## Where full and rolling history mode nodes will get their Tezos snapshots from.
-full_snapshot_url: https://mainnet.xtz-shots.io/full
-rolling_snapshot_url: https://mainnet.xtz-shots.io/rolling
+# When spinning up nodes, tezos-k8s will attempt to download a snapshot
+# from a known source.
+# If you want to sync from scratch, set to 'null'
+# For a private chain, set to 'null'
+snapshot_source: https://xtz-shots.io/tezos-snapshots.json
 
-## Alternatively to a Tezos snapshot, you can download an LZ4-compressed
-## filesystem tar of a node's data directory by setting `archive_tarball_url`
-## and `rolling_tarball_url` fields to the URL of the file. NOTE:
-## `rolling_tarball_url` and `rolling_snapshot_url` are mutually exclusive and
-## cannot both be specified at the same time.
-archive_tarball_url: https://mainnet.xtz-shots.io/archive-tarball
-# rolling_tarball_url: "https://mainnet.xtz-shots.io/rolling-tarball
+# By default, tezos-k8s will download and unpack snapshots.
+# A tarball is a LZ4-compressed filesystem tar of a node's data directory.
+# You must trust the tarball provider to provide good data, as no check is
+# performed by the node.
+# If you prefer tarballs, set to "true" below.
+prefer_tarballs: false
+
+# By default, tezos-k8s will attempt to download the right artifact from
+# `snapshot_source` set above.
+# You can override and hard-code a snapshot URL source below.
+# When any of the below variables is set, `snapshot_source` above will be
+# ignored for all artifact types.
+## NOTE: `*_tarball_url` and `*_snapshot_url` are mutually exclusive
+## and cannot both be specified at the same time.
+full_snapshot_url: null
+full_tarball_url: null
+rolling_snapshot_url: null
+archive_tarball_url: null
+rolling_tarball_url: null
 
 # List of peers for nodes to connect to. Gets set under config.json `p2p` field
 bootstrap_peers: []
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -626,7 +626,7 @@ def create_node_snapshot_config_json(history_mode):
     """Create this node's snapshot config"""
 
     network_name = NETWORK_CONFIG.get("chain_name")
-    prefer_tarballs = os.environ["PREFER_TARBALLS"] == True
+    prefer_tarballs = os.environ["PREFER_TARBALLS"].lower() in ('true', '1', 't')
     artifact_type = "tarball" if prefer_tarballs else "tezos-snapshot"
     rolling_tarball_url = os.environ["ROLLING_TARBALL_URL"]
     full_tarball_url = os.environ["FULL_TARBALL_URL"]
--- a/utils/config-generator.py
+++ b/utils/config-generator.py
@@ -665,7 +665,10 @@ def create_node_snapshot_config_json(history_mode):
                 print(f"Error: history mode {history_mode} is not known.")
                 sys.exit(1)
 
-    octez_container_version = os.environ["OCTEZ_VERSION"]
+    if "images" in MY_POD_CLASS and "octez" in MY_POD_CLASS["images"]:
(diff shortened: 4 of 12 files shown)


utils/snapshot-downloader.sh
```diff
@@ -19,48 +19,55 @@ fi
...
+      echo "Error: not enough disk space available to download artifact." >&2
+      exit 1
+    else
+      echo "There is sufficient free space to download the artifact of size ${filesize_bytes}." >&2
+    fi
+  fi
+  curl -LfsS $1 | tee >(sha256sum > ${snapshot_file}.sha256sum)
+  if [ ! -z "${sha256}" ]; then
+    if [ "${sha256}" != "$(cat ${snapshot_file}.sha256sum | head -c 64)" ]; then
+      echo "Error: sha256 checksum of the downloaded file did not match checksum from metadata file." >&2
```
harryttd, line 53: I may not be correct but just from looking at this code my understanding is that the snapshot file will still exist on the machine even if the SHA doesn't match and we exit. The container would then restart. 

If we are downloading a tezos-snapshot, i think it would just try to download again even though it was previously downloaded and exists on the machine. If it is a tarball then the node data dir is imported here and not in the snapshot-importer container. So this container would restart and `if [ -d "$node_data_dir/context" ]; then` at the top of this file would return true and would exit the container cleanly.

charts/tezos/scripts/snapshot-importer.sh
```diff
@@ -1,11 +1,19 @@
...
 node="$bin_dir/tezos-node"
 snapshot_file=${node_dir}/chain.snapshot
+if [ -f ${node_dir}/chain.snapshot.block_hash ]; then
+    block_hash=$(cat ${node_dir}/chain.snapshot.block_hash)
+fi
+
+if [ ! -f ${snapshot_file} ]; then
+    echo "No snapshot to import."
+    exit 0
+fi
```
harryttd, line 16: should this be the first `if` done?

harryttd (changes requested):

charts/tezos/values.yaml
```diff
@@ -265,17 +265,31 @@ signers: {}
...
+# You can override and hard-code a snapshot URL source below.
+# When any of the below variables is set, `snapshot_source` above will be
+# ignored for all artifact types.
+## NOTE: `*_tarball_url` and `*_snapshot_url` are mutually exclusive
+## and cannot both be specified at the same time.
+full_snapshot_url: null
+full_tarball_url: null
+rolling_snapshot_url: null
+archive_tarball_url: null
+rolling_tarball_url: null
```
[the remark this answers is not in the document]
nicolasochem: The behavior you are suggesting is more confusing to the user, than what I have currently implemented.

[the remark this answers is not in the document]
nicolasochem: Also the vast majority of users will only ever need one artifact type per chart.

utils/config-generator.py
```diff
@@ -600,5 +621,91 @@ def create_node_config_json(
...
+                    return {"url": full_snapshot_url, "artifact_type": "tezos-snapshot"}
+                return {}
+            case "archive":
+                if archive_tarball_url:
+                    return {"url": archive_tarball_url, "artifact_type": "tarball"}
+                return {}
+            case _:
+                return {}
+
+    octez_container_version = os.environ["OCTEZ_VERSION"]
```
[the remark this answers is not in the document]
nicolasochem: you are right, and I wasn't aware that you could override the version in sts level field `images`.

In this case, I'd rather fix it separately? I think OCTEZ_VERSION env var should be set by helm to the true image of the pod. This requires a change in `_containers.tpl`, where `OCTEZ_VERSION` would not be coming from the configmap, instead it will be set directly in the pod by helm. @harryttd

utils/config-generator.py
```diff
@@ -600,5 +621,91 @@ def create_node_config_json(
...
+    )
+    # find snapshot matching all the requested fields
+    try:
+        matching_snapshots = [
+            s
+            for s in all_snapshots
+            if s["history_mode"] == history_mode
+            and s["artifact_type"] == artifact_type
+            and s["chain_name"] == network_name
+            and octez_version in s["tezos_version"]
```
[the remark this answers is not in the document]
nicolasochem: Thanks. If `octez_version` is none, that means that the regex failed to find `vNN` in the version field. That could mean that we are running an unreleased container (as is the case in mondaynet etc). At this point, `octez_verison in s["tezos_version"]` will always be False so it will loop over the whole list needlessly. But I think in this case it should still download an artifact. Therefore, I will be changing this code, so when `octez_version` is None, it won't try to match by version.

utils/config-generator.py
```diff
@@ -600,5 +621,91 @@ def create_node_config_json(
...
+                    return {"url": full_tarball_url, "artifact_type": "tarball"}
+                elif full_snapshot_url:
+                    return {"url": full_snapshot_url, "artifact_type": "tezos-snapshot"}
+                return {}
+            case "archive":
+                if archive_tarball_url:
+                    return {"url": archive_tarball_url, "artifact_type": "tarball"}
+                return {}
+            case _:
+                return {}
```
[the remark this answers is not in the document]
nicolasochem: ok, I am now throwing an error in this case

charts/tezos/scripts/snapshot-importer.sh
nicolasochem: yes

charts/tezos/scripts/snapshot-importer.sh
```diff
@@ -16,8 +25,12 @@ fi
 
 cp -v /etc/tezos/config.json ${node_data_dir}
 
+if [ "${block_hash}" != "" ]; then
+  block_hash_arg="--block ${block_hash}"
+fi
```
harryttd, line 30: could this be in the check above `if [ -f ${node_dir}/chain.snapshot.block_hash ]; then`?

utils/config-generator.py
```diff
@@ -600,5 +622,93 @@ def create_node_config_json(
...
+        f"""
+Searching for snapshots from {snapshot_source}
+with history mode {history_mode}
+and artifact type {artifact_type}
+and chain name {network_name}
+and octez version {octez_version}.
+    """
+    )
+    # find snapshot matching all the requested fields
+    try:
```
harryttd, line 695: Do we need to run this if no snapshot_src was found? We already know that no ovrride for a specific hist mode was set. And so if no snap_src was set either we don't need to loop through anything

I also don't see why we need to wrap this all in a try/except. If we use `.get()`, no lookups will fail. When we check for the `-1` snapshot we can default to null and just return the var.

utils/config-generator.py
```diff
@@ -600,5 +622,93 @@ def create_node_config_json(
...
+    """Create this node's snapshot config"""
+
+    network_name = NETWORK_CONFIG.get("chain_name")
+    prefer_tarballs = os.environ["PREFER_TARBALLS"].lower() in ('true', '1', 't')
+    artifact_type = "tarball" if prefer_tarballs else "tezos-snapshot"
+    rolling_tarball_url = os.environ["ROLLING_TARBALL_URL"]
+    full_tarball_url = os.environ["FULL_TARBALL_URL"]
+    archive_tarball_url = os.environ["ARCHIVE_TARBALL_URL"]
+    rolling_snapshot_url = os.environ["ROLLING_SNAPSHOT_URL"]
+    full_snapshot_url = os.environ["FULL_SNAPSHOT_URL"]
```
harryttd, line 635: I'm thinking we might need .get() instead of just array access, similar to looking up chain_name. Bec if the vars are not found bec they were commented out in the config then we will error out.

utils/snapshot-downloader.sh
nicolasochem: you are right; in fact, it was not even erroring out. The exit statement was caught in the pipe and the program was showing success. I have now fixed this, when a tarball with wrong sha256sum is downloaded, it flushes everything then tries again.

utils/config-generator.py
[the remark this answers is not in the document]
nicolasochem: done, thanks for noticing this

charts/tezos/templates/configs.yaml
```diff
@@ -16,6 +16,9 @@ data:
   ROLLING_SNAPSHOT_URL: "{{ .Values.rolling_snapshot_url }}"
   ROLLING_TARBALL_URL: "{{ .Values.rolling_tarball_url }}"
   ARCHIVE_TARBALL_URL: "{{ .Values.archive_tarball_url }}"
+  PREFER_TARBALLS: "{{ .Values.prefer_tarballs }}"
```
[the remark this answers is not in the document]
nicolasochem: You mean the quotes? when I remove them here's what helm gives me

```
Error: INSTALLATION FAILED: ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal bool into Go struct field ConfigMap.data of type string
```

utils/config-generator.py
nicolasochem: If no snapshot_src was found, we return earlier, so this code is not accessed.

I removed the try/except and did as you said.

charts/tezos/templates/configs.yaml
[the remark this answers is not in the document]
harryttd: ok makes sense. Especially as you updated the code in config-gen to handle the str into a bool

utils/config-generator.py
```diff
@@ -600,5 +622,92 @@ def create_node_config_json(
     return node_config
 
 
+def create_node_snapshot_config_json(history_mode):
+    """Create this node's snapshot config"""
+
+    network_name = NETWORK_CONFIG.get("chain_name")
+    prefer_tarballs = os.environ["PREFER_TARBALLS"].lower() in ('true', '1', 't')
```
harryttd:
```suggestion
    prefer_tarballs = os.environ.get("PREFER_TARBALLS". "").lower() in ('true', '1', 't')
```

utils/snapshot-downloader.sh
```diff
@@ -19,48 +19,71 @@ fi
...
+  # * checks that there is enough space to download the file
+  # * verifies the sha256sum
+  filesize_bytes=$(cat ${data_dir}/snapshot_config.json | jq -r '.filesize_bytes // empty')
+  sha256=$(cat ${data_dir}/snapshot_config.json | jq -r '.sha256 // empty')
+  if [ ! -z "${filesize_bytes}" ]; then
+    free_space=$(findmnt -bno size -T ${data_dir})
+    echo "Free space available in filesystem: ${free_space}" >&2
+    if [ "${filesize_bytes}" -gt "${free_space}" ]; then
+      echo "Error: not enough disk space available to download artifact." >&2
+      touch ${data_dir}/disk_space_failed
```
harryttd, line 45: do we need to return from this function so that it doesn't continue to run if there isn't enough disk space?

utils/snapshot-downloader.sh
```diff
@@ -19,48 +19,71 @@ fi
...
+    else
+      echo "There is sufficient free space to download the artifact of size ${filesize_bytes}." >&2
+    fi
+  fi
+  curl -LfsS $1 | tee >(sha256sum > ${snapshot_file}.sha256sum)
+  if [ ! -z "${sha256}" ]; then
+    if [ "${sha256}" != "$(cat ${snapshot_file}.sha256sum | head -c 64)" ]; then
+      echo "Error: sha256 checksum of the downloaded file did not match checksum from metadata file." >&2
+      touch ${data_dir}/sha256sum_failed
```
harryttd, line 54: we should prob explicitly return here too so the function doesn't continue to run

nicolasochem:
I retested all scenarios listed in PR description with `bf43020`, it still works. I also tried with latest master branch build, but it fails for a separate reason. Please re-review @harryttd

utils/snapshot-downloader.sh
```diff
@@ -19,48 +19,71 @@ fi
...
-elif [ -n "$tarball_url" ]; then
-  echo "Downloading and extracting tarball from $tarball_url"
-  curl -LfsS "$tarball_url" | lz4 -d | tar -x -C "$data_dir"
+  block_hash=$(cat ${data_dir}/snapshot_config.json | jq -r '.block_hash // empty')
+  download "$artifact_url" > "$snapshot_file"
+  if [ -f "${data_dir}/sha256sum_failed" ]; then
+    # sha256 failure
+    rm -rvf "${data_dir}/sha256sum_failed"
+    exit 1
+  fi
```
harryttd, line 70: does the snapshot file also needs to be deleted?

utils/snapshot-downloader.sh
harryttd: Is there anything we can do to stop the process from continuously trying to download the same file when the hash doesn't match?

utils/config-generator.py
```diff
@@ -600,5 +622,93 @@ def create_node_config_json(
...
+        matching_snapshots = [
+            s
+            for s in all_snapshots
+            if s["history_mode"] == history_mode
+            and s["artifact_type"] == artifact_type
+            and s["chain_name"] == network_name
+        ]
+        if octez_version:
+            matching_snapshots = [ s for s in matching_snapshots if  octez_version in s["tezos_version"] ]
+        matching_snapshots = sorted(matching_snapshots, key=lambda d: d["block_height"])
```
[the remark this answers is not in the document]
harryttd: I think all of the prop lookups need `.get()`, not just block_height. If the aritifact json file is missing any field it will cause the script to crash.

utils/snapshot-downloader.sh
nicolasochem: no, it will be overwritten later

utils/snapshot-downloader.sh
nicolasochem: right. it was the case, but I removed it during previous review round. re-added

utils/config-generator.py
harryttd: I think this still needs one more thing: `prefer_tarballs = os.environ.get("PREFER_TARBALLS", "").lower() in ("true", "1", "t")
It needs to default to "", otherwise it will try .lower() on None type

utils/snapshot-downloader.sh
nicolasochem: not needed, that's the end of the function

utils/snapshot-downloader.sh
harryttd: why not delete it though? Allow the current process to clean up after itself and not leave anything hanging from it?

nicolasochem: done

nicolasochem:
> Is there anything we can do to stop the process from continuously trying to download the same file when the hash doesn't match?

no, but you could argue it's the fault of the snapshot website so it's fine.

utils/config-generator.py
```diff
@@ -600,5 +622,94 @@ def create_node_config_json(
...
+    matching_snapshots = [
+        s
+        for s in all_snapshots
+        if s.get("history_mode") == history_mode
+        and s.get("artifact_type") == artifact_type
+        and s.get("chain_name") == network_name
+    ]
+    if octez_version:
+        matching_snapshots = [
+            s for s in matching_snapshots if octez_version in s.get("tezos_version")
```
harryttd, line 707: If tezos_version doesn't exist it will be trying to look up a str in Nonetype
```suggestion
            s for s in matching_snapshots if octez_version in s.get("tezos_version", "")
```

utils/snapshot-downloader.sh
harryttd: should we log the size of the file for visibility so they can easily tell how much more space is required? I don't think they'll know easily offhand bec config gen does a lot of sorting to pick the correct config. So it isn't easy to determine the size of the file chosen to download.

utils/snapshot-downloader.sh
harryttd: true although if we continue to add to add logic to this function we may not realize that we didn't return from that logic block.

utils/snapshot-downloader.sh
nicolasochem: done

utils/snapshot-downloader.sh
nicolasochem: done

nicolasochem:
I retested all scenarios listed in PR description with [0a98d1b](https://github.com/oxheadalpha/tezos-k8s/pull/525/commits/0a98d1bd970e38a98ebd653f934198bc8e3d7097), it still works. Please re-review @harryttd

utils/config-generator.py
```diff
@@ -141,17 +143,29 @@ def main():
...
         print("Generated config.json :")
-        print(config_json)
+        print(node_config_json)
+        print("Generated snapshot_config.json :")
+        print(node_snapshot_config_json)
         with open("/etc/tezos/config.json", "w") as json_file:
-            print(config_json, file=json_file)
+            print(node_config_json, file=json_file)
+        with open("/var/tezos/snapshot_config.json", "w") as json_file:
+            print(node_snapshot_config_json, file=json_file)
```
harryttd, line 168: Perhaps we shouldn't even be creating this file if the snap config function returns None. Then `if [ "$(cat ${data_dir}/snapshot_config.json)" == "null" ];` in the downloader needs to check if the file exists. Instead of assuming it always exists with the possibility of the file being created without containing any data.

harryttd (approved):
I haven't tested only reviewed the code but i'm assuming it all works now.

nicolasochem:
Yes, I just retested everything on the very latest commit, it's fine. I also tried spinning up a rolling node while only setting the archive tarball url, to verify that the file is not being created.

PR #526, opened 2022-12-20
Add rpc service port for tezos node

puppetninja:
Currently rpc port is export from a NodePort service with a selector, this is not that
flexible in some use scenarios. 

This commit exports the rpc port for tezos node service with ClusterIP, this allows more flexible service route for k8s ingress, e.g. backend rpc services can be directed based on uri path to different tezos node services.

### DIFF
--- a/charts/tezos/templates/static.yaml
+++ b/charts/tezos/templates/static.yaml
@@ -22,6 +22,8 @@ metadata:
     appType: tezos-node
 spec:
   ports:
+    - port: 8732
+      name: rpc
     - port: 9732
       name: p2p
     - port: 9932
--- a/test/charts/mainnet.expect.yaml
+++ b/test/charts/mainnet.expect.yaml
@@ -89,6 +89,8 @@ metadata:
     appType: tezos-node
 spec:
   ports:
+    - port: 8732
+      name: rpc
     - port: 9732
       name: p2p
     - port: 9932
--- a/test/charts/mainnet2.expect.yaml
+++ b/test/charts/mainnet2.expect.yaml
@@ -128,6 +128,8 @@ metadata:
     appType: tezos-node
 spec:
   ports:
+    - port: 8732
+      name: rpc
     - port: 9732
       name: p2p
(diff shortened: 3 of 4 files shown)


harryttd:
Thanks for this PR.

I don't see any problem with implementing your change, but just for my understanding, can you elaborate on what you were trying to do that this solves?

FYI this is not a ClusterIP service. It is a headless service as the ClusterIP field is set to `None`. Headless service gives each pod their own dns record that we can target individually. As long as the container port is exposed, which it is in _containers.tpl, it can be targeted. Honestly I'm not even sure we really need to specify any port at all on a headless service, but see https://stackoverflow.com/questions/51196903/do-we-really-need-port-for-a-headless-service.

puppetninja:
@harryttd My use scenario is that I am using one ingress to route different rpc traiffic with different URI path to different types of nodes, like rolling nodes and archive nodes. so targeting each pod is what I need, but as the rpc port is not forwarded so I am adding one here.

orcutt989:
Looking at our currently deployed nodes we have RPC being forwarded to a NodePort of 32006. A NodePort `Exposes the Service on each Node's IP at a static port (the NodePort). To make the node port available, Kubernetes sets up a cluster IP address, the same as if you had requested a Service of type: ClusterIP.`. https://kubernetes.io/docs/concepts/services-networking/service/ I believe this is the default config?

puppetninja:
Hi @harryttd @orcutt989 I will elaborate more about my scenario, so I deployed two types of nodes rolling and archive, which deployed two services here, and one nodeport service

file:///home/chao/Pictures/Screenshots/Screenshot%20from%202022-12-21%2017-49-36.png

orcutt989:
Sounds good we discussed it with @nicolasochem and he explained also. @harryttd should be around with his thoughts and an approval soon. Thanks again for contributing! 🙏

orcutt989 (approved):

harryttd (approved):

PR #528, opened 2022-12-21
v15+lima default for tezos-k8s

nicolasochem:
* default mkchain chain is a lima chain
* examples in values.yaml show lima activation
* move default version of octez to v15
* rename tezos-node and -client to octez-node and -client across the board
* update tests to remove some old tezedge validation

I did test mkchain with no parameters and it activates properly.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -8,7 +8,7 @@ is_invitation: false
 
 # Images not part of the tezos-k8s repo go here
 images:
-  octez: tezos/tezos:v14-release
+  octez: tezos/tezos:v15-release
 # Images that are part of the tezos-k8s repo go here with 'dev' tag
 tezos_k8s_images:
   utils: ghcr.io/oxheadalpha/tezos-k8s-utils:master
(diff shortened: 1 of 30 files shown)


charts/tezos/values.yaml
```diff
@@ -327,7 +327,7 @@ protocols:
   ##   "on" and "off" must be between quotes.
   ## Note that we are not providing default votes since every baker needs
   ## to make an explicit educated choice on every toggle.
-  - command: 014-PtKathma
+  - command: PtLimaPt
```
harryttd, line 330: i'm assuming this doesnt need the number suffix like it used to

nicolasochem: Yes, they changed it in v15 to prevent confusion between protocol version and octez version.

mkchain/tqchain/keys.py
```diff
@@ -57,9 +57,9 @@ def gen_key(image):
         image,
         "sh",
         "-c",
-        "'/usr/local/bin/tezos-client "
+        "'/usr/local/bin/octez-client "
         + "--protocol PsDELPH1Kxsx gen keys mykey && "
```
harryttd, line 61: not related to this PR but i'm curious why we never updated this to a later proto

nicolasochem: no need, this is just to generate keys and the key format has never changed

orcutt989 (approved):

harryttd (approved):

PR #531, opened 2023-01-12
Exit config-init script on error

harryttd:
We ran into a problem where config-init exited cleanly after running `octez-node config init...` and failed with `(/usr/local/bin/octez-node) Exit: exit because of uncaught exception: Failure("resolution failed: name resolution failed")`. 

`teztnets.xyz` as part of the `--network` flag failed to resolve. The container then moved on to config-generator which failed bec it couldn't find the config.json file as it didn't get written.

The config-init script should exit on error so that k8s retries the container. In the case where the network url didn't resolve at first, it may on a subsequent try.

### DIFF
--- a/charts/tezos/scripts/config-init.sh
+++ b/charts/tezos/scripts/config-init.sh
@@ -1,8 +1,7 @@
+set -e
+
 echo "Writing custom configuration for public node"
 mkdir -p /etc/tezos/data
-
-#
-# This is my comment.
 
 /usr/local/bin/octez-node config init		\
     --config-file /etc/tezos/data/config.json	\
--- a/test/charts/mainnet.expect.yaml
+++ b/test/charts/mainnet.expect.yaml
@@ -216,11 +216,10 @@ spec:
           args:
             - "-c"
             - |
+              set -e
+              
               echo "Writing custom configuration for public node"
               mkdir -p /etc/tezos/data
-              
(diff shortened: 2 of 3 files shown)


nicolasochem:
Did you try on a local container? I don't see why `set -e` would fix things, because previously, if the network was not downloaded, it would fail to write a config so the `cat` command would exit in error, and therefore the init container itself would error out? See this error you pasted on slack:

```
│ Writing custom configuration for public node                                                                         │
│ Warning                                                                                                              │
│   The executable with name tezos-node has been renamed to octez-node. The name tezos-node is now                     │
│   deprecated, and it will be removed in a future release. Please update                                              │
│   your scripts to use the new name.                                                                                  │
│                                                                                                                      │
│ (/usr/local/bin/octez-node) Exit: exit because of uncaught exception: Failure("TLS to non-TCP currently unsupported: │
│                                                                                                                                                                                                                                                                                                                                                     │
│ cat: can't open '/etc/tezos/data/config.json': No such file or directory
```

Do we understand why the container succeeds with the code on master?

harryttd:
I tested locally. It will fail the script for any cmd that exists with a non 0 error code. In this case the config init cmd exits with 126 which is an error code indicating a failure internally inside an application.

nicolasochem (approved):
thank you!

PR #532, opened 2023-01-18
tezos-proto-cruncher chart

nicolasochem:
This chart deploys a daemonset in your cluster. the daemonset runs a parallel search for a vanity protocol hash.

It takes s3 bucket credentials. Protocol is downloaded from bucket then results are written to the bucket.

It does not terminate: it keeps running until the cluster is torn down.

### DIFF
--- a/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
+++ b/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
@@ -16,14 +16,14 @@ proto_prefix = tb([2, 170])
 PROTO_NAME = os.getenv("PROTO_NAME")
 VANITY_STRING = os.getenv("VANITY_STRING")
 BUCKET_NAME = os.getenv("BUCKET_NAME")
-ENDPOINT_URL = os.getenv("HOST_BASE")
-BUCKET_REGION_NAME = os.getenv("REGION_NAME")
+BUCKET_ENDPOINT_URL = os.getenv("BUCKET_ENDPOINT_URL")
+BUCKET_REGION = os.getenv("BUCKET_REGION")
 
 if BUCKET_NAME:
     import boto3
     s3 = boto3.resource('s3',
-                        region_name=BUCKET_REGION_NAME,
-                        endpoint_url=f"https://{ENDPOINT_URL}")
+                        region_name=BUCKET_REGION,
+                        endpoint_url=f"https://{BUCKET_ENDPOINT_URL}")
 
 with open(proto_file, "rb") as f:
     proto_bytes = f.read()
--- a/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
+++ b/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
@@ -9,8 +9,11 @@ import re
 
 proto_file = sys.argv[1]
 
+
 def tb(l):
-    return b''.join(map(lambda x: x.to_bytes(1, 'big'), l))
+    return b"".join(map(lambda x: x.to_bytes(1, "big"), l))
+
+
 proto_prefix = tb([2, 170])
 
 PROTO_NAME = os.getenv("PROTO_NAME")
@@ -27,9 +30,10 @@ if not NUM_NONCE_DIGITS:
 
 if BUCKET_NAME:
     import boto3
-    s3 = boto3.resource('s3',
-                        region_name=BUCKET_REGION,
-                        endpoint_url=f"https://{BUCKET_ENDPOINT_URL}")
+
+    s3 = boto3.resource(
+        "s3", region_name=BUCKET_REGION, endpoint_url=f"https://{BUCKET_ENDPOINT_URL}"
+    )
 
 with open(proto_file, "rb") as f:
     proto_bytes = f.read()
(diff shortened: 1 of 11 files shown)


charts/tezos-proto-cruncher/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+# name of the bucket
+bucket_name: "tezos-proto-cruncher"
+
+# name of the protocol file in the bucket
+proto_name: "mumbai-f"
+#
+# vanity string we are looking for.
+# Must start with Pr, Ps, or Pt, but probably Pt as it is the convention.
+vanity_string: "PtMumbai"
```
harryttd, line 17: sorry... do you mind making these camel case to keep in line with the k8s convention of naming properties? The medium article would have to be updated too.

charts/tezos-proto-cruncher/templates/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: s3-env
+data:
+  AWS_ACCESS_KEY_ID: "{{ .Values.s3_access_key_id }}"
+  BUCKET_ENDPOINT_URL: "{{ .Values.bucket_endpoint_url }}"
+  BUCKET_NAME: "{{ .Values.bucket_name}}"
+  BUCKET_REGION: "{{ .Values.bucket_region}}"
```
harryttd, line 16: maybe some of these should be secrets as well? I would think at least the access key id would be preferably a secret. But whatever you think.

charts/tezos-proto-cruncher/scripts/proto-downloader.py
```diff
@@ -0,0 +1,17 @@
+import os
+
+BUCKET_NAME = os.getenv("BUCKET_NAME")
+BUCKET_ENDPOINT_URL = os.getenv("BUCKET_ENDPOINT_URL")
+BUCKET_REGION= os.getenv("BUCKET_REGION")
+
+PROTO_NAME = os.getenv("PROTO_NAME")
```
harryttd, line 7: Maybe fail fast via another getter if the env vars don't exist? Say endpoint doesn't exist. The url will be created with `None`. Might as well fail fast and the stack trace will clearly show why.

charts/tezos-proto-cruncher/scripts/tezos-proto-cruncher.sh
```diff
@@ -0,0 +1,8 @@
+apk add parallel
+pip install boto3
+
+python /proto-downloader.py
+
+# Launch one process per CPU core to maximize utilization
+num_cores=$(grep -c ^processor /proc/cpuinfo)
```
harryttd, line 7: perhaps should be configurable how many cores to run on? If the user only has minimal nodes and or node resources, this could really slow down their other workloads.

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
```diff
@@ -0,0 +1,60 @@
...
+def get_hash(vanity_nonce, proto_hash):
+    proto_hash.update(vanity_nonce)
+    return base58.b58encode_check(proto_prefix + proto_hash.digest()).decode('utf-8')
+
+print(f"Original proto nonce: {original_nonce} and hash: {get_hash(original_nonce, proto_hash.copy())}")
+
+while True:
+    # Warning - assuming the nonce is 16 chars in the original proto.
+    # If it is not, set the current number below, otherwise you will get bad nonces.
+    new_nonce_digits = ''.join(random.choice(string.digits) for _ in range(16))
```
harryttd, line 53: how is the user supposed to change this? Shouldn't it be configurable via values.y?

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
```diff
@@ -0,0 +1,60 @@
...
+    new_nonce = b'(* Vanity nonce: ' + bytes(new_nonce_digits, 'utf-8') + b' *)\n'
+
+    new_hash = get_hash(new_nonce, proto_hash.copy())
+    if re.match(f"^{VANITY_STRING}.*", new_hash):
+        print(f"Found vanity nonce: {new_nonce} and hash: {new_hash}")
+        if BUCKET_NAME:
+            s3.Object(BUCKET_NAME, f"{PROTO_NAME}_{new_hash}").put(Body=new_nonce)
```
harryttd, line 60: error handle? If this returns an error the script would exit it seems. Even though k8s should restart the container, depending upon the error type, perhaps the `while True` should continue running as the error may only be temporary. Like a network error or something.

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
```diff
@@ -0,0 +1,60 @@
...
+
+def tb(l):
+    return b''.join(map(lambda x: x.to_bytes(1, 'big'), l))
+proto_prefix = tb([2, 170])
+
+PROTO_NAME = os.getenv("PROTO_NAME")
+VANITY_STRING = os.getenv("VANITY_STRING")
+BUCKET_NAME = os.getenv("BUCKET_NAME")
+BUCKET_ENDPOINT_URL = os.getenv("BUCKET_ENDPOINT_URL")
+BUCKET_REGION = os.getenv("BUCKET_REGION")
```
harryttd, line 20: fail fast if vars not set?

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
```diff
@@ -0,0 +1,60 @@
...
+while True:
+    # Warning - assuming the nonce is 16 chars in the original proto.
+    # If it is not, set the current number below, otherwise you will get bad nonces.
+    new_nonce_digits = ''.join(random.choice(string.digits) for _ in range(16))
+    new_nonce = b'(* Vanity nonce: ' + bytes(new_nonce_digits, 'utf-8') + b' *)\n'
+
+    new_hash = get_hash(new_nonce, proto_hash.copy())
+    if re.match(f"^{VANITY_STRING}.*", new_hash):
+        print(f"Found vanity nonce: {new_nonce} and hash: {new_hash}")
+        if BUCKET_NAME:
```
harryttd, line 59: Isn't the goal of the chart to upload hashes? So if no bucket name exists this script will loop endlessly and not store the results anywhere. Shouldn't it error? Or at least at the top of the file error if the env var isnt set.

harryttd (changes requested):

harryttd:
Suggestion: explain how to uninstall the chart in the medium article when it isn't needed anymore

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
nicolasochem: as per the medium article, users are free to download and run this script on their computer as well. So I'm not going to fail if the s3 variables are not set. For the `VANITY_STRING`, sure.

harryttd: ah, i see. very good

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
harryttd: since i forgot about running this script locally, i guess it could be an env var

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
nicolasochem: if no bucket name exists, I assume this is a local run and printing the nonces out in the console will be sufficient.

charts/tezos-proto-cruncher/templates/config.yaml
nicolasochem: I think only stuff explicitly designated as secret should be a k8s secret.

charts/tezos-proto-cruncher/scripts/proto-cruncher.py
```diff
@@ -0,0 +1,81 @@
...
+            try:
+                s3.Object(BUCKET_NAME, f"{PROTO_NAME}_{new_hash}").put(Body=new_nonce)
+            except:
+                print("ERROR: upload of the nonce and hash to s3 failed.")
+                continue
```
harryttd, line 81: Thanks for adding this. The thing is that if there is a type of error that will never resolve itself, hashes will continue to be generated pointlessly, using a lot of resources. I assume there is a way to differentiate between diff errors and  raise an exception if there is an unresolvable error.

nicolasochem: Not really. If hashes can't upload, it's still good to generate them. You can look at the logs of your pod and retrieve them.

harryttd (approved):

PR #534, opened 2023-01-21
Set fitness to 1 instead of -1 at activation

nicolasochem:
A bug was uncovered during mumbai activation, where nodes were unable to activate and were crashing.

The [fix](https://gitlab.com/tezos/tezos/-/merge_requests/7462/diffs) broke dailynet because it is no longer possible to activate with a fitness of -1 (which is how we have been activating chains since the beginning of tezos-k8s).

Replacing this value to 1 fixes the issue for dailynet and also does not appear to be causing any issue with v15.1.

### DIFF
--- a/charts/tezos/scripts/chain-initiator.sh
+++ b/charts/tezos/scripts/chain-initiator.sh
@@ -15,6 +15,6 @@ echo Activating chain:
 $CLIENT -d /var/tezos/client --block					\
 	genesis activate protocol					\
 	{{ .Values.activation.protocol_hash }}				\
-	with fitness -1 and key						\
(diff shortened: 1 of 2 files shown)


harryttd:
Whats the reason again we used to activate with -1? And does our code still function the same going forward?

nicolasochem:
It has always been -1, but this no longer works due to the way tenderbake
chains activate. There are better explanations in the comments that the
linked MR is adding to octez. Vbot from nomadic confirmed that it was an ok
value, and I've also tested Lima activation with fitness 1 and it worked.

harryttd (approved):

PR #536, opened 2023-01-30
Proto.cruncher.tweaks

itkach:
TL;DR generate proto hashes with vanity prefix on a multi-core laptop

Case-insensitive "PtMumbai" prefix:
- Apple Silicon (M1, 9 cores): ~1.5 vanity hashes per hour
- Intel (i9, 16 cores): ~1 vanity hash per hour

Speedups:
- Major speedup comes from using [based58](https://pypi.org/project/based58/) library instead of pure Python `base58` (3-6x depending on CPU)
- Python code optimizations (faster nonce generation, faster match verification)
- execute in multiprocess worker pool

Initial version of this PR  keeps the original implementation and includes several options to allow comparisons controlled by the following env variables:
`IMPL` - choose implementation, one of `multiproc`, `multithread`, `original`, `singlethread`. Run `original` vs `singlethread` to compare original implementation  vs optimized.
`FORCE_PY_BASE58` - set to 1 to force using pure python `base58` even if `based58` is installed
`NPROCS` - number of processes/threads for the worker pool to use in multiproc and multithread implementations, by default equals to the number of cores reported by OS

It looks like `multithread` is the worst performer (probably because of high contention for GIL), included only for reference. `original` is also included for reference, probably should be removed before merging this.

### DIFF
--- a/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
+++ b/charts/tezos-proto-cruncher/scripts/proto-cruncher.py
@@ -1,11 +1,31 @@
 import hashlib
-import random
-import base58
-import string
+import locale
 import os
-import sys
+import random
 import re
+import string
+import sys
 import time
+
+from multiprocessing import Pool
+from multiprocessing.pool import ThreadPool
+
+if os.getenv("FORCE_PY_BASE58"):
+    import base58
+else:
+    try:
+        import based58 as base58
+
+        print(f"Using based58 from {base58.__file__}")
+    except:
+        print(
+            """
+        !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+        Unable to import based58, will fall back to slow "base58"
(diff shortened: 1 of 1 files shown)


itkach:
I think it would be better to extract proto cruncher into a standalone tool/separate repo targeting primarily local execution (remove s3 interaction, add proper command line interface rather than depend on a bunch of environment variables, add proper python packaging)

nicolasochem:
> I think it would be better to extract proto cruncher into a standalone tool/separate repo targeting primarily local execution (remove s3 interaction, add proper command line interface rather than depend on a bunch of environment variables, add proper python packaging)

Thanks! Yes, probably, we already have isolated mkchain as a standalone tool, we could do the same for the proto cruncher and install it in the utils vm.

That's assuming the chart is still useful. At 1.5h, you don't need any help from the cloud :sunglasses:

itkach:
one more variable I forgot to mention, set `EXACT_MATCH` if you really want exact, case-sensitive prefix, of course it takes considerably longer to find (still not unreasonable - I found one in 22 hours on M1 while I was still using slow `base58`)

nicolasochem:
Exact match is useful when proposing amendments, because then it does not change the binary names (`octez-baker-PtMumbai`).

PR #537, opened 2023-02-01
BCD indexer updates

harryttd:
Working BCD indexer on their 4.3.0 release. They have a 4.4.0 release but I don't want to default to that version yet bec BCD right now is primarily for DMV. DMV is running older protocol right now that won't work with latest BCD release.


charts/tezos/templates/_helpers.tpl
```diff
@@ -112,11 +112,21 @@ metadata:
...
+  Has per-block votes defined?
+*/}}
+{{- define "tezos.hasPerBlockVotes" }}
+  {{- range .Values.protocols }}
+    {{- if .vote }}
+      {{- "true" }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+
```
nicolasochem, line 125: this looks like an unexpected change

nicolasochem (changes requested):
As far as I can tell, this PR:

1. adds bcd pod to the tezos chart (optionally)
2. lints all python files it can find with black
3. re-adds a helm section that's gone from master (per_block_votes)

3. looks unvoluntary, should it be removed?

charts/tezos/templates/_helpers.tpl
harryttd: it's actually visually confusing the diff. this block is moved from the bottom of the file to here. Before the indexers

nicolasochem (approved):

harryttd:
> As far as I can tell, this PR:
> 
> 1. adds bcd pod to the tezos chart (optionally)
> 2. lints all python files it can find with black
> 3. re-adds a helm section that's gone from master (per_block_votes)
> 4. looks unvoluntary, should it be removed?

yes to everything except 3 and 4. I responded on the inline comment

Issue #378, opened 2022-02-23
Should not chmod 777 /var/tezos 


harryttd:
In config-generator.sh we run:
https://github.com/oxheadalpha/tezos-k8s/blob/4f3488153c5b2f8990552c45eab46e832b33c061/utils/config-generator.sh#L8-L9

The node on its initial creation will create files/dirs with default perms set by the Tezos docker image:
```shell
~ $ ls -l /var/tezos/ /var/tezos/node /var/tezos/node/data/
/var/tezos/:
total 24
drwxrwxrwx    2 root     users         4096 Feb 23 15:18 client
drwxrwxrwx    2 root     users        16384 Feb 23 15:18 lost+found
drwxr-xr-x    3 tezos    nogroup       4096 Feb 23 15:18 node

/var/tezos/node:
total 4
drwx------    4 tezos    nogroup       4096 Feb 23 15:19 data

/var/tezos/node/data/:
total 20
drwxr-xr-x    3 tezos    nogroup       4096 Feb 23 15:19 context
-rw-r--r--    1 tezos    nogroup        300 Feb 23 15:19 identity.json
-rw-r--r--    1 tezos    nogroup          1 Feb 23 15:18 lock
drwxr-xr-x    4 tezos    nogroup       4096 Feb 23 15:19 store
-rw-r--r--    1 tezos    nogroup         22 Feb 23 15:18 version.json
~ $
```

Only the `/var/tezos/client` is `777` because we create it in the script and then immediately run `chmod 777`. The node data dir doesn't exist yet and hence the less privileged perms.

If the node never restarts, these permissions will remain. If the pod restarts, now that the data dir does exist the `chmod 777` will give unnecessary perms. It will also change the group from `nogroup` to `users`:
```shell
 ~ $ ls -l /var/tezos/ /var/tezos/node /var/tezos/node/data/
/var/tezos/:
total 24
drwxrwxrwx    2 root     users         4096 Feb  1 02:41 client
drwxrwxrwx    2 root     users        16384 Feb  1 02:41 lost+found
drwxrwxrwx    3 tezos    users         4096 Feb  1 02:41 node

/var/tezos/node:
total 4
drwxrwxrwx    4 tezos    users         4096 Feb 18 19:58 data

/var/tezos/node/data/:
total 1508
drwxrwxrwx    3 tezos    users         4096 Feb  1 02:41 context
-rwxrwxrwx    1 tezos    users          300 Feb  1 02:41 identity.json
[...]

PR #547, opened 2023-02-17
Persist identity and peers when using local storage

nicolasochem:
Local storage is great because it's faster, and we can download tarball in minutes at every boot and have tolerable downtimes.

But at each boot, the node regenerates its identity. We already have a mechanism to pre-populate identities in the node yaml, but it's cumbersome.

Instead, with this PR, we create a mini-pvc of 1Gi to store node identities, whenever local storage is in use.

This way, local storage nodes always boot quickly, except the first time they boot, and there is no configuration burden. It is also possible to scale up the number of nodes without having to prepare identities.

This works by creating dangling symlinks in the newly created data dir for both files identity.json and peers.json, targeting the small persistent dir whenever local storage is used. When the node boots, it follows these symlinks and creates files at the target. (I have actually not seen this happen for peers.json but it works for identity.json).

I tested private chain and ghostnet node, in both local and non-local storage. It behaves as intended. I also tried to pre-populate an identity in the values file in local storage and confirmed that it is not broken.

### DIFF
--- a/charts/tezos/scripts/upgrade-storage.sh
+++ b/charts/tezos/scripts/upgrade-storage.sh
@@ -1,6 +1,6 @@
 set -ex
 
-if [ ! -e /var/tezos/node/data/context ]
+if [ ! -e /var/tezos/node/data/context/store.dict ]
 then
   printf "No context in data dir found, probably initial start, doing nothing."
   exit 0
--- a/charts/tezos/scripts/upgrade-storage.sh
+++ b/charts/tezos/scripts/upgrade-storage.sh
@@ -2,7 +2,7 @@ set -ex
 
 if [ ! -e /var/tezos/node/data/context/store.dict ]
 then
-  printf "No context in data dir found, probably initial start, doing nothing."
+  printf "No store in data dir found, probably initial start, doing nothing."
   exit 0
 fi
 octez-node upgrade storage --config /etc/tezos/config.json
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -70,6 +70,8 @@ spec:
   volumeClaimTemplates:
     - metadata:
 {{- if $.node_vals.local_storage | default false }}
+        # since we are using local storage, deploy a small persistent voluem
+        # to store identity.json and peers.json
         name: persistent-volume
 {{- else }}
         name: var-volume
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -70,7 +70,7 @@ spec:
   volumeClaimTemplates:
     - metadata:
 {{- if $.node_vals.local_storage | default false }}
-        # since we are using local storage, deploy a small persistent voluem
+        # since we are using local storage, deploy a small persistent volume
         # to store identity.json and peers.json
         name: persistent-volume
 {{- else }}
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -66,10 +66,14 @@ spec:
 {{- if $.node_vals.local_storage | default false }}
         - emptyDir: {}
           name: var-volume
-{{- else }}
+{{- end }}
   volumeClaimTemplates:
     - metadata:
+{{- if $.node_vals.local_storage | default false }}
+        name: persistent-volume
+{{- else }}
         name: var-volume
+{{- end }}
         namespace: {{ $.Release.Namespace }}
       spec:
         accessModes:
@@ -79,6 +83,9 @@ spec:
         {{- end }}
         resources:
           requests:
+{{- if $.node_vals.local_storage | default false }}
+            storage: "1Gi"
+{{- else }}
             storage: {{ default "15Gi" $.node_vals.storage_size }}
 {{- end }}
 ---
--- a/utils/config-generator.sh
+++ b/utils/config-generator.sh
@@ -6,6 +6,7 @@ cat /etc/tezos/data/config.json
 echo ------------------------------------------------------------
 
 mkdir -p /var/tezos/client
+mkdir -p /var/tezos/node/data
 chmod -R 777 /var/tezos
 set -e
 python3 /config-generator.py "$@"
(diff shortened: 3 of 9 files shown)


utils/config-generator.sh
```diff
@@ -17,6 +18,17 @@ set +e
...
 
 MY_CLASS=$(echo $NODES | jq -r ".\"${MY_NODE_CLASS}\"")
+
+# Set up symlinks for local storage
+local_storage=$(echo $MY_CLASS | jq -r ".local_storage")
+if [ "${local_storage}" == "true" ]; then
+  echo '{ "version": "2.0" }' > /var/tezos/node/data/version.json
+  ln -s /var/persistent/peers.json /var/tezos/node/data/peers.json
+  ln -s /var/persistent/identity.json /var/tezos/node/data/identity.json
+  chmod -R 777 /var/tezos/node/data
```
harryttd, line 28: 777 is not secure and gives all users r/w and exec perms on everything. https://linuxize.com/post/what-does-chmod-777-mean/#never-use-chmod-777

It should likely be the tezos user and not all users.

utils/config-generator.sh
```diff
@@ -6,6 +6,7 @@ cat /etc/tezos/data/config.json
 echo ------------------------------------------------------------
 
 mkdir -p /var/tezos/client
+mkdir -p /var/tezos/node/data
 chmod -R 777 /var/tezos
```
harryttd, line 10: This chmod 777 should also be changed. Not secure

utils/config-generator.sh
harryttd: Also why is this needed if chmod is run at the top of the file?

nicolasochem:
@harryttd actually this does not even work. My goal here is to save peers.json and identity.json across restarts when using local storage, to prevent 2 minute identity generation at boot and hard dependency on bootstrap peers. But tezos storage is too inflexible... it won't tolerate any symlinks present in the data dir.

nicolasochem:
> @harryttd actually this does not even work. My goal here is to save peers.json and identity.json across restarts when using local storage, to prevent 2 minute identity generation at boot and hard dependency on bootstrap peers. But tezos storage is too inflexible... it won't tolerate any symlinks present in the data dir.

I take this back. It actually works. My problem is that /var/persistent was not mounted. When it is, octez tolerates a dangling symlink in the data dir and will actually write to it at first boot.

nicolasochem:
@harryttd regarding `chmod 777`, Data dir needs to be world writable because the user running utils is not the tezos user. Same thing for the persistent dir (when it exists). When local storage is used, I write `version.json` and create 2 symlinks so I do `chmod -R 777` again in order to set everything straight.

I'm sure that we tried before to have config-generator use the tezos user and it was not working for some reason. Hence the 777 was already here. It's my excuse. Do you want me to revisit?

harryttd:
> I take this back. It actually works. My problem is that /var/persistent was not mounted. When it is, octez tolerates a dangling symlink in the data dir and will actually write to it at first boot.

Very cool. Was this tested in the cloud? You never know what's gonna happen with EBS.

> Instead, with this PR, we create a mini-pvc of 1Gi to store node identities, whenever local storage is in use.

Just as an FYI to clarify (that rhymes), I believe the lowest EBS volume size is 8Gi. So even if we specify a volume in kilobytes, AWS will create an 8Gi volume.

> @harryttd regarding chmod 777, Data dir needs to be world writable because the user running utils is not the tezos user. Same thing for the persistent dir (when it exists). When local storage is used, I write version.json and create 2 symlinks so I do chmod -R 777 again in order to set everything straight.
>
> I'm sure that we tried before to have config-generator use the tezos user and it was not working for some reason. Hence the 777 was already here. It's my excuse. Do you want me to revisit?

The original `777` was from a while ago. I think @elric1 was working on some code. I don't recall what the context was. Either way `777` is not a good idea and should be changed. I actually opened an issue regarding this a while ago https://github.com/oxheadalpha/tezos-k8s/issues/378.

nicolasochem:
@harryttd I removed `chmod 777` and tried both local storage and remote, minikube and AWS (on oxheadhosted). It consistently works. I assume this `chmod 777` is no longer needed and removed it. please re-review.

charts/tezos/scripts/upgrade-storage.sh
```diff
@@ -1,6 +1,6 @@
 set -ex
 
-if [ ! -e /var/tezos/node/data/context ]
+if [ ! -e /var/tezos/node/data/context/store.dict ]
 then
   printf "No context in data dir found, probably initial start, doing nothing."
```
harryttd, line 5: more accurate? 
```suggestion
  printf "No store in data dir found, probably initial start, doing nothing."
```

charts/tezos/templates/nodes.yaml
```diff
@@ -66,10 +66,14 @@ spec:
...
           name: var-volume
-{{- else }}
+{{- end }}
   volumeClaimTemplates:
     - metadata:
+{{- if $.node_vals.local_storage | default false }}
+        name: persistent-volume
+{{- else }}
         name: var-volume
+{{- end }}
```
harryttd, line 76: I'm confused why volumeClaimTemplates is needed if using local storage. That means we don't need any ebs backing volume if using the localstorage on the ec2 node.

nicolasochem: recall than when using local storage, I am still deploying a 1Gi PVC to store the identity file (and peers file if we ever manage to make it appear). This is what this code is doing.

harryttd: ok thx. Can we leave a comment this is what the volume ends up being for?

charts/tezos/templates/nodes.yaml
```diff
@@ -66,10 +66,16 @@ spec:
 {{- if $.node_vals.local_storage | default false }}
         - emptyDir: {}
           name: var-volume
-{{- else }}
+{{- end }}
   volumeClaimTemplates:
     - metadata:
+{{- if $.node_vals.local_storage | default false }}
+        # since we are using local storage, deploy a small persistent voluem
+        # to store identity.json and peers.json
```
harryttd, line 74:
```suggestion
        # since we are using local storage, deploy a small persistent volume
        # to store identity.json and peers.json
```
Also it may need to be wrapped in {{ /* */}}. Unless helm template works

harryttd:
Were snapshot / tarballs tested?

charts/tezos/templates/nodes.yaml
nicolasochem: helm template works

harryttd (approved):
awesome. thank you

PR #549, opened 2023-02-17
Network argument is broken so we're removing it

orcutt989:
`--network` has been retired as of v16 and it is redundant to `--data-dir`. We've tested it on both v15 and v16 and the omission of this does not break backwards compatability.


nicolasochem (approved):
thanks, I can confirm v15 is not broken with this.

PR #553, opened 2023-02-23
Abilities to create BCD Frontend + to Use External DB

harryttd:
- Add ability to deploy the BCD frontend
- Ability to spin up a postgres container for the indexer, or configure the settings to use an external DB such as AWS RDS Aurora.


nicolasochem (approved):
Thanks! Accepting, for the record there are a few issues to address later:

* BCD should be a chart instead of being part of tezos chart. We did it this way as we had only one chart in the tezos-k8s repo back then. Now we have multiple, and I don't see why BCD would be deployed as part of tezos chart.
   * similarly, there is a mechanism to optionally deploy a postgres db as part of the BCD deployment. But in a chart-centric view, BCD should be its own chart. So should postgres
* there is a hack with sed to make the UI recompile during the entrypoint, so the backend endpoint can be reached. We should work with the upstream software so it's no longer needed in the future.

PR #554, opened 2023-02-23
fix update storage logic to take into account config file

nicolasochem:
With v15 to v16 upgrades, we started seeing failures updating storage of non-mainnet networks.

We reported upstream:

https://gitlab.com/tezos/tezos/-/issues/4892#note_1287844009

However it turns out that the data dir we were passing did not have a config file, so it was unable to figure out which network the data dir was for.

This is fixed by passing config file instead of data dir.

I tried by launching v15 ghostnet from a snapshot, then upgrading to v16. I also tried launching a v16 tarball directly from scratch. Both worked.

test

### DIFF
--- a/charts/tezos/scripts/upgrade-storage.sh
+++ b/charts/tezos/scripts/upgrade-storage.sh
@@ -5,4 +5,4 @@ then
   printf "No context in data dir found, probably initial start, doing nothing."
   exit 0
 fi
-octez-node upgrade storage --data-dir /var/tezos/node/data
+octez-node upgrade storage --config /etc/tezos/config.json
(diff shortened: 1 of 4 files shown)


harryttd:
why doesn't it have a config file? I thought it get's copied over there

nicolasochem:
> why doesn't it have a config file? I thought it get's copied over there

No, it's ok to have tezos node have a data dir in one place and the config file in another place. The location of the data dir is set in the config file. Ever since the beginning of tezos-k8s, we've had the config file in /etc/tezos/config.json and the data dir in /var/tezos/node/data.

Or maybe it gets copied over there? In any case, if it does, this container runs before this happens, so the PR is necessary. At least with v16. With v15, it was not, for some reason.

orcutt989 (approved):

PR #556, opened 2023-02-28
rename tezos-baker to octez-baker

nicolasochem:
now that v14 is no longer relevant we can finish renaming. The current terminology will probably be deprecated in v17.


orcutt989 (approved):

harryttd:
Besides for the baker cmds, should tezos-signer be updated as well?

https://github.com/oxheadalpha/tezos-k8s/blob/c86117b7a5b3787db255289da8a2fd69d5e25b76/charts/tezos/scripts/remote-signer.sh#L9

nicolasochem:
@harryttd good catch, fixed as well.

harryttd (approved):

PR #559, opened 2023-03-08
Add external db support for tzkt

puppetninja:
This PR allows external postgres db to specified as backend db for tzkt

### DIFF
--- a/charts/tezos/templates/tzkt_indexer.yaml
+++ b/charts/tezos/templates/tzkt_indexer.yaml
@@ -29,7 +29,7 @@ data:
   POSTGRES_PASSWORD: {{ b64enc $tzkt_indexer.db.password }}
   POSTGRES_DB: {{ b64enc $tzkt_indexer.db.name }}
 
-{{- if $tzkt_indexer.db.snapshot_url }}
+{{- if and (not $tzkt_indexer.db.isExternal) $tzkt_indexer.db.snapshot_url }}
 ---
 {{/*
 This script is mounted into postgres's /docker-entrypoint-initdb.d directory
@@ -95,9 +95,8 @@ spec:
       labels:
         app: {{ .Values.tzkt_indexer_statefulset.name }}
     spec:
-    {{- if not $tzkt_indexer.db.isExternal }}
+    {{- if and (not $tzkt_indexer.db.isExternal) $tzkt_indexer.db.snapshot_url }}
       initContainers:
-      {{- if $tzkt_indexer.db.snapshot_url }}
         - image: {{ .Values.tezos_k8s_images.utils }}
           name: download-db-snapshot
           volumeMounts:
@@ -195,7 +194,6 @@ spec:
 
               pg_ctl stop
               [ "$import_exit_code" != 0 ] && exit 1 || exit 0
-      {{- end }}
     {{- end }}
       containers:
         - image: {{ .Values.tezos_k8s_images.utils }}
--- a/charts/tezos/templates/tzkt_indexer.yaml
+++ b/charts/tezos/templates/tzkt_indexer.yaml
@@ -1,5 +1,5 @@
 {{- define "tezos.tzkt.setDbConnectionEnvVar" -}}
-{{ print " " "server=localhost;port=5432;database=$(POSTGRES_DB);username=$(POSTGRES_USER);password=$(POSTGRES_PASSWORD);" }}
+{{ print " " "server=$(POSTGRES_HOST);port=5432;database=$(POSTGRES_DB);username=$(POSTGRES_USER);password=$(POSTGRES_PASSWORD);" }}
 {{- end -}}
 
 
@@ -9,7 +9,9 @@
  The path of a downloaded snapshot.
  The file's name format is "{CHAIN_NAME}-{indexer_image_tag}.backup"
 */}}
-{{- $snapshot_file := print "/etc/db-snapshot/" $tzkt_indexer.config.db_name "-" $tzkt_indexer.indexer_image_tag ".backup" }}
+{{- /* Extract tzkt version from tzkt image */}}
+{{- $indexer_image:= split ":" $tzkt_indexer.images.indexer }}
+{{- $snapshot_file := print "/etc/db-snapshot/" $tzkt_indexer.config.db_name "-" $indexer_image._1 ".backup" }}
 {{- /*
  A file every pod uses to remember its imported snapshot.
  By keeping track, pods won't try to re-upgrade their db on restarts.
@@ -21,11 +23,12 @@ kind: Secret
 metadata:
   name: db-creds
 data:
+  POSTGRES_HOST: {{ b64enc ($tzkt_indexer.config.db_host | default "localhost") }}
   POSTGRES_USER: {{ b64enc $tzkt_indexer.config.db_user }}
   POSTGRES_PASSWORD: {{ b64enc $tzkt_indexer.config.db_password }}
   POSTGRES_DB: {{ b64enc $tzkt_indexer.config.db_name }}
 
-{{- if $tzkt_indexer.db_snapshot_url }}
+{{- if $tzkt_indexer.db.snapshot_url }}
 ---
 {{/*
 This script is mounted into postgres's /docker-entrypoint-initdb.d directory
@@ -91,8 +94,9 @@ spec:
       labels:
         app: {{ .Values.tzkt_indexer_statefulset.name }}
     spec:
+    {{- if not $tzkt_indexer.db.isExternal }}
       initContainers:
-      {{- if $tzkt_indexer.db_snapshot_url }}
+      {{- if $tzkt_indexer.db.snapshot_url }}
         - image: {{ .Values.tezos_k8s_images.utils }}
           name: download-db-snapshot
           volumeMounts:
@@ -101,7 +105,7 @@ spec:
               subPath: db-snapshot
           env:
             - name: SNAPSHOT_URL
-              value: {{ .Values.indexers.tzkt.db_snapshot_url }}
+              value: {{ $tzkt_indexer.db.snapshot_url }}
             - name: SNAPSHOT_FILE
               value: {{ $snapshot_file }}
             - name: MY_IMPORTED_SNAPSHOTS_URL
@@ -134,7 +138,7 @@ spec:
 
               cat "$MY_IMPORTED_SNAPSHOTS_URL"
         - name: upgrade-db
-          image: postgres:13
+          image: "{{ $tzkt_indexer.images.postgres }}"
           volumeMounts:
             - name: indexer-db
               mountPath: /var/lib/postgresql/data
@@ -149,7 +153,7 @@ spec:
             - name: PGDATA
               value: /var/lib/postgresql/data/pgdata
             - name: SNAPSHOT_URL
-              value: {{ .Values.indexers.tzkt.db_snapshot_url }}
+              value: {{ $tzkt_indexer.db.snapshot_url }}
             - name: SNAPSHOT_FILE
               value: {{ $snapshot_file }}
             - name: MY_IMPORTED_SNAPSHOTS_URL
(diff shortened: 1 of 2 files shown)


harryttd:
Thanks for the contribution! I'll be leaving some comments

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -91,8 +94,9 @@ spec:
       labels:
         app: {{ .Values.tzkt_indexer_statefulset.name }}
     spec:
+    {{- if not $tzkt_indexer.db.isExternal }}
       initContainers:
-      {{- if $tzkt_indexer.db_snapshot_url }}
+      {{- if $tzkt_indexer.db.snapshot_url }}
```
harryttd, line 99: Why if we are using a separate db from the pod should we not handle snapshots anymore? Even if say there is a separate postgres db within the cluster, or an RDS instance, we should still be able to download snapshots for quick db restore.

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -91,8 +94,9 @@ spec:
       labels:
         app: {{ .Values.tzkt_indexer_statefulset.name }}
     spec:
+    {{- if not $tzkt_indexer.db.isExternal }}
```
harryttd, line 97: Why if we are using a separate db from the pod should we not handle snapshots anymore? Even if say there is a separate postgres db within the cluster, or an RDS instance, we should still be able to download snapshots for quick db restore.

charts/tezos/values.yaml
```diff
@@ -471,21 +471,25 @@ protocols:
...
+#       ## https://github.com/baking-bad/tzkt
+#       snapshot_url: https://snapshots.tzkt.io/tzkt_v1.11_mainnet.backup
 #
 #     ## Configurable tzkt fields
 #     config:
-#       ## Url of the archive node to index. You will need to create an archive node up
-#       ## above in the `nodes` section. The `rpc_url` field is looking for a node named
-#       ## `archive-node`.
+#       ## Url of the rpc node to index. No archive node is needed if the instance is bootstrapped
+#       ## from a snapshot
```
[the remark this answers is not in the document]
puppetninja: I wrote this down based on the doc here
https://github.com/baking-bad/tzkt

> No archive node needed. There is no need to use an archive node (running in "archive" mode). If you bootstrap the indexer from the most recent snapshot, using a simple rolling node will be enough.

I think archive node is only needed if you start from scratch for indexing, but if you import a db snapshot or dumped backup, you can use a rolling node to catch up to latest block for indexing

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -134,7 +138,7 @@ spec:
 
               cat "$MY_IMPORTED_SNAPSHOTS_URL"
         - name: upgrade-db
-          image: postgres:13
+          image: "{{ $tzkt_indexer.images.postgres }}"
```
harryttd, line 141: I just want to leave a comment here explaining my understanding of how this will work with the latest versions of tzkt. We are doing a big jump from the old v1.6 tzkt version we handled, to the latest v11 version. 

They have now implemented a way for some version upgrades to not require dropping the db and syncing from scratch or using a snapshot. Just by changing the docker image it can perform necessary db migrations. This container will always drop and re-create the db when there is a new snapshot url. That should still be the correct functionality. Because if a new url is being used that implies that the user needs a fresh db to import the snapshot into.

If the user just updates the docker images, that implies that there is no need to recreate the db. Rather the migrations would run and update automatically.

harryttd (changes requested):

charts/tezos/values.yaml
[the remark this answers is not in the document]
harryttd: oh very cool. Can you add to the comment that a rolling node is sufficient instead of an archive node?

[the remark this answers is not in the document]
puppetninja: yes, will do

charts/tezos/templates/tzkt_indexer.yaml
puppetninja: I am a a little bit lost here, basically if there is an external db instance, we manage the snapshot download and import independently from this chart ?

charts/tezos/templates/tzkt_indexer.yaml
puppetninja: I am not sure whether I follow here as I didn't change any db migration logic, and I tested with db in same pod as well, the only pain for me is the mainnet db import within the same pod takes ages for the pod to be up. So I decided to use chart from bitnami which allows me to customized and import the db snapshot.

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -283,10 +289,11 @@ spec:
           envFrom:
           - secretRef:
               name: db-creds
+        {{- end }}
       volumes:
         - name: tzkt-env
           emptyDir: {}
-        {{- if $tzkt_indexer.db_snapshot_url }}
+        {{- if not $tzkt_indexer.db.isExternal }}
```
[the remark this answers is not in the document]
puppetninja: > However, i left another comment saying that we should still provide the upgrade functionality for external db's if a snapshot url is provided.

Sorry I am not sure I follow here, may I ask why ? I would assume this will be handled by DBA or DB chart independently ?

charts/tezos/values.yaml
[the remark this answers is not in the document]
puppetninja: Done

charts/tezos/templates/tzkt_indexer.yaml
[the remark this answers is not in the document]
harryttd: I hear your point. We could continue to leave it up to the chart to handle snapshots and not need that a human needs to manually handle them. The code already exists. @nicolasochem What do you think?

charts/tezos/values.yaml
```diff
@@ -480,24 +480,35 @@ protocols:
...
+#       ## You can find more info from tzket documentation here
+#       ## https://github.com/baking-bad/tzkt
+#       ## No archive node needed.
+#       ## There is no need to use an archive node (running in "archive" mode).
+#       ## If you bootstrap the indexer from the most recent snapshot,
+#       ## using a simple rolling node will be enough.
+#       ##
+#       rpc_url: http://rolling-node-0.rolling-node:8732
```
harryttd, line 511: please clean up. things are repeated and the tzkt github link is used above

```suggestion
#       ## RPC url of the node to index. A rolling node is sufficient if the
#       ## indexer is bootstrapped from the most recent snapshot. Otherwise an
#       ## archive node is needed.
#       rpc_url: http://rolling-node-0.rolling-node:8732
```

charts/tezos/templates/tzkt_indexer.yaml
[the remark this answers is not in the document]
nicolasochem: The right time to import a snapshot is when the DB starts. If the DB is managed by another chart, then this other chart is responsible for downloading the snapshot. Having logic in our chart to download snapshot only makes sense when the chart also handles the DB. Otherwise, without control of the DB itself, it's impossible to decide when/whether to import or not.

DB management is something that we provide for prototyping and demos, I would expect prod deployments to use an externally provisioned DB.

So it makes sense to not import anything is the db is external.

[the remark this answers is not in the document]
harryttd: ok makes sense. Thank you. @puppetninja can you please also make sure that `volumeClaimTemplates` are only rendered if we are not using an external db?

charts/tezos/values.yaml
```diff
@@ -466,10 +466,10 @@ protocols:
 ##
 # indexers:
 #   tzkt:
-#     api_image: "bakingbad/tzkt-api"
-#     api_image_tag: "1.6.4"
-#     indexer_image: "bakingbad/tzkt-sync"
-#     indexer_image_tag: "1.6.4"
+#     images:
+#       api: "bakingbad/tzkt-api:1.11.1"
+#       indexer: "bakingbad/tzkt-sync:1.11.1"
```
harryttd, line 471: they just came out with v12 https://github.com/baking-bad/tzkt/releases/tag/v1.12.0

charts/tezos/templates/tzkt_indexer.yaml
puppetninja: I just used the bitnami chart to allow me to deploy postgres in a more sophisticated way like multiple replicas and utilized the init  script to import the snapshot just for once

charts/tezos/templates/tzkt_indexer.yaml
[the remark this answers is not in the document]
puppetninja: HI @harryttd 
This is what I see 
Screenshot from 2023-03-29 21-51-53
Screenshot from 2023-03-29 21-53-22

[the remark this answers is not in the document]
puppetninja: @harryttd done, can you please review again ?

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -1,5 +1,5 @@
 {{- define "tezos.tzkt.setDbConnectionEnvVar" -}}
-{{ print " " "server=localhost;port=5432;database=$(POSTGRES_DB);username=$(POSTGRES_USER);password=$(POSTGRES_PASSWORD);" }}
+{{ print " " "server=$(POSTGRES_HOST);port=5432;database=$(POSTGRES_DB);username=$(POSTGRES_USER);password=$(POSTGRES_PASSWORD);command timeout=600;" }}
```
puppetninja, line 2: This is needed  per
https://github.com/baking-bad/tzkt/blob/master/docker-compose.yml#L24

harryttd: I don't think we should be doing this. If something is broken the connection will still wait 10 minutes. The docker compose file defaults to 600s but still lets a user configure the value.

puppetninja: For mainnet we did hit the issue you can find more here
https://github.com/baking-bad/tzkt/issues/146

charts/tezos/values.yaml
```diff
@@ -481,24 +481,30 @@ protocols:
...
-#       ## `archive-node`.
-#       rpc_url: http://archive-node-0.archive-node:8732
-#       db_name: db
-#       db_user: foo
-#       db_password: bar
+#       ## Url of the rpc node to index. No archive node is needed if the instance is bootstrapped
+#       ## from a snapshot
+#       ## You can find more info from tzket documentation here
+#       ## https://github.com/baking-bad/tzkt
+#       rpc_url: http://rolling-node-0.rolling-node:8732
```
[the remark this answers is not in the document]
puppetninja: Done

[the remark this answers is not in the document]
puppetninja: And will not do force pushes

charts/tezos/templates/tzkt_indexer.yaml
harryttd: ok very cool. Let's make it configurable though?

puppetninja: Okay, I can check into this

puppetninja: I was not sure whether I should create a dedicated configmap for this, as this is not really go into db creds secret imo

puppetninja: same for db port maybe

puppetninja: @harryttd done and I verified on our testing cluster

charts/tezos/templates/tzkt_indexer.yaml
```diff
@@ -91,8 +95,9 @@ spec:
       labels:
         app: {{ .Values.tzkt_indexer_statefulset.name }}
     spec:
+    {{- if not $tzkt_indexer.db.isExternal }}
       initContainers:
-      {{- if $tzkt_indexer.db_snapshot_url }}
+      {{- if $tzkt_indexer.db.snapshot_url }}
```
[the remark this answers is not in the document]
puppetninja: @harryttd done

harryttd (approved):
thank you!

PR #560, opened 2023-03-08
Add more k8s template configuration options for BCD

harryttd:


puppetninja:
I think it would be good to add some comments to values.yaml for this

harryttd:
I ended up specifying the added fields explicitly in values.yaml

nicolasochem (approved):
Thanks! I am assuming it works properly. For values.yaml I think it's pretty self explanatory because you can find the same fields in many charts, including the empty chart created by `helm new`.

PR #562, opened 2023-03-09
Add lock to tezos sidecar

nicolasochem:
When the node has a lot of contention due to a lot of RPC requests, I noticed the following:

* node keeps processing new blocks
* node does not reply to any RPC request even /version, even locally
* doing netstat on the node shows a lot of connections from localhost (sidecar connections)
* sometimes the node recovers on its own, sometimes not

I am wondering if the sidecar is not making things worse in this context, deadlocking the node even more. So I am proposing that the sidecar will only make one http request to the node at a time using multiprocessing.Lock

I also asked Nomadic engineers what they think about this.

### DIFF
--- a/utils/sidecar.py
+++ b/utils/sidecar.py
@@ -3,6 +3,7 @@
 from flask import Flask, escape, request
 import requests
 import datetime
+from multiprocessing import Lock
 
 import logging
 
@@ -13,6 +14,7 @@ application = Flask(__name__)
 
 AGE_LIMIT_IN_SECS = 600
 
+lock = Lock()
 
 @application.route("/is_synced")
 def sync_checker():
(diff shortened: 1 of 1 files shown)


utils/sidecar.py
```diff
@@ -22,33 +24,37 @@ def sync_checker():
...
-    block_age = datetime.datetime.utcnow() - datetime.datetime.strptime(
-        timestamp, "%Y-%m-%dT%H:%M:%SZ"
-    )
-    age_in_secs = block_age.total_seconds()
-    if age_in_secs > AGE_LIMIT_IN_SECS:
-        err = (
-            "Error: Chain head is %s secs old, older than %s"
-            % (age_in_secs, AGE_LIMIT_IN_SECS),
-            500,
+    with lock:
```
puppetninja, line 31: I am not sure about here, should we timeout if we cannot obtain a lock ? I will google around tomorrow

harryttd:
The  questions would seem to me why is the sidecar creating so many connections? It only needs one.

harryttd:
After discussing on slack, based on our understanding the solution should be implementing a global lock. This way every ping from the probe allows the invoked function to check if there is an already existing lock from a previous invocation. If there is, we assume the node is stuck. The probe runs every 5s and it shouldn't take 5s for the node to respond. If there is a lock the python process will exit with an error.

This method will prevent the probe from pinging the node needlessly and it will also prevent python processes from piling on top of each other. If the lock wasn't global we would end up with queued processes waiting to ping the node, resulting in overloading the sidecar container.

utils/sidecar.py
puppetninja: I am still with my idea here, a lock without timeout will not establish a connection to the node, when there is a long hanging connection, the kubelet will still not obtain the status of the readiness probe as it is not returning anything

utils/sidecar.py
```diff
@@ -22,33 +24,37 @@ def sync_checker():
...
+        try:
+            r = requests.get("http://127.0.0.1:8732/chains/main/blocks/head/header")
```
puppetninja, line 33: I would suggest to add a timeout parameter here for like 15s, which is too long for head request, if it times out we return error

harryttd:
@nicolasochem closing due to #581

PR #563, opened 2023-03-17
snapshotEngine: skip snapshot website

orcutt989:
This PR adds a configurable value to the snapshot engine chart `skipWebsite` that when set to `true` tells the `zip and upload` job to not build and upload the website. This is a preemptive measure for a multiple S3 vendor project that is currently underway, but I think would be useful to introduce now.

### DIFF
--- a/charts/snapshotEngine/templates/configmap.yaml
+++ b/charts/snapshotEngine/templates/configmap.yaml
@@ -13,6 +13,7 @@ data:
   ARCHIVE_SLEEP_DELAY: {{ $.Values.artifactDelay.archive }}
   ROLLING_SLEEP_DELAY: {{ $.Values.artifactDelay.rolling }}
   SCHEMA_URL: {{ $.Values.schemaUrl }}
+  SKIP_WEBSITE: {{ $.Values.skipWebsite}}
 kind: ConfigMap
 metadata:
   name: snapshot-configmap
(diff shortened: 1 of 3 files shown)


nicolasochem (commented):
There is already a parameter to pass the jekyll page url.

It's better to skip the website build if the jekyll page is empty or unset?

Also we should probably not merge until we are done with both work streams (alternative bucket provider and faulty tarball repository). It's good to have a branch though so we can leave this open as draft.

orcutt989:
> There is already a parameter to pass the jekyll page url.
> 
> It's better to skip the website build if the jekyll page is empty or unset?
> 
> Also we should probably not merge until we are done with both work streams (alternative bucket provider and faulty tarball repository). It's good to have a branch though so we can leave this open as draft.

Do you mean the jekyllRemoteThemeRepository or the snapshotWebsiteDomain @nicolasochem ?

nicolasochem:
> > There is already a parameter to pass the jekyll page url.
> > It's better to skip the website build if the jekyll page is empty or unset?
> > Also we should probably not merge until we are done with both work streams (alternative bucket provider and faulty tarball repository). It's good to have a branch though so we can leave this open as draft.
> 
> Do you mean the jekyllRemoteThemeRepository or the snapshotWebsiteDomain @nicolasochem ?

Any of these 2 would work. snapshotWebsiteDomain is more logical. So if it's absent, don't deploy a website.

nicolasochem (approved):
thanks

PR #565, opened 2023-03-23
fix tezos pod crash when config already exists

nicolasochem:
Today I found a baking node in this state: the init container was in a failure loop. `config-init` had the following message:

```
Writing custom configuration for public node

octez-node: Error:

              Pre-existing configuration file at
/etc/tezos/data/config.json, use `reset`.
```

Addressing this by ensuring config-init always works, despite pre-existing data. This does not solve the root cause (unknown) but ensures that the tezos pod starts in this scenario.

This is not usual; config-volume is an emptydir so when the pod restart, it should be empty. It could be that this specific cloud provider does not clear the config dir upon pod reboot, or the init-containers re-run in the pod when config-volume is not empty.

### DIFF
--- a/charts/tezos/scripts/config-init.sh
+++ b/charts/tezos/scripts/config-init.sh
@@ -5,9 +5,9 @@ mkdir -p /etc/tezos/data
 
 # if config already exists (container is rebooting), dump and delete it.
 if [ -e /etc/tezos/data/config.json ]; then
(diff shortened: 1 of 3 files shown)


harryttd:
was this running on EKS?

harryttd (approved):

nicolasochem:
> was this running on EKS?

no, a small cloud provider called scaleway

PR #568, opened 2023-03-28
CI: Pypi publish tag update

orcutt989:
```You are using "pypa/gh-action-pypi-publish@master". The "master" branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use "pypa/gh-action-pypi-publish@release/v1". If you feel adventurous, you may opt to use use "pypa/gh-action-pypi-publish@unstable/v1" instead. A more general recommendation is to pin to exact tags or commit shas.```

**pypa/gh-action-pypi-publish@master**

to 

**pypa/gh-action-pypi-publish@release/v1**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
 
       - name: Publish package to PyPI
         if: github.event_name == 'release' && github.event.action == 'created'
-        uses: pypa/gh-action-pypi-publish@master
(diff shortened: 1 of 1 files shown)


nicolasochem (approved):
ok but let's make sure that's indeed the problem with release pipeline before re-releasing.

PR #575, opened 2023-04-06
rework proto cruncher to use new rust tz-proto-vanity

nicolasochem:
The previous proto cruncher was in python and embedded in this helm chart, running in the utils container. Our new one leverages Igor's project:

https://github.com/oxheadalpha/tz-proto-vanity


nicolasochem:
@orcutt989 I'd appreciate a review since Aryeh went off already for the week.

orcutt989 (approved):

PR #576, opened 2023-04-11
snapshotEngine: Switch chart from AWS to Digitalocean

orcutt989:
This branch allows for user-configurable values `s3BucketOverride` and `cloudProvider`.

By default the snapshotEngine helm chart is very AWS and configuration dependent at this time and we plan on making it interchangeable with the popular cloud vendors in the future.

By default - the S3 bucket name is assumed from `<namespace>`.`<snapshot_website_domain_name>`. Eventually we would like to anticipate this from the infrastructure in some way, but for now its a static value.  `s3BucketOverride` allows one to simply override this assumed value with a value of their own.  

The Zip and Upload job anticipates that alternative credentials will be mounted as a secret named `cloud-provider` in the same namespace. This secret is mounted as a volume and then retrieved and used in an `alias` for the `aws` CLI command.


snapshotEngine/getAllSnapshotMetadata.py
```diff
@@ -20,7 +20,7 @@
 # Get each subdomain's base.json and combine all artifacts into 1 metadata file
 for subDomain in allSubDomains:
     baseJsonUrl = (
-        "https://" + subDomain + "." + snapshotWebsiteBaseDomain + "/base.json"
+        "https://" + subDomain + "-shots" + "." + s3Endpoint + "/base.json"
```
nicolasochem, line 23: Look like once merged, digitalocean will be hardcoded. Is it expected?

snapshotEngine/zip-and-upload.sh
```diff
@@ -5,75 +5,105 @@ BLOCK_HASH=$(cat /"${HISTORY_MODE}"-snapshot-cache-volume/BLOCK_HASH)
...
 
+# Sets aws command credentials depending on cloud provider
+# Alias functionality outside of command blocks is not consistent
+#  so we opted for this.
+# $1 is file name
+set_aws_command_creds(){
+    if [[ -n ${CLOUD_PROVIDER} ]]; then
+        echo "AWS_ACCESS_KEY_ID=$(cat /cloud-provider/access-id) \
+            AWS_SECRET_ACCESS_KEY=$(cat /cloud-provider/secret-key) \
+            aws --endpoint-url https://nyc3.digitaloceanspaces.com "
```
nicolasochem, line 41: Some more hardcoding here? if `CLOUD_PROVIDER` is set then the endpoint is necessarily DO?

nicolasochem (commented):
On the one hand it looks like you are aiming for backward-compatibility with AWS based on the presence/absence of CLOUD_PROVIDER, but on the other hand I've seen places where it looks like Digitalocean is always assumed? I left comments.

Not necessarily a problem as I believe we can deprecate AWS support when DO works...

snapshotEngine/zip-and-upload.sh
orcutt989: The idea was that when this function is `eval`ed if `CLOUD_PROVIDER` is set then digitalocean is the endpoint, or rather this region.  So I wouldnt say DO is hardcoded here, only the region.  If `CLOUD_PROVIDER` is not set then the creds remain AWS.

snapshotEngine/getAllSnapshotMetadata.py
orcutt989: We absolutely could! 😄

orcutt989: The scope started out to make it configurable but then we honed in to just make DO work for now.

snapshotEngine/zip-and-upload.sh
harryttd: What if a user sets it to aws? Then this would run as digital ocean. In other places in the code execution is determined based on what the value is and not assumed just if it has a value.

snapshotEngine/getAllSnapshotMetadata.py
nicolasochem: I'm not sure actually. No one else is using the snapshot engine, and we want snapshots to go away completely in favor of warp sync, so we should focus on our own use case.

nicolasochem (approved):

nicolasochem (approved):

snapshotEngine/getAllSnapshotMetadata.py
orcutt989: I wholeheartedly agree - however we were very AWS-dependent to begin making configuration difficult. My intention with this branch was to make it configurable, but our cost-savings goals have become more pressing versus usability unfortunately.  That is why this branch makes it seem like its configurable, but then when I did the metadata I just flipped it so we could get to the next step. You can see my journey in @nicolasochem's comments.

I am working in a linear fashion and currently the goal is to get the creation of artifacts off of AWS and onto DigitalOcean. This is an old branch whose functionality is needed for another branch I am working on that does this.

Unfortunately enabling more cross-cloud configuration will require a bit more effort and time that will take away from our current cost-savings goal. Happy to put some effort into it later, but the charges we're incurring from shipping data from AWS to DigitalOcean are pressing. If we're also going to be pitching snapshots then that also suggests we should think about it later or just forget about it.

snapshotEngine/zip-and-upload.sh
orcutt989: If its not set then the creds are run via the service account mounted to the pod which is AWS-authenticated as normal.

snapshotEngine/getAllSnapshotMetadata.py
nicolasochem: yes, as said before, let's just do DigitalOcean. We are not helping anyone by making this code multi-cloud.

PR #577, opened 2023-04-11
Add ability to set operations-pool per baker

nicolasochem:
Operations-pool allows the baker to target a flashbake endpoint or other external mempool provider by passing a URL. This was previously done by passing BAKER_EXTRA_ARGS env variable, but it would be the same for every baker in your pod. This allows setting an endpoint per baking address.

### DIFF
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -88,17 +88,18 @@ metadata:
   name: tezos-secret
   namespace: {{ .Release.Namespace }}
 ---
-{{- if (include "tezos.hasPerBlockVotes" .) }}
 apiVersion: v1
 data:
 {{- range $k, $v := .Values.accounts }}
 {{- range ( $v.protocols | default $.Values.protocols ) }}
   {{ $k }}-{{ .command }}-per-block-votes.json: {{ .vote | toJson | quote }}
 {{- end }}
+{{- if $v.operations_pool }}
+  {{ $k }}_operations_pool: {{ $v.operations_pool | quote }}
+{{- end }}
 {{- end }}
 kind: ConfigMap
 metadata:
-  name: per-block-votes
+  name: baker-config
   namespace: {{ .Release.Namespace }}
 ---
-{{- end }}
--- a/charts/tezos/templates/configs.yaml
+++ b/charts/tezos/templates/configs.yaml
@@ -88,6 +88,7 @@ metadata:
   name: tezos-secret
   namespace: {{ .Release.Namespace }}
 ---
+{{- if (include "tezos.shouldDeployBakerConfig" .) }}
 apiVersion: v1
 data:
 {{- range $k, $v := .Values.accounts }}
@@ -103,3 +104,4 @@ metadata:
   name: baker-config
   namespace: {{ .Release.Namespace }}
 ---
+{{- end }}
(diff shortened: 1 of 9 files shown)


harryttd (changes requested):

charts/tezos/templates/configs.yaml
```diff
@@ -73,17 +73,18 @@ metadata:
...
 {{- end }}
+{{- if $v.operations_pool }}
+  {{ $k }}_operations_pool: {{ $v.operations_pool | quote }}
+{{- end }}
 {{- end }}
 kind: ConfigMap
 metadata:
-  name: per-block-votes
+  name: baker-config
   namespace: {{ .Release.Namespace }}
```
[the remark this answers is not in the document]
nicolasochem: Good point, previously the `baker-config` cm would always exist. I have now re-added a condition, but it checks for the presence of bakers rather than a specific toggle vote. Whenever there is a baker, it's mandatory to set the toggle vote in values.yaml so it's ok now.

I tested in minikube on a private chain and the cm existed. Then I tried a public chain (mumbainet) and it didn't.

charts/tezos/templates/nodes.yaml
```diff
@@ -58,11 +58,9 @@ spec:
         - name: tezos-accounts
           secret:
             secretName: tezos-secret
-{{- if (include "tezos.hasPerBlockVotes" $) }}
-        - name: per-block-votes
+        - name: baker-config
           configMap:
-            name: per-block-votes
-{{- end }}
```
harryttd, line 65: I beleive this still needs a check. If a baker is not running then the cm doesn't exist

nicolasochem: now fixed

charts/tezos/values.yaml
```diff
@@ -72,6 +76,7 @@ accounts: {}
 #     key: edpk...
 #     is_bootstrap_baker_account: false
 #     bootstrap_balance: "4000000000000"
+#     operations_pool: http://flashbake-endpoint-baker-listener:12732
 #     protocols:
 #     - command: PtMumbai
 #       vote:
```
harryttd, line 82: I just realized, is this example inaccurate? It's an account for a baker, but the key passed is a public key and is_bootstrap_baker_account is false. To bake it needs a secret key unless passed a `signer_url` or `key_id`. Then public key is fine. But we show those examples later.

Perhaps baker0 should be what baker1 is but named `node0` or the like, and is shouldn't have `operations_pool` and `protocols`, because in this example baker1 isn't actually baking.

nicolasochem: In case of public chains, we don't actually need `is_bootstrap_baker_account` and `bootstrap_balance`. This is only relevant for private chains, as explained a bit above in the same file.

I fixed it in a different way, by making sure baker1 has a secret key and removing the fields specific to private chain.

test/charts/private-chain.expect.yaml
```diff
@@ -690,9 +690,9 @@ spec:
         - name: tezos-accounts
           secret:
             secretName: tezos-secret
-        - name: per-block-votes
+        - name: baker-config
```
harryttd, line 693: This sts isn't baking so I don't believe the baker config should be here

nicolasochem: fixed

harryttd (approved):
looks good

PR #579, opened 2023-04-15
EVM Rollup + DAL Nodes support

nicolasochem:
tezos-k8s "tezos" chart values files now support two new object definitions: `smartRollupNodes` and `dalNodes`.

This is to support teztnet's dailynet and mondaynet bootstrap smart rollup and nodes.

## Smart Rollup

* supports injection of smart rollup at genesis thanks to a `fromfile#` shortcut in the values.yaml. The rollup itself is too big to be passed through the helm values
* adds a deployment for EVM proxy as well

## DAL node

* supports creation of an arbitrary number of DAL nodes with bootstrap or attester profile. The baking address for the attestor is specified
* baker can be configured to target an attester by URL


charts/tezos/templates/octez-rollup-node.yaml
```diff
@@ -0,0 +1,218 @@
...
+        command:
+          - /bin/sh
+        args:
+          - "-c"
+          - |
+{{ tpl ($.Files.Get "scripts/evm-proxy.sh") $ | indent 12 }}
+      securityContext:
+        fsGroup: 1000
+---
+{{- if $v.evm_proxy_ingress | default false }}
```
harryttd, line 49: I think this is missing in value.y spec

charts/tezos/templates/octez-rollup-node.yaml
```diff
@@ -0,0 +1,218 @@
...
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+    - host: {{ $v.evm_proxy_ingress.host }}
+      http:
+        paths:
+          - pathType: ImplementationSpecific
```
harryttd, line 80: this should use the pathType prop from values.yaml and default to Prefix. I think our ingresses everywhere default to Prefix

charts/tezos/templates/octez-rollup-node.yaml
```diff
@@ -0,0 +1,218 @@
...
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+    - host: {{ $v.ingress.host }}
+      http:
+        paths:
+          - pathType: ImplementationSpecific
```
harryttd, line 209: Should use Prefix field from values.y

charts/tezos/scripts/smart-rollup-node.sh
```diff
@@ -0,0 +1,17 @@
+set -ex
+
+TEZ_VAR=/var/tezos
+TEZ_BIN=/usr/local/bin
+CLIENT_DIR="$TEZ_VAR/client"
+ROLLUP_DATA_DIR="$TEZ_VAR/rollup"
+
+xxd -ps -c 0  /usr/local/share/tezos/evm_kernel.wasm   | tr -d '\n' > /var/tezos/smart-rollup-boot-sector
```
harryttd, line 8: comment here whats going on?

charts/tezos/scripts/evm-proxy.sh
```diff
@@ -0,0 +1,9 @@
+set -ex
+
+TEZ_BIN=/usr/local/bin
+
+CMD="$TEZ_BIN/octez-evm-proxy-server run \
+  --rpc-addr 0.0.0.0 \
+  --rollup-node-endpoint  http://rollup-evm:8932"
```
harryttd, line 7: does the url need to be a variable bec there can be multiple sts's with their own services?

charts/tezos/scripts/chain-initiator.sh
```diff
@@ -4,17 +4,92 @@ until $CLIENT rpc get /chains/main/blocks/head/header | grep '"level":'; do
...
 
-set -x
 set -o pipefail
 if ! $CLIENT rpc get /chains/main/blocks/head/header | grep '"level": 0,'; then
     echo "Chain already activated, considering activation successful and exiting"
     exit 0
 fi
 
+# Substitute #fromfile with the hex encoded files in question.
+# This is for bootstrapped smart rollups.
```
harryttd, line 14: what is this all for?

charts/tezos/scripts/dal-node.sh
```diff
@@ -0,0 +1,17 @@
...
+
+mkdir -p ${DAL_DATA_DIR}
+
+CMD="$TEZ_BIN/octez-dal-node \
+  run \
+  --data-dir ${DAL_DATA_DIR} \
+  --endpoint http://tezos-node-rpc:8732 \
+  --net-addr 0.0.0.0:11732 \
+  --rpc-addr 0.0.0.0:10732 \
+  --use-unsafe-srs-for-tests"
```
linoscope, line 15: Head up: This option will be removed in https://gitlab.com/tezos/tezos/-/merge_requests/9228/ .

Context: After 9228, the DAL node will automatically sync it's dal configuration with the octez node, hence we don't need to configure this flag on the dal-node side.

PR #581, opened 2023-04-27
Add timeout for /is_synced endpoint in sidecar.py

puppetninja:
In out prod environment we could observe that requests piling up as there is no timeout in default requests get method. Here we timeout and close the connection timely to avoid the piling up.

This is an alternative way to fix the issue mentioned here https://github.com/oxheadalpha/tezos-k8s/pull/562

### DIFF
--- a/utils/sidecar.py
+++ b/utils/sidecar.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
-
 from flask import Flask, escape, request
 import requests
+from requests.exceptions import ConnectTimeout, RequestException
 import datetime
 
 import logging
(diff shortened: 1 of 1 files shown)


nicolasochem (approved):
thanks, it looks like a better way to implement sidecar, as it does not leave any state behind.

utils/sidecar.py
```diff
@@ -12,7 +12,9 @@
 application = Flask(__name__)
 
 AGE_LIMIT_IN_SECS = 600
-
+# Default readiness probeperid timeout is 1s, timeout sync request before that and return a
```
harryttd, line 15: what is probeperid?

puppetninja: it is a typo...
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
periodSeconds: How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1.

harryttd: the docs you reference say default is 10s but your comment says 1s

harryttd (changes requested):
please see my latest comment

puppetninja:
> please see my latest comment
I have  updated the comment with right probe filed name, please have a review

PR #583, opened 2023-05-04
Add Node RPC and BCD ingresses

harryttd:
Adds ability to deploy ingresses for the node rpc endpoint as well as for the BCD ui and api.

I put bcd ing in the bcd_indexer.yaml file. It's all the resources in one file. When we move bcd into it's own chart we will split the resources into different files.

I created a new ingress.yaml file for the rpc ing.

### DIFF
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -269,6 +269,16 @@ services:
     selector:
 #      rpc_node: "true"
 
+## Configuration for the tezos-node-rpc ingress
+ingress:
+  enabled: false
+  host: ""
+  annotations: {}
+  className: ""
+  labels: {}
+  pathType: Prefix
+  tls: []
+
 # Prometheus Operator is required in your cluster in order to enable
 # serviceMonitor below.
 # ServiceMonitor allows you to scrape the prometheus endpoints of your tezos nodes.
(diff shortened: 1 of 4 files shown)


charts/tezos/values.yaml
```diff
@@ -549,15 +558,28 @@ bcdIndexer:
     gui: ghcr.io/baking-bad/bcdhub-gui:4.4.0
     postgres: postgres:14
 
+  # Url of the RPC backend for the indexer to index
+  rpcUrl: "" # http://archive-baking-node-0.archive-baking-node:8732
+  # When creating an ingress, this will be the BCD api's hostname/FQDN (stripped
+  # of http(s)). When deploying the BCD gui, this will also be the api's full
+  # url called by the frontend.
+  apiUrl: "" # http://localhost:14000
```
nicolasochem, line 567: It says "stripped of http(s)" but the example has a "http://"

harryttd, line 566: I hope this is clear. Instead of needing to repeat the url and hostname for the ingress section and gui section, we define it just once and Helm does lookups for both gui and ing.

nicolasochem (changes requested):

charts/tezos/values.yaml
harryttd: yes, but is also refers to when users provide their own https urls

nicolasochem (approved):
Looks good, I didn't try but I am told by the author that it's already deployed in a private deployment and works fine.

charts/tezos/values.yaml
gisele-igore: Nice one

PR #584, opened 2023-05-09
Octez: increase request to 700Mi and limit to 1Gi

orcutt989:
A tezos node runs anywhere between 600MB and 700MB of RAM.  

By having the `request` so low (Currently set at 100Mi) it prevents the pod from being scheduled accurately to a node.  I discovered this while downscaling oxheadinfra. 

What happens is the schedule controller only sees the 100Mi request, and goes "Ok node you can handle 40 of these or more", but then since its actually 6x that the node OOMs and can't run the pods until a sizable node is made available.  

Now with a proper request pods wont be scheduled to a node that cant actually handle it.

### DIFF
--- a/charts/tezos/templates/_containers.tpl
+++ b/charts/tezos/templates/_containers.tpl
@@ -254,7 +254,7 @@
 
 {{- define "tezos.container.sidecar" }}
   {{- if or (not (hasKey $.node_vals "readiness_probe")) $.node_vals.readiness_probe }}
-    {{- $sidecarResources := dict "requests" (dict "memory" "80Mi") "limits" (dict "memory" "100Mi") -}}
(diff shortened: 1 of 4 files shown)


nicolasochem:
Here you are modifying the sidecar resources, not the octez resources themselves.

It's already possible to set the octez resources (requests and limits): https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/values.yaml#L216-L220

Now the question is: is it right that default values are unset (and therefore only the sidecar request applies)?

I believe it is fine, since when deploying an octez node you are likely to need just one (like one baker or one node) and therefore you won't run into scaling issues either way. And when you have a fleet (as we do) you can modify values.yaml and the chart does not need any modification.

PR #585, opened 2023-05-15
add persistent volume for pyrometer

nicolasochem:
Pyrometer saves its message queue on disk. Upon restart, we ensure that
the messages in the queue will be processed.

### DIFF
--- a/charts/pyrometer/templates/volume.yaml
+++ b/charts/pyrometer/templates/volume.yaml
@@ -2,6 +2,7 @@ apiVersion: "v1"
 kind: PersistentVolumeClaim
 metadata:
   name: pyrometer-volume
+  namespace: {{ .Release.Namespace }}
 spec:
  storageClassName:
  accessModes:
--- a/charts/pyrometer/templates/volume.yaml
+++ b/charts/pyrometer/templates/volume.yaml
@@ -1,0 +1,11 @@
+apiVersion: "v1"
+kind: PersistentVolumeClaim
+metadata:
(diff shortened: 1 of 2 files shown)


itkach (approved):
lgtm

charts/pyrometer/templates/volume.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: "v1"
+kind: PersistentVolumeClaim
+metadata:
+  name: pyrometer-volume
```
harryttd, line 4: Can you add namespace field?

nicolasochem: sure, I was wondering why do we have namespace fields in all our charts?

harryttd: It's a normal field in namespace scoped resources. The Helm generated charts add it. I've missed adding it before and it created issues with the resource being deployed in the wrong namespace. I believe the default one.

harryttd (approved):

charts/pyrometer/templates/volume.yaml
nicolasochem: But I assume when you use `helm --namespace x` then all resources will go to x. except perhaps the namespace-less control plane resources.

harryttd: i know that with `helm template` no namespace is added to the resources. Even if you pass `--namespace`

PR #588, opened 2023-06-09
switch pyrometer to port 8080

nicolasochem:
fix after teztnets breakage. there is no pyrometer twin PR as the port is configurable purely from config.json without a docker change.


harryttd:
where is the value to change the port? Only templates are in this PR

nicolasochem:
> where is the value to change the port? Only templates are in this PR

It should be in values.yaml, in the `config:` section, but there is no example of config in the values file. See pyrometer readme.

harryttd (approved):

PR #589, opened 2023-06-23
Tezos-k8s nairobi release

nicolasochem:
As usual:

* mkchain chain gets activated using nairobi proposal
* bump version by default to v17

See previous PR for reference: https://github.com/oxheadalpha/tezos-k8s/pull/538


harryttd (approved):
thank you

PR #594, opened 2023-07-11
Add local_storage_size option in values

puppetninja:
Some cloud provider have minimum pvc size requirement, like Aliyun, this option would also tezos-k8s to be deployed under such scenarios.

### DIFF
--- a/charts/tezos/templates/nodes.yaml
+++ b/charts/tezos/templates/nodes.yaml
@@ -86,7 +86,7 @@ spec:
         resources:
           requests:
 {{- if $.node_vals.local_storage | default false }}
-            storage: "1Gi"
+            storage: {{ default "1Gi" $.node_vals.local_storage_size }}
 {{- else }}
             storage: {{ default "15Gi" $.node_vals.storage_size }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


charts/tezos/templates/nodes.yaml
```diff
@@ -86,7 +86,7 @@ spec:
         resources:
           requests:
 {{- if $.node_vals.local_storage | default false }}
-            storage: "1Gi"
+            storage: {{ default "1Gi" $.node_vals.local_storage_size }}
 {{- else }}
             storage: {{ default "15Gi" $.node_vals.storage_size }}
 {{- end }}
```
harryttd, line 92: I don't think we need to introduce a brand new field here. You can use logic to determine the default size and just use `storage_size`.

harryttd (changes requested):

charts/tezos/templates/nodes.yaml
puppetninja: I think it would be confusing, to configure pvcs for different purposes with same helm value, they even have different pvc names.

harryttd: i see what you are saying. Maybe we rename local_storage_size then because that sounds like the size of the local temp memory of the machine which isn't persistent and isn't what is being configured. Maybe `local_storage_pv_size`?

harryttd (approved):
Thank you!

puppetninja:
I rebased it instead of squashing my commits, sorry for that

PR #595, opened 2023-07-21
Fix missing vote files + misc cleanup

harryttd:
- Fixes bash script that checks for vote files for bakers. Script had bug if the files it expected didn't exist.
- Fixes missing vote files for accounts that will be created automatically via `should_generate_unsafe_deterministic_data`.
- closes https://github.com/oxheadalpha/tezos-k8s/issues/284
- Does a little cleanup in values.yaml so published docs on tezos-k8s.xyz has more markdown formatting.

PR #596, opened 2023-07-25
update pyyaml to 6.0.1

orcutt989:
Fixes broken pyyaml arm64 build in our CI.

### DIFF
--- a/rpc-auth/requirements.txt
+++ b/rpc-auth/requirements.txt
@@ -22,7 +22,7 @@ pysodium==0.7.5
 pytezos==2.5.11
 python-dateutil==2.8.1
 pytzdata==2020.1
-PyYAML==5.4
+PyYAML==6.0.1
 redis==3.5.3
 requests==2.26.0
 secp256k1==0.13.2


nicolasochem (approved):
thanks

harryttd (commented):
i'm not sure this should be merged yet. It is a major version update of a package which seems to be a dep of another. If the code hasn't been updated we don't know what will happen.

orcutt989:
> Please see my comment here: [#595 (comment)](https://github.com/oxheadalpha/tezos-k8s/pull/595#issuecomment-1648728333) I believe we should wait for the maintainers to fix.

Thanks @harryttd. Yes I looked at the PR and I wasn't able to understand the issue. It seems as if the arm64 build randomly stopped working for PyYAML 5.4. Could you elaborate on how the PR could be related?

harryttd:
Yeah.. as the days go on they don't seem to want to fix this for older versions https://github.com/yaml/pyyaml/pull/731

Did you test the code to make sure it still works with the major release version?

nicolasochem:
> Did you test the code to make sure it still works with the major release version?

We are not doing anything fancy with pyyaml so I'll assume it can still parse our yaml fine.

PR #597, opened 2023-08-18
PoW Faucet

harryttd:
Tezos faucet chart updates for the new PoW challenge functionality.

See https://github.com/oxheadalpha/tezos-faucet-backend/pull/2 and https://github.com/oxheadalpha/tezos-faucet/pull/16

Notable changes:
- Can configure separate RPC url for the backend code. This way the backend can query the k8s node rpc service directly.
- Deploying the frontend is now optional, so you can have just a programmatic faucet.
- Added pod security context, service account, plus other Helm k8s configuration options.
- Removed unused secret.
- General cleanup and improvement.

### DIFF
--- a/charts/tezos-faucet/templates/configmap.yaml
+++ b/charts/tezos-faucet/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.enableUI }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -11,17 +12,18 @@ data:
 {{- end }}
 {{- $_ := set .Values.config.application "profiles" $newProfiles }}
 {{ .Values.config | mustToPrettyJson | indent 4 }}
-
 ---
+{{- end }}
+
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: faucet-backend-config
   namespace: {{ .Release.Namespace }}
 data:
-  AUTHORIZED_HOST: {{ .Values.authorizedHost }}
+  AUTHORIZED_HOST: "{{ .Values.authorizedHost }}"
   DISABLE_CHALLENGES: "{{ .Values.disableChallenges }}"
   ENABLE_CAPTCHA: "{{ .Values.enableCaptcha }}"
   MAX_BALANCE: "{{ .Values.maxBalance }}"
-  REDIS_URL: {{ .Values.redis.url }}
-  RPC_URL: {{ .Values.config.network.rpcUrl }}
+  REDIS_URL: "{{ .Values.redis.url }}"
+  RPC_URL: "{{ required "config.network.rpcUrl is required." .Values.config.network.rpcUrl }}"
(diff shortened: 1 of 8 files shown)


charts/tezos-faucet/values.yaml
```diff
@@ -1,33 +1,72 @@
...
-  tezos_faucet_backend: ghcr.io/oxheadalpha/tezos-faucet-backend:latest
-authorizedHost: "'*'"
+# Faucet backend: https://github.com/oxheadalpha/tezos-faucet-backend
+# Faucet frontend: https://github.com/oxheadalpha/tezos-faucet
 
-# if your balance is over the max, the faucet won't give you more tokens
-maxBalance: 12000
+images:
+  tezosFaucet: ghcr.io/oxheadalpha/tezos-faucet:2.0.0
+  tezosFaucetBackend: ghcr.io/oxheadalpha/tezos-faucet-backend:2.0.0
```
nicolasochem, line 6: why not latest?

charts/tezos-faucet/templates/deployment.yaml
```diff
@@ -15,34 +15,77 @@ spec:
...
         # https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }} 
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
     spec:
+      restartPolicy: Always
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "tezos-faucet.serviceAccountName" . }}
```
nicolasochem, line 25: why is this needed?

charts/tezos-faucet/templates/configmap.yaml
```diff
@@ -1,21 +1,32 @@
...
 apiVersion: v1
-data:
-  config.json: |
-{{ .Values.config | mustToPrettyJson | indent 4 }}
 kind: ConfigMap
 metadata:
   name: faucet-config
   namespace: {{ .Release.Namespace }}
+data:
+  config.json: |
```
nicolasochem, line 8: why replace a kitchen-sink configuration passthrough with something more rigid?

charts/tezos-faucet/values.yaml
harryttd: If faucet code changes and releases, then  it's possible a faucet deployment could end up pulling the latest images, while the chart code is not updated to handle it.

charts/tezos-faucet/templates/deployment.yaml
harryttd: it's just one of the properties that exist on sts's and deployments. Chart generator will create one with it. It's so the pods can interact as a specific service account that may have custom permissions.

charts/tezos-faucet/templates/configmap.yaml
harryttd: Profiles now have added configuration which is for the challenges. I can't just pass the profiles object like it currently is because it contains that info. If i did, the frontend would be getting that info and it shouldn't. The frontend should not know anything about the difficulty and amount of challenges needed. Albeit right now these numbers are not dynamic, they will be more dynamic as we go forward.

nicolasochem (approved):
lgtm

PR #598, opened 2023-08-25
snapshotengine: move sleep to reduce volumes

orcutt989:
Currently the artifact job sleeps a user-defined amount of time to reduce frequency of artifact creation.  This resulted in PVCs remaining bound to the kubernetes nodes.  With the migration to Digital Ocean a new limit was encountered- only 7 volumes allowed per droplet(node) which was resulting in a need for more droplets.  

This PR moves moves the sleep out of the artifact creation job causing it to terminate. The maker job detects this, deletes the job and PVCs and sleeps instead freeing up slots.

Additional info- AWS volume-per-instance limit was at minimum 32 so this was not an issue, but is now on Digital Ocean because of the 7 volume limit per droplet.  The volume-per-droplet limit is hard platform limit and not changeable by Digital Ocean support.


orcutt989:
Setup QEMU is failing so I opened a ticket https://github.com/docker/setup-qemu-action/issues/94

Looks like something is wrong with one of the images it uses.

nicolasochem (approved):
thanks. This has been deployed for a couple of weeks and appears to be working as expected.

Issue #600, opened 2023-09-11
pre-existing activation detection misfired on Mondaynet


nicolasochem:
The below logic checks that the chain has already activated and does not activate again:

https://github.com/oxheadalpha/tezos-k8s/blob/master/charts/tezos/scripts/chain-initiator.sh#L3-L12

On Sept 11 it malfunctioned, because the RPC gave a successful reply and then failed again, therefore the job failed silently.

We need to capture the answer and not do a subsequent query.

harryttd:
closed via #601

PR #601, opened 2023-09-11
chain activation: avoid misfiring

nicolasochem:
when activate job gets a successful RPC response for the first time, it should not repeat the query; instead, just parse the output again and check whether the level is zero.

This way, if the second rpc query fails, it will just restart the job instead of passing silently, resulting in a non-activated chain stuck at level 0, as happened on Mondaynet today.

I tried:

* private chain on minikube: I see the `until` loop run a few times, and then the chain gets activated
* delete the activation job and run `helm upgrade`: job runs again, but is satisfied that activation already happened and exits silently


harryttd (approved):
closes #600

PR #608, opened 2023-10-03
snapshotEngine: Dynamic sleep time

orcutt989:
Instead of sleeping for the configured amount of time this will pull the date on the newest artifact of a given network, version, and history mode.

If the date is older than our sleep time then it starts the artifact creation job.

If the date is newer than it subtracts this from the configured sleep time and sleeps for this new amount.

This was created because we run our nodes pretty slim and sometimes they OOM. This causes the jobs to restart when they are sleeping resulting in more artifacts created than expected with a given sleep time and increased costs.

Now even if the jobs get restarted it will still dynamically calculate if the job needs to sleep or not.


nicolasochem (changes requested):
Will it work if there are no artifacts yet? (new namespace)

nicolasochem (approved):
thanks

PR #610, opened 2023-10-23
Signer authorized keys

nicolasochem:
Authorized key is the tezos native method to authenticate signing
requests, one that we use in the new tezos-kms-signer-lambda.

This adds the required support on tezos-k8s to sign with such a signer.

The way it works in octez is:

* when the baker/client connects to the signer for the first time,
  signer answers with a list of "authorized_keys" that the signature
  request must be signed with. These authorized keys are just tezos
  accounts
* if the baker/client has the secret key for one of these authorized
  keys, they will just sign every request with it. otherwise, there will
  be an error
* this can't be nested. the authorized_key can't be remote

We add support in tezos-k8s by assuming the authorized_keys are just
standard "accounts". Then, you may configure a baker as follows:

```
nodes:
  mybaker:
    bake_using_accounts:
    - mybakeraddy
    authorized_keys:
    - my_authorized_key
```

config-generator then ensures that the private authorized key is
accessible to the baker.

We also add support on octez-signer end:

```
octezSigners:
  mysigner:
    sign_for_accounts:
    - mybakeraddy
    authorized_keys:
    - my_authorized_key
```

When set, the signer mandates requests to be authenticated. Otherwise,
it signs anything.

This way, you can test end-to-end in a private chain.

We modify mkchain to do this by default: mkchain now generates an
authorized key and uses it to sign by default.

Also, mkchain was previously defaulting to using one remote signer, but
this broke when adding support for tacoInfra signer. I fixed it.

I have tested it with 3 bakers and 2 signers, one authorized and one
not. It's all working. I haven't tried zerotier and public chains.

Other changes:

* switch default version to 17.3
* no magic byte restriction from signer - prevents activation

### DIFF
--- a/charts/tezos/scripts/remote-signer.sh
+++ b/charts/tezos/scripts/remote-signer.sh
@@ -6,7 +6,11 @@ CLIENT_DIR="$TEZ_VAR/client"
 NODE_DIR="$TEZ_VAR/node"
 NODE_DATA_DIR="$TEZ_VAR/node/data"
 
-CMD="$TEZ_BIN/octez-signer -d $CLIENT_DIR launch http signer --magic-bytes 0x11,0x12,0x13 --check-high-watermark -a 0.0.0.0 -p 6732"
+extra_args=""
+if [ -f ${CLIENT_DIR}/authorized_keys ]; then
+  extra_args="${extra_args} --require-authentication"
+fi
+CMD="$TEZ_BIN/octez-signer -d $CLIENT_DIR ${extra_args} launch http signer -a 0.0.0.0 -p 6732"
 
 # ensure we can run tezos-signer commands without specifying client dir
 ln -s /var/tezos/client /home/tezos/.tezos-signer
--- a/charts/tezos/templates/_helpers.tpl
+++ b/charts/tezos/templates/_helpers.tpl
@@ -207,3 +207,31 @@ metadata:
   {{- end }}
   {{- "true" }}
 {{- end }}
+
+{{/*
+  Get list of authorized keys. Fails if any of the keys is not defined in the accounts.
+*/}}
+{{- define "tezos.getAuthorizedKeys" }}
+  {{- $allAuthorizedKeys := list }}
+  {{- /* Gather keys from nodes */}}
+  {{- range $node := .Values.nodes }}
+    {{- range $instance := $node.instances }}
+      {{- if .authorized_keys }}
+        {{- $allAuthorizedKeys = concat $allAuthorizedKeys .authorized_keys }}
+      {{- end }}
+    {{- end }}
+  {{- end }}
+  {{- /* Gather keys from octezSigners */}}
+  {{- range $signer := .Values.octezSigners }}
+    {{- if $signer.authorized_keys }}
+      {{- $allAuthorizedKeys = concat $allAuthorizedKeys $signer.authorized_keys }}
+    {{- end }}
+  {{- end }}
+  {{- /* Ensure all keys are defined in accounts and fail otherwise */}}
+  {{- $allAuthorizedKeys = uniq $allAuthorizedKeys }}
+  {{- range $key := $allAuthorizedKeys }}
+    {{- if not (index $.Values.accounts $key) }}
+      {{- fail (printf "Authorized key '%s' is not defined in accounts." $key) }}
+    {{- end }}
+  {{- end }}
+{{- end }}
--- a/charts/tezos/templates/_helpers.tpl
+++ b/charts/tezos/templates/_helpers.tpl
@@ -230,7 +230,7 @@ metadata:
   {{- /* Ensure all keys are defined in accounts and fail otherwise */}}
   {{- $allAuthorizedKeys = uniq $allAuthorizedKeys }}
   {{- range $key := $allAuthorizedKeys }}
-    {{- if not (index $.Values.accounts $key) }}
+    {{- if not (index $.Values.accounts $key "key") }}
       {{- fail (printf "Authorized key '%s' is not defined in accounts." $key) }}
     {{- end }}
   {{- end }}
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -317,7 +317,7 @@ octezSigners: {}
 #     - baker0
 #    authorized_keys:
 #     # if set, baker will only sign request authenticated by one of the authorized_keys
-#     - authorized_key0
+#     - authorized-key-0
 # ```
 #
 # Deploys a signer using AWS KMS to sign operations.
--- a/charts/tezos/values.yaml
+++ b/charts/tezos/values.yaml
@@ -8,7 +8,7 @@ is_invitation: false
 
 # Images not part of the tezos-k8s repo go here
 images:
-  octez: tezos/tezos:v17.1
+  octez: tezos/tezos:v17.3
   tacoinfraRemoteSigner: ghcr.io/oxheadalpha/tacoinfra-remote-signer:0.1.0
 # Images that are part of the tezos-k8s repo go here with 'dev' tag
 tezos_k8s_images:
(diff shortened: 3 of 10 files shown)


utils/config-generator.py
```diff
@@ -463,17 +476,21 @@ def import_keys(all_accounts):
...
         f"{tezdir}/public_key_hashs",
+        f"{tezdir}/authorized_keys",
     )
     print(f"\n  Writing {sk_path}")
     json.dump(secret_keys, open(sk_path, "w"), indent=4)
     print(f"  Writing {pk_path}")
     json.dump(public_keys, open(pk_path, "w"), indent=4)
     print(f"  Writing {pkh_path}")
     json.dump(public_key_hashs, open(pkh_path, "w"), indent=4)
+    if MY_POD_TYPE == "signing" and len(authorized_keys) > 0:
```
harryttd, line 491: could be simplified to 
```suggestion
    if MY_POD_TYPE == "signing" and len(authorized_keys):
```

harryttd (changes requested):

mkchain/tqchain/mkchain.py
```diff
@@ -243,7 +244,7 @@ def main():
         baking_accounts = {
             f"{ARCHIVE_BAKER_NODE_NAME}-{n}": {} for n in range(args.number_of_bakers)
         }
-        for account in baking_accounts:
+        for account in [*baking_accounts, *["authorized-key-0"]]:
```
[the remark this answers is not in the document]
nicolasochem: no, because mkchain currently generates only one signer for all the bakers... if there was one key per baker, every baker could sign with each other's auth key, which is not realistic...

utils/config-generator.py
nicolasochem: your suggestion is harder to read IMO

charts/tezos/scripts/remote-signer.sh
```diff
@@ -6,7 +6,11 @@ CLIENT_DIR="$TEZ_VAR/client"
 NODE_DIR="$TEZ_VAR/node"
 NODE_DATA_DIR="$TEZ_VAR/node/data"
 
-CMD="$TEZ_BIN/octez-signer -d $CLIENT_DIR launch http signer --magic-bytes 0x11,0x12,0x13 --check-high-watermark -a 0.0.0.0 -p 6732"
```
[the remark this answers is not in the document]
nicolasochem: I am not too sure. I want sane defaults for both private chain and public baker. A public baker will by default not use remote signer, then it can sign everything anyway.

It has happened to me in the past that I launch a private chain with just one baker, shell into the node and try to do transfers from the only account that exists - the baker account - but the signer won't let me. It's annoying.

I am not sure whether I would recommend mainnet baking with a hot key anyway. The guides we have posted make clear we prefer people to use KMS.

utils/config-generator.py
```diff
@@ -332,6 +332,11 @@ def expose_secret_key(account_name):
     pod.  It returns the obvious Boolean.
     """
     if MY_POD_TYPE == "activating":
+        all_authorized_keys = [key for node in NODES.values() for instance in node['instances'] for key in instance.get('authorized_keys', [])]
+        if account_name in all_authorized_keys:
+            # populate all known authorized keys in the activation account.
+            # This avoids annoying edge cases for activating private chains, when security is not critical.
+            return True
```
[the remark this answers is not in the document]
nicolasochem: reformatting done. Yes, the activation itself is the edge case. I rephrased.

nicolasochem:
I have tested this end-to-end with the lambda on ghostnet so marking this ready:

https://ghost.tzstats.com/tz1Qf1pSbJzMN4VtGFfVJRgbXhBksRv36TxW#baking

charts/tezos/scripts/remote-signer.sh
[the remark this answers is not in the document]
harryttd: perhaps in the future we'll have a config that let's users specify which operations to allow/disallow

charts/tezos/values.yaml
```diff
@@ -174,6 +174,10 @@ should_generate_unsafe_deterministic_data: false
 #                            Don't also set `bake_using_accounts`.
 #    - `bake_using_accounts`: List of account names that should be used for baking.
 #                            Don't also set `bake_using_account`.
+#    - `authorized_keys`: Keys used to authenticate a baker to a signer.
+#                         When a baker uses a remote signer that requires
+#                         authentication, the relevant key from this list
+#                         will be used to sign every signature request.
```
harryttd, line 180: It's not clear except for the signer comment doc later on that this is a list of account names. Not actual keys. The comment indicates that keys are what should be used.

charts/tezos/values.yaml
```diff
@@ -314,6 +318,12 @@ octezSigners: {}
 #  tezos-signer-0:
 #    accounts:
 #     - baker0
+#    authorized_keys:
+#     # Keys used to authenticate the baker to the signer.
+#     # The baker must have the private key for one of the
+#     # listed accounts. The signer will only sign a request
+#     # from a baker authenticated by an allowed key.
+#     - authorized-key-0
```
harryttd, line 326: Comment implies keys should be used while the example show's an account name.

harryttd (changes requested):

harryttd:
Perhaps Helm should do validation on the auth accounts. It could fail fast if for example, a signer/baker specifies an account that doesn't exist.

harryttd:
My thinking is if an auth key is not defined, and generate unsafe data is true, that they can be auto generated similar to baking accounts. https://github.com/oxheadalpha/tezos-k8s/blob/cc7781527a207f6e9b3d8b1fdc30518b55e914bc/utils/config-generator.py#L221

charts/tezos/values.yaml
nicolasochem: fixed

charts/tezos/values.yaml
nicolasochem: fixed

nicolasochem:
> Perhaps Helm should do validation on the auth accounts. It could fail fast if for example, a signer/baker specifies an account that doesn't exist.

Done.

> My thinking is if an auth key is not defined, and generate unsafe data is true, that they can be auto generated similar to baking accounts.

Can I skip? I actually want to deprecate account autogeneration. I think generation of accounts should happen when you create the yaml file (with mkchain or otherwise).

charts/tezos/templates/_helpers.tpl
```diff
@@ -207,3 +207,31 @@ metadata:
...
+  {{- /* Gather keys from octezSigners */}}
+  {{- range $signer := .Values.octezSigners }}
+    {{- if $signer.authorized_keys }}
+      {{- $allAuthorizedKeys = concat $allAuthorizedKeys $signer.authorized_keys }}
+    {{- end }}
+  {{- end }}
+  {{- /* Ensure all keys are defined in accounts and fail otherwise */}}
+  {{- $allAuthorizedKeys = uniq $allAuthorizedKeys }}
+  {{- range $key := $allAuthorizedKeys }}
+    {{- if not (index $.Values.accounts $key) }}
```
harryttd, line 233: This is great. I would also make sure that an actual key is provided on the account definition. 
```suggestion
    {{- if not (index $.Values.accounts $key "key") }}
```

utils/config-generator.py
```diff
@@ -332,6 +332,17 @@ def expose_secret_key(account_name):
...
+        all_authorized_keys = [
+            key
+            for node in NODES.values()
+            for instance in node["instances"]
+            for key in instance.get("authorized_keys", [])
+        ]
+        if account_name in all_authorized_keys:
+            # Populate authorized keys known by all bakers in the activation account.
+            # This ensures that activation will succeed with a remote signer that requires auth,
+            # regardless of which baker does it.
```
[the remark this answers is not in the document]
nicolasochem: Yes, right, it probably does not have to be a baker, since at block 0, the list of bakers is not known... But what's the alternative? create a dedicated list of authorized_keys for activation? I've never activated a chain with an account that's not a baker, so I think it's fine.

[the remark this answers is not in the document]
harryttd: yeah i think it's fine as is. 

What could be done is look up the signer that is signing for the baker baking with the activation key, get this signer's auth keys, and then only import those keys. If no signer is found, then no auth key would be required for activation.

harryttd (approved):

PR #611, opened 2023-11-18
when snapshots can't be found for version, just pick the most recent

nicolasochem:
Right now there are no v18 snapshots on xtz-shots, so it's not possible to kickstart a node. This PR will pick the v17 snapshot instead.


nicolasochem:
verified: I was able to boot a v18 rolling node from a v17 snapshot or tarball.

